Issue #57, opened 2019-05-24
Upgrade Kubebuilder to v2.0.0


baluchicken:
Upgrade Kubebuilder to v2.0.0

PR #65, opened 2019-05-30
FIX: generate target depends on vendor packages

tibkiss:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | Not this time (I'm lazy)

### What's in this PR?
Build fix on freshly checked out instance.

### Why?
'make' on fresh checkout fails due to missing vendor folder.
Added vendor as a dependency for the generate target to fix the issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,7 +104,7 @@ vet:
 	go vet ./...
 
 # Generate code
-generate:
+generate: vendor
 ifndef GOPATH
 	$(error GOPATH not defined, please define GOPATH. Run "go help gopath" to learn more about GOPATH)
 endif


baluchicken:
Thanks for the contribution, the reason why the vendor goal  is missing from make is to make the developer process (testing, generating, building) as fast as possible. Make vendor only needs to run when some dependency is updated, or checking out the repo on the first time. So running it every time during make just slows down the development. On the other hand kubebuilder version 2.0.0 will be released hopefully on the next week which supports go modules. Please follow this issue #57  and stay tuned.

tibkiss:
Makes sense. 
I'd recommend that one documents this step on the README so that it is easier to get a head start for 3rd parties.

baluchicken:
Hi @tibkiss,
You should check this: https://github.com/banzaicloud/kafka-operator/blob/master/docs/developer.md

tibkiss:
👍

PR #94, opened 2019-07-31
Cruise Control Topic create fix

baluchicken:
| Q               | A
| --------------- | ---
| Bug fix?        |yes|
| New feature?    |no|
| API breaks?     |no|
| Deprecations?   |no|

### What's in this PR?

This PR contains the following fixes:

- fix namespace issue when operator deployed to a different namespace than cluster
- change go client lib to samara to leverage new admin client feature
- implement better reconcile for cc topic create
- fix benchmark doc links

### Checklist

- [x] Implementation tested


pkg/resources/cruisecontrol/topicManager.go
```diff
@@ -19,73 +19,70 @@ import (
...
+	err = adminClient.CreateTopic("__CruiseControlMetrics", &sarama.TopicDetail{
 		NumPartitions:     12,
 		ReplicationFactor: 3,
-	}
+	}, false)
 
-	err = conn.CreateTopics(tConfig)
-	if err != nil {
-		return emperror.Wrap(err, "could not create topic for CC")
+	if err != nil && err.(*sarama.TopicError).Err != sarama.ErrTopicAlreadyExists {
```
bonifaido: I'm sure `CreateTopic` can return something other than `TopicError`, in which case this cast will `panic`.

pkg/resources/cruisecontrol/topicManager.go
```diff
@@ -19,73 +19,73 @@ import (
...
 		ReplicationFactor: 3,
-	}
+	}, false)
 
-	err = conn.CreateTopics(tConfig)
-	if err != nil {
-		return emperror.Wrap(err, "could not create topic for CC")
+	var tError *sarama.TopicError
+
+	if err != nil && !(errors.As(err, &tError) && tError.Err == sarama.ErrTopicAlreadyExists) {
```
pepov: tError is already a pointer is `&` necessary here?

baluchicken: yes, because the error interface is implemented by using *TopicError as receiver.
The As method dereference the pointer so if I use just `tError` here it means the type will be TopicError which actually does not implement the error interface. It is a little bit odd I know:)

PR #96, opened 2019-08-05
add support for broker classes

almariah:
| Q               | A
| --------------- | ---
| Bug fix?        | no|yes
| New feature?    | yes
| API breaks?     | no|yes
| Deprecations?   | no|yes

### What's in this PR?

This PR add support for defined brokers spec (`brokerclasses`) to be used in scaling kafka cluster. The current implementation use many annotations in the alert definitions which is not supporting all broker config like CPU/Memory resources. Instead of these annotations, a single annotation of alerts could be used (`brokerClass`). Further the classes could be defined in `KafkaCluster` spec. The current implementation will also be kept for compatibility.

Also the prometheus namespace label (`kubernetes_namespace`) is replaced with (`namespace`) which is more common for latest prometheus/ prometheus-operator. The `ServiceMonitor` CRD by default will re-label `__meta_kubernetes_namespace` to `namespace`.

### Checklist

- [x] Implementation tested

### To Do

- [x] Modify CRD
- [x] Modify prometheus config

### DIFF
--- a/pkg/apis/banzaicloud/v1alpha1/kafkacluster_types.go
+++ b/pkg/apis/banzaicloud/v1alpha1/kafkacluster_types.go
@@ -28,17 +28,17 @@ import (
 
 // KafkaClusterSpec defines the desired state of KafkaCluster
 type KafkaClusterSpec struct {
-	HeadlessServiceEnabled bool                         `json:"headlessServiceEnabled"`
-	ListenersConfig        ListenersConfig              `json:"listenersConfig"`
-	ZKAddresses            []string                     `json:"zkAddresses"`
-	RackAwareness          *RackAwareness               `json:"rackAwareness,omitempty"`
-	BrokerConfigs          []BrokerConfig               `json:"brokerConfigs"`
-	BrokerClasses          map[string]*BrokerConfig     `json:"brokerClasses"`
-	OneBrokerPerNode       bool                         `json:"oneBrokerPerNode"`
-	CruiseControlConfig    CruiseControlConfig          `json:"cruiseControlConfig"`
-	EnvoyConfig            EnvoyConfig                  `json:"envoyConfig,omitempty"`
-	ServiceAccount         string                       `json:"serviceAccount"`
-	MonitoringConfig       MonitoringConfig             `json:"monitoringConfig,omitempty"`
+	HeadlessServiceEnabled bool                     `json:"headlessServiceEnabled"`
+	ListenersConfig        ListenersConfig          `json:"listenersConfig"`
+	ZKAddresses            []string                 `json:"zkAddresses"`
+	RackAwareness          *RackAwareness           `json:"rackAwareness,omitempty"`
+	BrokerConfigs          []BrokerConfig           `json:"brokerConfigs"`
+	BrokerClasses          map[string]*BrokerConfig `json:"brokerClasses"`
+	OneBrokerPerNode       bool                     `json:"oneBrokerPerNode"`
+	CruiseControlConfig    CruiseControlConfig      `json:"cruiseControlConfig"`
+	EnvoyConfig            EnvoyConfig              `json:"envoyConfig,omitempty"`
+	ServiceAccount         string                   `json:"serviceAccount"`
+	MonitoringConfig       MonitoringConfig         `json:"monitoringConfig,omitempty"`
 }
 
 // KafkaClusterStatus defines the observed state of KafkaCluster
--- a/pkg/apis/banzaicloud/v1alpha1/kafkacluster_types.go
+++ b/pkg/apis/banzaicloud/v1alpha1/kafkacluster_types.go
@@ -33,7 +33,7 @@ type KafkaClusterSpec struct {
 	ZKAddresses            []string                 `json:"zkAddresses"`
 	RackAwareness          *RackAwareness           `json:"rackAwareness,omitempty"`
 	BrokerConfigs          []BrokerConfig           `json:"brokerConfigs"`
-	BrokerClasses          map[string]*BrokerConfig `json:"brokerClasses, omitempty"`
+	BrokerConfigGroups     map[string]*BrokerConfig `json:"brokerConfigGroups,omitempty"`
 	OneBrokerPerNode       bool                     `json:"oneBrokerPerNode"`
 	CruiseControlConfig    CruiseControlConfig      `json:"cruiseControlConfig"`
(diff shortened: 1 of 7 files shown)


baluchicken:
Thanks for submitting this PR. I like the idea of having a feature like this, but now the alerts and the cluster CR is a separate thing, I would like to keep this behaviour. I would propose the following:
Create an annotation eg.: `brokerConfig` to the Prometheus Alert which contains the BrokerConfig struct as string, if this annotation is present code will use that but otherwise the code falls back to the previous behaviour. With this we can keep the alerts and the cluster CR separated, but get the required functionality. What do you think?

almariah:
@baluchicken If I understand your idea correctly, I think it make sense to have the `brokerConfig` for scaling up Independent on the cluster config itself. Considering the following:
* Defining different class or as you mentioned `brokerConfig` (which could grow in the future like adding toleration, nodeSelector, etc) in alert annotation, and also the annotations for action, any desired annotations and etc would be a lot for the alert definition.
* Having broker config that will be added could be better and more transparent if they are in the same cluster config. If for example I would like to compare them with current broker that is already defined.
* Defining the `brokerConfig` as string in annotation could be miss-configured easily.
* Configuring heterogeneous brokers is a great feature of the operator, but also configuring homogeneous could be desired in some cases. Imagine someone wanted 20 broker cluster with same config, then you need to define them explicitly. What about and `brokerGroupsConfigs` property that could be configured as follow:
```yaml
apiVersion: banzaicloud.banzaicloud.io/v1alpha1
kind: KafkaCluster
metadata:
  name: kafka
spec:
  headlessServiceEnabled: false
  zkAddresses:
    - ...
  oneBrokerPerNode: false
  brokerClasses:
    default:
      image: "wurstmeister/kafka:2.12-2.1.0"
      .....
  brokerGroupsConfigs:
    - replica: 20
      class: default
    - replica: 4
      class: extra-large
  # disabled explicit broker config
  # They will populated here with controller based on brokerGroupsConfigs
  # You can still use `brokerConfigs`
  # brokerConfigs:
```

Having such feature of `brokerGroupsConfigs` will simplify the broker configs. It will using the classes definitions. That means the having such classes in the cluster config will be used to define the cluster itself and moreover the brokers that could be scaled up.

What do you think? if the idea of `brokerGroupsConfigs` makes sense, I can work on it in another PR. Further we can have initially the `brokerClasses` (or whatever the name for it) to be used for scaling and later for defining the cluster more easily. 

Thanks!

baluchicken:
Sorry for the late response @almariah , thanks for the detailed proposal.
This PR made us rethink the operator's whole CRD structure. Bottom lime, we should definitely modify the CRD in a way which supports the required features.

How about considering the followings:

Kafka Supports three different configs:
- ReadOnly
- ClusterWide
- PerBroker

Based on these we should create the following CRDs:

```
ClusterImage: string
ReadOnlyConfig: string
ClusterWideConfig: string
BrokerConfigGroups: map[string]BrokerConfig
[]Brokers:
  ID: int32
  Image: string
  BrokerConfigGroup: string
  NodeAffinity: corev1.Nodeaffinity
  BrokerConfig: BrokerConfig
```

ReadOnlyConfig -> Kafka Config type ReadOnly
ClusterWide -> Kafka Config type ClusterWide
BrokerConfig -> Kafka Config type PerBroker  

ReadOnlyConfig and ClusterWide config should be set globally and only the perBroker config should be grouped.
Also I think we should give the ability to the user to define the `brokerImages` globally/clusterwide. (The image specified in broker will overwrite the globally specified image.)

I think this CRD will ease the process to create homogeneous cluster with the operator, also `BrokerConfigGroups` can be used in alerts as well.
I think from now we can merge the `brokerClasses` implementation which will be replaced by `BrokerConfigGroups` once it is complete.
What do you think?
Thanks!

pkg/apis/banzaicloud/v1alpha1/kafkacluster_types.go
```diff
@@ -28,16 +28,17 @@ import (
...
-	CruiseControlConfig    CruiseControlConfig `json:"cruiseControlConfig"`
-	EnvoyConfig            EnvoyConfig         `json:"envoyConfig,omitempty"`
-	ServiceAccount         string              `json:"serviceAccount"`
-	MonitoringConfig       MonitoringConfig    `json:"monitoringConfig,omitempty"`
+	HeadlessServiceEnabled bool                     `json:"headlessServiceEnabled"`
+	ListenersConfig        ListenersConfig          `json:"listenersConfig"`
+	ZKAddresses            []string                 `json:"zkAddresses"`
+	RackAwareness          *RackAwareness           `json:"rackAwareness,omitempty"`
+	BrokerConfigs          []BrokerConfig           `json:"brokerConfigs"`
+	BrokerClasses          map[string]*BrokerConfig `json:"brokerClasses"`
```
baluchicken:
```suggestion
	BrokerClasses          map[string]*BrokerConfig `json:"brokerClasses, omitempty"`
```

almariah:
@baluchicken That is a good idea.  I can imagine it like

```yaml
ClusterImage: string
ReadOnlyConfig: string
ClusterWideConfig: string
BrokerConfigGroups: map[string]BrokerConfig
[]Brokers:
  ID: int32
  BrokerConfigGroup: string
  BrokerConfig: BrokerConfig
```

and for `BrokerConfig`

```yaml
Image: string
NodeAffinity: corev1.Nodeaffinity
StorageConfigs:   []StorageConfig
NodeSelector: ..
Tolerations: ...
```

Where:

* `BrokerConfigGroups` defines `BrokerConfig` groups/classes/templates
*  `[]Brokers:` defined list of desired brokers.
*  Each broker should have `ID`
*  Each broker could have `BrokerConfigGroup` or `BrokerConfig` or both (where `BrokerConfig` will overrides `BrokerConfigGroup` if applicable)
* `BrokerConfig` could have anything that could configured per broker (except the ID)

It is slight different from yours. Anyway I like the idea which will simplify the CRD and make more usable. 

If possible also to consider replica count, if someone is lazy to write a list of many brokers. Just a replica count and  `BrokerConfigGroup`.

almariah:
@baluchicken just about the current PR, should we change `BrokerClasses` to `BrokerConfigGroups`, for future change to not deprecate `BrokerClasses` soon or just keep like that ?

baluchicken:
@almariah yeah it is a good idea change `BrokerClasses` to `BrokerConfigGroups`. Thanks. Also thanks for the proposal we are going to start working on that soon.

bonifaido:
Thank you @almariah !

PR #105, opened 2019-08-23
Use the latest kubebuilder v2.0.0

baluchicken:
| Q               | A
| --------------- | ---
| Bug fix?        |no|
| New feature?    |yes|
| API breaks?     |no|
| Deprecations?   |no|
| Related tickets | fixes #57 
| License         | Apache 2.0

### What's in this PR?

This PR updates the operator to use the new v2.0.0 kubebuilder.
Which includes newer controller libs and go mod support

### Checklist

- [x] Implementation tested
- [x] Error handling code meets the [guideline](https://github.com/banzaicloud/pipeline/blob/master/docs/error-handling-guide.md)
- [x] Logging code meets the guideline
- [x] User guide and development docs updated (if needed)

### To Do
Implement the validating webhook, which was by default generated with kubebuilder v1.0.0

PR #108, opened 2019-08-23
added imagepullsecret to the CRD/pod creation

prune998:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes https://github.com/banzaicloud/kafka-operator/issues/106

### What's in this PR?
Added `ImagePullSecrets` parameter to the `KafkaCluster` CRD so we can define the secrets needed to pull the Container image from a private repository.

### Why?
see Issue https://github.com/banzaicloud/kafka-operator/issues/106

### Checklist

- [ ] Implementation tested

### DIFF
--- a/charts/kafka-operator/templates/operator-kafka-crd.yaml
+++ b/charts/kafka-operator/templates/operator-kafka-crd.yaml
@@ -53,7 +53,9 @@ spec:
                   nodeSelector:
                     type: object
                   tolerations:
-                    type: object
+                    items:
+                      type: object
+                    type: array
                   resourceReqs:
                     type: object
                   storageConfigs:
@@ -97,6 +99,10 @@ spec:
               type: object
             headlessServiceEnabled:
               type: boolean
+            imagePullSecrets:
+              items:
+                type: object
+              type: array
             listenersConfig:
               properties:
                 externalListeners:
--- a/config/crds/banzaicloud_v1alpha1_kafkacluster.yaml
+++ b/config/crds/banzaicloud_v1alpha1_kafkacluster.yaml
@@ -28,6 +28,8 @@ spec:
           type: object
         spec:
           properties:
+            brokerConfigGroups:
+              type: object
             brokerConfigs:
               items:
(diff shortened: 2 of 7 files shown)


prune998:
I just added it.
As a side note, the `Deployments` used by Envoy and CC are really basics. Even the `ServiceAccount` wasn't set in them.
I added it for now, but one could decide to use it's own Kafka image while using public onces for CC or Envoy.
If we don't want to be too limitative, we should allow to use different SA and ImagePullSecrets for each one of them... making a more complex `CRD`...

Let me know what you think while reviewing this commit.

Thanks.

prune998:
Maybe I should also update `config/crds/banzaicloud_v1alpha1_kafkacluster.yaml` and `charts/kafka-operator/templates/operator-kafka-crd.yaml` to reflect the change  ? 
Latest K8s versions may reject the CR if they contains fields not explecitely defined in the CRD... don't remember which...

(I'm new to Kubebuilder, and I'm just testing the V2 now, I was previously using coreOS Operator SDK, please don't blame me for my ignorance... :) )

prune998:
BTW, I already signed the CLA, and I just did again...

baluchicken:
No worries, I am just testing the v2 version as well. Check the #105 

You are right please run the `make manifests` and `make generate` to regenerate the crd. It will regenerate `config/crds/banzaicloud_v1alpha1_kafkacluster.yaml` but not the one in dir `chart`. Please update that with the help of the first one.

We are in a middle of redesigning the whole CRD. It is indeed a good point to allow to use different `ServiceAccount` and `ImagePullSecrets`. I am thinking to use `PodSpec` in the CRD for Brokers, Envoy, CruiseControl. With that we can get every field which is configurable but the CRD does not get too messy.

baluchicken:
@prune998 yes all commits should be verified, but no worries I will merge this as is, but next time please use the right user. Thanks

PR #115, opened 2019-08-29
Remove legacy crds dir and add cluster scope to CRD

baluchicken:
| Q               | A
| --------------- | ---
| Bug fix?        |yes|
| New feature?    |no|
| API breaks?     |no|
| Deprecations?   |no|

### What's in this PR?

This PR restores the Cluster Scope of the CRD and removes legacy crds dir


baluchicken:
The related chart is not released yet, we are planning to do it when all feature is complete on Rolling Upgrade.

Issue #47, opened 2019-05-21
User Management via CRD


baluchicken:
**Describe the solution you'd like to see**
It would be useful to manage users using a custom CRD

prune998:
`Strimzi` is doing this well I think.... https://strimzi.io/docs/latest/#type-KafkaUser-reference

baluchicken:
Thanks for doing this @tinyzimmer

Issue #46, opened 2019-05-21
Topic management via CRD


pbalogh-sa:
**Describe the solution you'd like to see**
It would be useful if topic management could be configurable via CRD

prune998:
`Strimzi` is doing this well... not sure if something more complicated is needed ? 
Been using Kafka for years and never had to use more tweaking that what is presented in https://strimzi.io/docs/latest/#type-KafkaTopic-reference

At least it would be a starting point

prune998:
Anyone working on this ?

baluchicken:
Hi @prune998, thanks for asking. Currently we have higher priority features please check the [roadmap](https://github.com/banzaicloud/kafka-operator/blob/master/docs/roadmap.md) for more details.

tinyzimmer:
@baluchicken FWIW I am working on a kafka on Kubernetes implementation right now just using helm's `incubator/kafka` hacked up to properly support SSL (and disable PLAINTEXT). I wanted topic/user mgt via CRDs and wrote a sidecar operator that can handle them using `operator-sdk` and talking to the brokers via `sarama.ClusterAdmin`. 

It could probably be easily plugged into this project either also as a sidecar to the clusters, or since it's already go scaffolding for the most part, it could also be lifted into the main operator I think. It currently gets its broker URL and TLS config from environment variables and a mounted secret, but if it were to be brought into the main operator it would probably make more sense to get auth information per event directly from the cluster CRDs and secrets (using a `clusterRef` from the KafkaUser/KafkaTopic CR specs or something). 

The operator assumes a `cert-manager` issuer backing the SSL for users but it could be refactored to point at some other TLS provider. I have another tiny operator that can inject a `cert-manager` certificate-secret with JKS format (I use it for the brokers), but I can't decide if that's really needed or where it's final home would be. I haven't paid any attention to SASL as an option. 

Maybe we can help each other out and then I could move to using a cluster that is a lot more impressive than my current setup, but still have my CRDs for CI/CD 😛 . I can put the controller on my Github, but I'd probably want to rename the apigroups and stuff first. If you'd like me to submit a PR I'd be open to any design changes you all recommend.

tinyzimmer:
To follow up, I decided to just fork and give a go at merging my stuff into this operator. I managed to get it working and if you'd be open to a PR I can definitely get one started.  The fork is here: https://github.com/tinyzimmer/kafka-operator 

A few main things to note:

[...]

PR #116, opened 2019-09-01
KafkaTopic/KafkaUser CRD Support

tinyzimmer:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | maybe? `cert-manager` dependency in current state
| Deprecations?   | no
| Related tickets | #46 #47 

### What's in this PR?

This PR provides `KafkaUser` and `KafkaTopic` CRDs. Additionally, it allows the user the option to have the controller create a full PKI for the cluster backed by `cert-manager`. 

If the user still provides their own secret, a `ClusterIssuer` is made from that secret for provisioning kafka users. This introduces a hard-dependency on `cert-manager` for SSL support if it wasn't already there. I'm contemplating adding the logic to have the controller carry out all the CA functions. It would not be that difficult. 

Below are examples of the CRs:

```yaml
apiVersion: banzaicloud.banzaicloud.io/v1alpha1
kind: KafkaTopic
metadata:
  name: test-topic
  namespace: kafka
spec:
  clusterRef:
    name: kafka
  name: test-topic
  partitions: 6
  replicationFactor: 2
  config:
    "retention.ms": "800000"
    "cleanup.policy": "delete"
---
apiVersion: banzaicloud.banzaicloud.io/v1alpha1
kind: KafkaUser
metadata:
  name: test-kafkauser
  namespace: kafka
spec:
  clusterRef:
    name: kafka
  name: test-kafkauser
  secretName: test-kafkauser-secret
  includeJKS: true
  topicGrants:
    - topicName: test-topic
      accessType: read
    - topicName: test-topic
      accessType: write

```

### Why?

Because awesomeness

### Additional context

I've tested various SSL KafkaCluster configurations and all appears to work as expected. I am able to then mount my `KafkaUser` secrets inside a pod and produce/consume from Kafka using both the console producer and consumer. 

I was having some issue with testing that could have been related to my additions, or possibly just the environment I was testing against. `Namespace` kept being an empty string when trying to create clusters. Then the reconciling steps for envoy kept failing due to "namespace being required." Removing `//+kubebuilder:resource:path=kafkaclusters,scope=Cluster` from the type fixed this for me, but it's possible there is a better fix. 

### Checklist

- [X] Implementation tested
- [X] Error handling code meets the [guideline] - _maybe? you may see some places for improvement_
- [X] Logging code meets the guideline - _same as error handling_

### To Do

- [x] User guide and development docs updated 
- [x] Update helm chart with webhook-service, potential new dependencies, and new CRDs
- [x] There is some leftover `operator-sdk` commentary that I should prune out. I haven't used `kubebuilder` before so still learning.
- [x] There is some redundant code about now, mainly due to the new `kafkautil` package addition.
- [x] I should probably give the cluster ownership over users/topics made in reference to it. That way they can be properly garbage collected on deleting a `kafkacluster`. 
- [x] Increase memory limits in `config/manager/manager.yaml`? They are realllly low.


tinyzimmer:
I see the me not copying licensing info made ci mad. I can take care of that later.

controllers/kafkatopic_controller.go
```diff
@@ -0,0 +1,308 @@
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// we got a topic back
+	if existing != nil {
+		// check if requesting a partition decrease, we can't do this
+		reqLogger.Info("Topic already exists, verifying configuration")
+		if existing.NumPartitions > instance.Spec.Partitions {
+			reqLogger.Info(fmt.Sprint("IGNORING: Spec is requesting partition decrease from", existing.NumPartitions, "to", instance.Spec.Partitions))
```
tinyzimmer: I think I need to make sure I don't requeue here

tinyzimmer:
Another thought, since I am canning the ACLs. I am almost wondering if it makes more sense to have distinct `KafkaConsumer` and `KafkaProducer` CRs. Then instead of `topicGrants` they would each just have a list of `topics`. 

A later implementation could have a `KafkaAdmin` that could be used for all the various off-the-shelf manager utilities that teams use for development.

tinyzimmer:
One more thought before I take off for the night - Should I make the `KafkaTopic` names become the actual topic name and then remove the `name` field from the spec?

Considerations

If I don't, I need to implement logic to make sure we don't write over an existing topic with a new CR using the same name.

If I do, the user will not be able to use certain characters in their topic names. For example, underscores (_).

baluchicken:
Hi @tinyzimmer Thanks for this awesome PR, I just ran through and looks cool. I will have time to review it, during the second half of the week. 

About the empty namespace issue, I mistakenly added `//+kubebuilder:resource:path=kafkaclusters,scope=Cluster` but it is not required.

tinyzimmer:
I've done most of the cleanup I had originally planned for. I'm beating my head on one more issue that I can't decide the best way to approach.

1. User creates `KafkaTopic`
2. User deletes `KafkaCluster` - _causes previous KafkaTopic CR to be deleted_
3. User creates new `KafkaCluster` pointing at same zookeeper as last one
4. User attempts to recreate `KafkaTopic` and webhook denies them because zookeeper already has the topic in its store.

baluchicken:
> I've done most of the cleanup I had originally planned for. I'm beating my head on one more issue that I can't decide the best way to approach.
> 
> 1. User creates `KafkaTopic`
> 2. User deletes `KafkaCluster` - _causes previous KafkaTopic CR to be deleted_
> 3. User creates new `KafkaCluster` pointing at same zookeeper as last one
> 4. User attempts to recreate `KafkaTopic` and webhook denies them because zookeeper already has the topic in its store.
> 
> I'm almost wondering if I want to leverage the zookeeper associated with the kafka cluster to store some operator runtime data and have it "send messages" to itself. For example, place a key that references topics to a cluster, then on the admission-webhook, use that to see if the `KafkaCluster` for that topic still exists. If it's gone "release the lock". Something about this idea feels very ugly though.

I don't think we should interact with the zookeeper at all. 
If I understand the code well, when the cluster is removed, the topic controller simply removes the finaliser from the KafkaTopics owned by the cluster so because of the owner reference Kubernetes will handle the rest and delete the CRs.
I think what we should do is the following, we should place a finaliser on the KafkaCluster CR so Kubernetes will not delete it immediately. Before the Operator removes the KafkaCluster CR-s finaliser it should delete all associated KafkaTopics,KafkaUser CRs. That way the Topics/Users will be deleted before the cluster. I think this is our only chance since, we need the KafkaCluster up and running to be able to delete topics. Kubernetes owner reference works in a way that it only triggers the "childrens" delete when the owner is deleted, but it is too late for us.

tinyzimmer:
Yea that makes sense. I'll take a look at it tonight and see what I can do.

pkg/resources/pki/certs.go
```diff
@@ -0,0 +1,228 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+// A full PKI for Kafka
+
+func (r *Reconciler) kafkapki() ([]runtime.Object, error) {
+	rootCertMeta := templates.ObjectMeta(fmt.Sprintf(brokerCACertTemplate, r.KafkaCluster.Name), labelsForKafkaPKI(r.KafkaCluster.Name), r.KafkaCluster)
+	rootCertMeta.Namespace = "cert-manager"
+
+	if r.KafkaCluster.Spec.ListenersConfig.SSLSecrets.Create {
```
pbalogh-sa: If the `sslSecrets` isn't defined in CR because you want to try without ssl, this will be a nil pointer dereference.

tinyzimmer: Good catch, I think I might have that problem in the end of `pki.go` also. Will fix.

tinyzimmer: I fixed this by just returning at the beginning of the Reconcile if `sslSecrets` is nil. I also found an issue with when I took out a bunch of hardcoded strings and made them constants. I left one wrong by mistake.

PR #118, opened 2019-09-03
add external nodeport access

### LINKED ISSUE #107: Can we support nodeport for external listener access instead of lb?

I want to deploy this operator in my private k8s with the default yaml kafkacluster_without_ssl.yml. But it failed because operator wanted to get the ip of lb. Can I deploy this without lb? I just want to use NodePort.

A clear and concise description of what would you like to happen.

**Additional context**

yeya24:
| Q               | A
| --------------- | ---
| Bug fix?        | no|yes
| New feature?    |yes
| API breaks?     | yes
| Deprecations?   | no|yes
| Related tickets | fixes #107 , partially #Y, mentioned in #Z

### What's in this PR?

Support ServiceType `NodePort` for external access

### Checklist

- [ ] Implementation tested

### To Do


pkg/resources/kafka/configmap.go
```diff
@@ -37,17 +41,38 @@ func (r *Reconciler) configMapPod(broker banzaicloudv1alpha1.BrokerConfig, loadB
...
+		pod := pods.Items[0]
+		if err := cli.Get(context.TODO(), types.NamespacedName{Namespace: namespace,
+			Name: fmt.Sprintf("%s-%d-svc", crName, broker.Id)}, &svc); err != nil {
+			log.Error(err)
+		}
+
+		for _, eListener := range l.ExternalListeners {
+			advertisedListenerConfig = append(advertisedListenerConfig,
+				fmt.Sprintf("%s://%s:%d", strings.ToUpper(eListener.Name), pod.Status.HostIP, svc.Spec.Ports[0].NodePort))
+		}
```
tinyzimmer: I decided to get fancy on my branch and try out using a go template for the broker configuration. I'm not really tied to it though and can throw away my changes there if we want. I just thought going forward it might look a little cleaner. Here it is so you don't have to go fishing - https://github.com/tinyzimmer/kafka-operator/blob/master/pkg/resources/kafka/configmap.go#L32 

But if we do want to keep it, I'll need to re-implement this I think because it won't merge cleanly.

yeya24: It' s OK. I can wait until your branch is merged and I will rebase.

pkg/resources/kafka/configmap.go
```diff
@@ -37,17 +41,38 @@ func (r *Reconciler) configMapPod(broker banzaicloudv1alpha1.BrokerConfig, loadB
...
+		pod := pods.Items[0]
+		if err := cli.Get(context.TODO(), types.NamespacedName{Namespace: namespace,
+			Name: fmt.Sprintf("%s-%d-svc", crName, broker.Id)}, &svc); err != nil {
+			log.Error(err)
+		}
+
+		for _, eListener := range l.ExternalListeners {
+			advertisedListenerConfig = append(advertisedListenerConfig,
+				fmt.Sprintf("%s://%s:%d", strings.ToUpper(eListener.Name), pod.Status.HostIP, svc.Spec.Ports[0].NodePort))
```
baluchicken: Using `pod.Status.HostIP` as an external access address will not work in every case.
For example.: on an Amazon cluster the `pod.Status.HostIP` will contain an address which is not a public IP.
Unfortunately the public IP of the Node will not be available in the Pod. 
What you need to do is instead wait until the pod is scheduled to a node, get the `nodeName` field from the Pod's spec, get the given node from Kubernetes. 
The node has something called `addresses` in it's status which contains the required information. https://kubernetes.io/docs/concepts/architecture/nodes/#addresses.

I recommend to use a logic like we did in case of the rack awareness feature.
Basically we delayed the ConfigMap creation on the broker until we have all required information. Since Kubernetes schedules the Pod on a suitable Node before attaching the ConfigMap, which means you have all required information to generate the config.

config/samples/kafkacluster_without_ssl.yaml
```diff
@@ -54,7 +55,7 @@ spec:
     externalListeners:
       - type: "plaintext"
         name: "external"
-        externalStartingPort: 19090
+        externalStartingPort: 30010
```
baluchicken: Please create a whole new sample for the NodePort access.

pkg/resources/kafka/configmap.go
yeya24: Thanks for your advise. I will re-check it.

baluchicken: Sorry for the delayed reply. If that is your case I think we should not hard code anything. I recommend to let the user to make the decision. There are multiple `NodeAddressType` available in the above mentioned status field, I think we should redesign the CRD to allow to set the address type.

baluchicken:
Any update on this @yeya24?

Issue #104, opened 2019-08-22
many "failed to reconcile resource"


prune998:
**Describe the bug**
When creating a cluster (deploying the CR), I get many errors like : 
```
"error":"failed to reconcile resource: updating cr with rack awareness info failed: fetching Node rack awareness labels failed: Node \"\" not found"
```
and
```
"error":"failed to reconcile resource: updating resource failed: Service \"kf-kafka2-112\" is invalid: [metadata.ownerReferences.apiVersion: Invalid value: \"\": version must not be empty, metadata.ownerReferences.kind: Invalid value: \"\": kind must not be empty]"
```

**Steps to reproduce the issue:**
upload a CR like : 

```
apiVersion: banzaicloud.banzaicloud.io/v1alpha1
kind: KafkaCluster
metadata:
  labels:
    controller-tools.k8s.io: "1.0"
  name: kf-kafka2
spec:
  headlessServiceEnabled: false
  zkAddresses:
    - "test-zookeeper-client.alerting:2181"
  rackAwareness:
    labels:
      - "failure-domain.beta.kubernetes.io/region"
      - "failure-domain.beta.kubernetes.io/zone"
  oneBrokerPerNode: false
  brokerConfigs:
    - image: "wurstmeister/kafka:2.12-2.1.0"
      id: 111
      resourceReqs:
        limits:
          memory: "2Gi"
        requests:
          memory: "2Gi"
          cpu: "1"
      config: |
        auto.create.topics.enable=false
        delete.topic.enable=true
        num.partitions=32
        default.replication.factor=2
        offsets.topic.replication.factor=2
        transaction.state.log.replication.factor=2
        transaction.state.log.min.isr=1
        ProducerConfig.RETRIES_CONFIG=10
        ConsumerConfig.MAX_POLL_INTERVAL_MS_CONFIG=1000
        session.timeout.ms=200000
        num.recovery.threads.per.data.dir=8
      storageConfigs:
        - mountPath: "/kafka-logs"
          pvcSpec:
            accessModes:
              - ReadWriteOnce
            storageClassName: ssd
            resources:
              requests:
                storage: 256Gi
...
```

**Expected behavior**
there should not be any error

**Additional context**
K8s cluster on AKS 1.14.5
```
kubectl version
[...]

PR #122, opened 2019-09-10
Clean up error handling in reconcile loops

tinyzimmer:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | addresses #104 

### What's in this PR?

~~This basically grabs:~~

~~- every instance of `return ctrl.Result{}, nil` and replaces it with a common `return reconciled()`~~
~~- every instance of `return ctrl.Result{}, err` and replaces it with a common `return logErrorAndRequeue(logger, message, err)`.~~

~~There is also a small cleanup in the new user/topic stuff where I was originally adding a field to a CR spec if it did not exist. On second look that seemed a poor design decision on my part and instead I just set the value I need in memory and leave the spec in tact.~~

This PR adds a new `errorfactory` package that can be used from "lower-level" operations to wrap errors in a check-able type along with the stacktraces and fields that `emperror` provides. 

Then in the top-level controller loops, the errors can be checked and dealt with depending on their type. For now in this PR, `ResourceNotReady` have the failed reconcile suppressed and are re-queued. But, the functionality could later be extended to other parts of the operator. 

### Why?

The `failed to reconcile resource` errors can be confusing for the passerby. This suppresses those when it's just a resource not being ready, in favor of a simple log of the error and then forcing a requeue without passing the original error.

~~The effect of this in practice is that instead of the `failed to reconcile resource` errors you'll only see the more common things like `x resource could not be found, may not be ready`. So hopefully it'll make more sense what is happening.~~

### Checklist

- [X] Implementation tested


controllers/controller_common.go
```diff
@@ -0,0 +1,29 @@
...
+package controllers
+
+import (
+	"github.com/go-logr/logr"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+func logErrorAndRequeue(logger logr.Logger, msg string, err error) (ctrl.Result, error) {
+	logger.Error(err, msg)
+	return ctrl.Result{Requeue: true}, nil
```
baluchicken: I think sadly this is not that easy. Returning `ctrl.Result{}, err` tells the controller to requeue  the request and set an exponential backoff strategy, on the other hand if we use `ctrl.Result{Requeue: true}, nil` it just tells the controller to requeue immediately. We need to distinguish these two. 
We should use `ctrl.Result{Requeue: true}, nil` when we are waiting for something which we know will occur. E.g.: a load balancer external IP provided. 
Although, we should use `ctrl.Result{}, err` when an unplanned error happens.

baluchicken: See example: https://github.com/banzaicloud/kafka-operator/blob/d4dfb484f223c0c1062123c0dd6d78e0e86961e9/controllers/kafkacluster_controller.go#L107

tinyzimmer: Makes sense, I'll try to clean it up later

PR #123, opened 2019-09-11
Implement cluster state and healthcheck for cluster

baluchicken:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

This PR adds a simple cluster state to the CR and implement a health check using sarama golang Kafka lib. Also it removes some code duplications around state update code

### Checklist

- [x] Implementation tested

### DIFF
--- a/controllers/kafkacluster_controller.go
+++ b/controllers/kafkacluster_controller.go
@@ -17,6 +17,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"time"
 
 	"emperror.dev/errors"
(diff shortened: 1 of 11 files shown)


controllers/kafkacluster_controller.go
```diff
@@ -42,8 +43,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
+	"github.com/banzaicloud/kafka-operator/api/v1alpha1"
```
pbalogh-sa: this package is already imported as banzaicloudv1alpha1

PR #126, opened 2019-09-16
Stop extra reconcile attempts on clusterissuers and bootstrap secret

tinyzimmer:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Two cleanups around original reconciling of a KafkaCluster with `sslSecrets.create: true`. ClusterIssuers were being defined with the cluster namespace, which the API was ignoring since they are cluster-scoped. But every reconcile attempt the operator would still touch the object because it thought the namespace changed. 

Same issue, different context, with the generated JKS password secret. It was a new value for the password every loop so the operator touched it every time. There might be a cleaner way to do this - but this makes sure we only create the secret once, and if it exists just assume it's ours. 

### Why?

Cleans up the logs mostly, you currently see a lot of resource-diffs that aren't actually diffs. 

### Checklist

- [X] Implementation tested

### DIFF
--- a/pkg/resources/pki/certs.go
+++ b/pkg/resources/pki/certs.go
@@ -26,6 +26,7 @@ import (
 	certv1 "github.com/jetstack/cert-manager/pkg/apis/certmanager/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


baluchicken:
LGTM the @tinyzimmer

PR #127, opened 2019-09-16
Add the ability to rolling upgrade cluster gracefully

baluchicken:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

This PR contains some refactor and the initial implementation of rolling upgrade

### Checklist

- [x] Implementation tested
- [] Error handling code meets the [guideline](https://github.com/banzaicloud/pipeline/blob/master/docs/error-handling-guide.md)
- [] User guide and development docs updated (if needed)

### DIFF
--- a/api/v1alpha1/kafkacluster_types.go
+++ b/api/v1alpha1/kafkacluster_types.go
@@ -233,7 +233,7 @@ func (cConfig *CruiseControlConfig) GetCCImage() string {
 	if cConfig.Image != "" {
 		return cConfig.Image
 	}
-	return "solsson/kafka-cruise-control@sha256:c70eae329b4ececba58e8cf4fa6e774dd2e0205988d8e5be1a70e622fcc46716"
+	return "solsson/kafka-cruise-control@sha256:f3f3775f3b5e2a5ae2da6fdae60ed118793ac32c80f13fba31be8f025a57f6ac"
 }
 
 // GetImage returns the used image for Prometheus JMX exporter
--- a/api/v1alpha1/kafkacluster_types.go
+++ b/api/v1alpha1/kafkacluster_types.go
@@ -192,12 +192,12 @@ func (bConfig *BrokerConfig) GetResources() *corev1.ResourceRequirements {
 	}
 	return &corev1.ResourceRequirements{
 		Limits: corev1.ResourceList{
-			"cpu":    resource.MustParse("500m"),
-			"memory": resource.MustParse("1Gi"),
+			"cpu":    resource.MustParse("1500m"),
+			"memory": resource.MustParse("3Gi"),
 		},
 		Requests: corev1.ResourceList{
-			"cpu":    resource.MustParse("200m"),
-			"memory": resource.MustParse("500Mi"),
+			"cpu":    resource.MustParse("1000m"),
+			"memory": resource.MustParse("2Gi"),
 		},
 	}
 }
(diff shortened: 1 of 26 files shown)


pkg/kafkautil/healthcheck.go
```diff
@@ -38,3 +40,32 @@ func (k *kafkaClient) OfflineReplicaCount() (int, error) {
...
+			replicas, err := k.client.Replicas(topic, partition)
+			if err != nil {
+				return false, err
+			}
+			isrreplicas, err := k.client.InSyncReplicas(topic, partition)
+			if err != nil {
+				return false, err
+			}
+			if len(replicas) != len(isrreplicas) {
+				log.Info("not all replicas are in sync")
```
pepov: possibly worth to gather all the out of sync replicas and log which are out of sync? for debugging/troubleshooting...

pkg/resources/kafka/kafka.go
```diff
@@ -245,7 +246,7 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
 				r.pvc,
```
pepov: is this inner for loop still required?

pkg/resources/kafka/kafka.go
```diff
@@ -299,9 +300,9 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
 			r.pod,
```
pepov: is this for loop still required?

pkg/kafkautil/client.go
```diff
@@ -73,13 +81,11 @@ func New(opts *KafkaConfig) (client KafkaClient, err error) {
 	}
 
 	if kclient.brokers, err = kclient.DescribeCluster(); err != nil {
-		kclient.admin.Close()
```
tinyzimmer: I think this is necessary still for garbage collection. If it got to the point where it is trying to `DescribeCluster`, it has already successfully opened an admin client. If that errs out then we want to make sure we close it I think.

api/v1alpha1/kafkacluster_types.go
```diff
@@ -192,12 +191,12 @@ func (bConfig *BrokerConfig) GetResources() *corev1.ResourceRequirements {
 	}
 	return &corev1.ResourceRequirements{
 		Limits: corev1.ResourceList{
-			"cpu":    resource.MustParse("500m"),
-			"memory": resource.MustParse("1Gi"),
+			"cpu":    resource.MustParse("1500m"),
```
bonifaido: I would make these resource requirements configurable through the CR. People usually have different usage patterns.

pkg/kafkautil/healthcheck.go
tinyzimmer: This is actually getting done by the kafkatopic CR itself too for whatever thats worth. At least if the topic was made through that. The status will be kept up to date and detail any out of sync replicas along with partition leaders and such.

api/v1alpha1/kafkacluster_types.go
baluchicken: You are absolutely right, but we are planning to deal with this on a separate PR. This one is getting huge

bonifaido: sure thing

pkg/kafkautil/client.go
baluchicken: Thanks, I just added this line

PR #129, opened 2019-09-19
Prometheus operator

### LINKED ISSUE #125: Bump Kafka-Operator helm chart to use Prometheus Operator
**Describe the bug*
In this PR we changed the label names used to get cr in case of an alert from `kubernetes_namespace` -> `namespace`
https://github.com/banzaicloud/kafka-operator/pull/96

This new labelling requires a Prometheus which was installed by the Prometheus Operator.
Replace the operator helm chart requirement to use the Prometheus Operator chart instead.

pbalogh-sa:
| Q               | A
| --------------- | ---
| Bug fix?        |yes
| New feature?    |yes
| API breaks?     | no
| Deprecations?   |yes
| Related tickets | fixes #125 

### What's in this PR?
Prometheus helm chart dependency was removed and an example which is using prometheus-operator was added instead.

### Additional changes:
Some changes were implemented for fixing some bugs in broker pods and cruise control deployment creation and alert processing.

### Checklist

- [x] Implementation tested


charts/kafka-operator/values.yaml
```diff
@@ -6,8 +6,8 @@ replicaCount: 1
 
 operator:
   image:
-    repository: banzaicloud/kafka-operator
-    tag: 0.5.4
+    repository: anzaicloud/kafka-operator
```
baluchicken:
```suggestion
    repository: banzaicloud/kafka-operator
```

internal/alertmanager/currentalert/current_alerts.go
```diff
@@ -119,10 +120,12 @@ func (a *currentAlerts) HandleAlert(alertFp model.Fingerprint, client client.Cli
 
 func (a *currentAlerts) GetRollingUpgradeAlertCount() int {
 	alertCount := 0
-	for _, alert := range a.alerts {
-		for key := range alert.Labels {
-			if key == "rollingupgrade" {
-				alertCount++
+	if len(a.alerts) > 0 {
```
baluchicken: We don't need this since `range a.alerts` will handle this behaviour

PR #132, opened 2019-09-19
Fix Storage Config

bernardoVale:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no|yes
| API breaks?     | no|yes
| Deprecations?   | no|yes

### What's in this PR?
This PR fix the configuration of kafka logs. 

### Why?
The current implementation is not configuring the `kafka.logs` property making the cluster use the default directory which is `/tmp/kafka-logs`. The line added was borrowed from master branch

### Checklist

- [ ] Implementation tested

### To Do

### DIFF
--- a/pkg/resources/kafka/configmap.go
+++ b/pkg/resources/kafka/configmap.go
@@ -55,7 +55,9 @@ metric.reporters=com.linkedin.kafka.cruisecontrol.metricsreporter.CruiseControlM
 cruise.control.metrics.reporter.bootstrap.servers={{ .CruiseControlBootstrapServers }}
 broker.id={{ .Id }}
 
-{{ .StorageConfig }}
+{{ if .StorageConfig }}
+log.dirs={{ .StorageConfig }}
+{{ end }}
 
 {{ .AdvertisedListenersConfig }}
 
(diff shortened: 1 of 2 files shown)


pkg/resources/kafka/configmap.go
```diff
@@ -55,7 +55,9 @@ metric.reporters=com.linkedin.kafka.cruisecontrol.metricsreporter.CruiseControlM
 cruise.control.metrics.reporter.bootstrap.servers={{ .CruiseControlBootstrapServers }}
 broker.id={{ .Id }}
 
-{{ .StorageConfig }}
+{{ if .StorageConfig }}
```
baluchicken: I would like to leave the {{ . StorageConfig }} and add the fix to the generator function.
https://github.com/banzaicloud/kafka-operator/blob/d8ef6afa8968fb02213afdf86e3c60f91def87a0/pkg/resources/kafka/configmap.go#L122
I think we should modify it to return `return fmt.Sprintf("log.dirs=%s\n", strings.Join(mountPaths, ","))`

bernardoVale: that was the original implementation. I decided to use the `if` statement because if you don't declare a `storageConfig` you will end up with an empty `log.dirs=` declaration.

If you think that's not a problem a can fix that. We'll also need to fix all tests to include the empty `log.dist=`

baluchicken: Thanks for the clarification, I get your point and you are right.

PR #150, opened 2019-09-30
Prune/imagePullSecrets

prune998:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | none

### What's in this PR?
This PR add support for setting a imagePullSecrets in the Helm Chart to deploy the Kafka Operator from a private repository.
The `imagePullSecrets` is added to the `ServiceAccount`

### Why?
Because Private Docker repo needs control...

### Additional context

Also, make use of the `replicaCount` value to set the number of replicas of the Operator.

### DIFF
--- a/charts/kafka-operator/values.yaml
+++ b/charts/kafka-operator/values.yaml
@@ -4,6 +4,10 @@
 
 replicaCount: 1
 
+# Lists the secrets you need to use to pull kafka-operator image from a private registry.
+imagePullSecrets: []
+  # - private-registry-key
+
 operator:
   image:
     repository: banzaicloud/kafka-operator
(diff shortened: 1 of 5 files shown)


charts/kafka-operator/values.yaml
```diff
@@ -4,6 +4,10 @@
 
 replicaCount: 1
 
+# Lists the secrets you need to use to pull Istio images from a private registry.
```
kozmagabor:
```suggestion
# Lists the secrets you need to use to pull kafka-operator image from a private registry.
```

PR #157, opened 2019-10-01
User/topic cleanup and stability improvements

### LINKED ISSUE #148: Create CC topic using the provided Topic CRD
An error can happen if Kafka is reusing zookeeper on the same path for a different Kafka Cluster.
Since the CC topic is created from code during the cluster creation, Zookeeper will hold the CC topic info even when the cluster is deleted, so creating a new cluster in the same zk znode will fail since the client which tries to create the topic fails with already exists error.

We are planning to use the newly introduced Topic CRD to create this topic, so the operator can remove it from zookeeper during cluster deletion.

tinyzimmer:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #148 

### What's in this PR?

This is mostly a round-up of things I wanted to clean up in the initial user/topic branch as well as a few enhancements. There is also the ability to use `vault` instead of `cert-manager` for the PKI and `KafkaUsers`. Doing this also adds support for using things like the `bank-vaults` webhook injector instead of Kubernetes Secrets for pod deployments accessing a `KafkaCluster` secured with SSL. 

I also add a crap load of unit tests and will probably eventually add more.

A full list of things tackled:

- Bump `cert-manager` version to 0.10.1 in README
- `KafkaTopic` CR status improvements
- Main PKI components are packaged off behind interface for easier development/mocking
- Allow for variable PKI backends with a `KafkaCluster` (defaults to cert-manager still)
- Add experimental support for using vault as a PKI backend
- Update helm chart to allow optional verbose logging/configure vault access
- Move CC topic to `KafkaTopic` CR
- Replace `pem-to-jks` init containers with prepared secret mounts made from `KafkaUsers` when using SSL
- Add optional ability to supply `dnsNames` with a `KafkaUser` to support the above
- Add `hack/kafka-test-pod` for verifying SSL deployments and the blog
- General GC/finalization improvements 
- General stability/readability improvements
- A lot of unit tests and some mock implementations

### Why?

The tests and more clearly defined `PKIManager` interface makes me feel a lot better overall about the `KafkaUser`/`KafkaTopic` stability. Getting rid of that SSL init container also significantly decreases the time to provision a new cluster with SSL. The `vault` support is because of moon.

The improved garbage collection and finalizations probably aid the developer more than the user right now. However, it essentially is now much more stable in terms of everything getting cleaned up properly when a `KafkaCluster` is deleted. 

### Additional context

This branch was started before the last release while the move to the `v1beta1` API was already underway. So it kind of just kept collecting fixes/improvements until I could pull and merge upstream 😛 . 

### Checklist

- [X] Implementation tested

### To Do
- [ ] Update `kustomize` templates for configuring vault access on the operator deployment


baluchicken:
@tinyzimmer Thanks for this huge PR I will start to test and check this ASAP.

pkg/pki/pki_manager.go
```diff
@@ -0,0 +1,87 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// MockBackend is used for mocking during testing
+var MockBackend = v1beta1.PKIBackend("mock")
+
+// GetPKIManager returns a PKI/User manager interface for a given cluster
+func GetPKIManager(client client.Client, cluster *v1beta1.KafkaCluster) pkicommon.PKIManager {
+	switch cluster.Spec.ListenersConfig.SSLSecrets.PKIBackend {
```
pbalogh-sa: If SSL isn't specified in KafkaCluster CR this will be a nil pointer dereference when the `GetPKIManager` function will be called in `checkFinalizers`.

pkg/errorfactory/errorfactory_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+func TestNew(t *testing.T) {
+	for _, errType := range errorTypes {
+		err := New(errType, errors.New("test-error"), "test-message")
+		expected := "test-message: test-error"
+		got := err.Error()
+		if got != expected {
+			t.Error("Expected:", expected, "got:", got)
+		}
+		if reflect.TypeOf(err) != reflect.TypeOf(errType) {
```
baluchicken: I think instead of reflect we should use errors.As

controllers/kafkacluster_controller.go
```diff
@@ -147,84 +142,167 @@ func (r *KafkaClusterReconciler) Reconcile(request ctrl.Request) (ctrl.Result, e
...
 		}
+		if cluster, err = r.removeFinalizer(ctx, cluster, clusterUsersFinalizer); err != nil {
+			return requeueWithError(log, "failed to remove users finalizer from kafkacluster", err)
+		}
+	}
+
+	// Do any necessary PKI cleanup - a PKI backend should make sure any
+	// user finalizations are done before it does its final cleanup
+	log.Info("Tearing down any PKI resources for the kafkacluster")
+	if err = pki.GetPKIManager(r.Client, cluster).FinalizePKI(ctx, log); err != nil {
```
pbalogh-sa: If SSL isn't specified in KafkaCluster CR this will be a nil pointer dereference when the `GetPKIManager` function will be called.

tinyzimmer: Fixing. I'm trying to decide if I want some sort of check in the `KafkaUser` controller also. Creating a `KafkaUser` doesn't make sense if you are not using SSL right now - but looking closer that would probably also panic if someone tried. 

Might be abother good use-case to expand the admission webhook

controllers/kafkacluster_controller.go
```diff
@@ -147,84 +142,167 @@ func (r *KafkaClusterReconciler) Reconcile(request ctrl.Request) (ctrl.Result, e
...
+		}
+	}
+
+	// If any of the topics still exist, it means their finalizer is still running.
+	// Wait to make sure we have fully cleaned up zookeeper. Also if we delete
+	// our kafkausers before all topics are finished cleaning up, we will lose
+	// our controller certificate.
+	log.Info("Ensuring all topics have finished cleaning up")
+	var childTopics v1alpha1.KafkaTopicList
+	if err = r.Client.List(ctx, &childTopics); err != nil {
```
baluchicken: If I saw that right, all the KafkaTopics are labeled with some cluster specific label, why don't we use List filters to get back only those which are relevant for us, with that I think we can remove the `if belongsToCluster` part.

controllers/kafkatopic_controller.go
```diff
@@ -240,108 +258,91 @@ func (r *KafkaTopicReconciler) syncTopicStatus(cluster *banzaicloudv1beta1.Kafka
...
 		}
 	}
 	return reconciled()
 }
 
-func (r *KafkaTopicReconciler) removeFinalizer(topic *banzaicloudv1alpha1.KafkaTopic) error {
+func (r *KafkaTopicReconciler) removeFinalizer(ctx context.Context, topic *v1alpha1.KafkaTopic) error {
 	topic.SetFinalizers(util.StringSliceRemove(topic.GetFinalizers(), topicFinalizer))
-	return r.Client.Update(context.TODO(), topic)
+	return r.Client.Update(ctx, topic)
```
baluchicken: we should set the typeMeta here as well, like
```
typeMeta := topic.typeMeta
r.Client.Update(ctx,topic)
topic.typeMeta = typeMeta
```
Update method mutates topic and will cause problems since update loses the typeMeta of the config that's used later when setting ownerrefs

controllers/kafkatopic_controller.go
```diff
@@ -189,23 +165,36 @@ func (r *KafkaTopicReconciler) Reconcile(request reconcile.Request) (reconcile.R
...
+		instance.SetFinalizers(append(instance.GetFinalizers(), topicFinalizer))
 	}
 
 	// push any changes
-	if err = r.Client.Update(context.TODO(), instance); err != nil {
-		return requeueWithError(reqLogger, "failed to update KafkaTopic with controller reference", err)
+	// TODO (tinyzimmer): This is sometimes failing the first reconcile attempt
+	// still with an "object already modified" error. It's benign, but should dig
+	// deeper at some point.
+	if err = r.Client.Update(ctx, instance); err != nil {
```
baluchicken: same here, please save the typeMeta and reapply it after update

controllers/kafkatopic_controller.go
```diff
@@ -189,23 +165,36 @@ func (r *KafkaTopicReconciler) Reconcile(request reconcile.Request) (reconcile.R
...
+		return requeueWithError(reqLogger, "failed to update KafkaTopic", err)
+	}
+
+	// Fetch the updated object
+	instance = &v1alpha1.KafkaTopic{}
```
baluchicken: This one is not required since update mutates the instance and updates the latest changes to it

pkg/resources/reconciler.go
```diff
@@ -40,6 +40,9 @@ type Resource func() runtime.Object
 // ResourceWithLogs function with log parameter
 type ResourceWithLogs func(log logr.Logger) runtime.Object
 
+// ResourceWithLogsAndClientPassowrd function with log and password parameter
+type ResourceWithLogsAndClientPassowrd func(log logr.Logger, clientPass string) runtime.Object
```
baluchicken: Typo in password

pkg/pki/certmanagerpki/certmanager_pki.go
```diff
@@ -0,0 +1,254 @@
...
+	caCert := secret.Data[v1alpha1.CACertKey]
+	caSecret := &corev1.Secret{
+		ObjectMeta: templates.ObjectMeta(fmt.Sprintf(pkicommon.BrokerCACertTemplate, cluster.Name), pkicommon.LabelsForKafkaPKI(cluster.Name), cluster),
+		Data: map[string][]byte{
+			v1alpha1.CoreCACertKey:  caCert,
+			corev1.TLSCertKey:       caCert,
+			corev1.TLSPrivateKeyKey: caKey,
+		},
+	}
+	controllerutil.SetControllerReference(cluster, caSecret, scheme)
```
baluchicken: Please handle error properly

pkg/pki/certmanagerpki/certmanager_pki.go
```diff
@@ -0,0 +1,254 @@
...
+	selfsignerMeta.Namespace = metav1.NamespaceAll
+	selfsigner := &certv1.ClusterIssuer{
+		ObjectMeta: selfsignerMeta,
+		Spec: certv1.IssuerSpec{
+			IssuerConfig: certv1.IssuerConfig{
+				SelfSigned: &certv1.SelfSignedIssuer{},
+			},
+		},
+	}
+	controllerutil.SetControllerReference(cluster, selfsigner, scheme)
```
baluchicken: Please handle error properly

pkg/pki/certmanagerpki/certmanager_pki.go
```diff
@@ -0,0 +1,254 @@
...
+			SecretName: fmt.Sprintf(pkicommon.BrokerCACertTemplate, cluster.Name),
+			CommonName: fmt.Sprintf(pkicommon.CAFQDNTemplate, cluster.Name, cluster.Namespace),
+			IsCA:       true,
+			IssuerRef: certv1.ObjectReference{
+				Name: fmt.Sprintf(pkicommon.BrokerSelfSignerTemplate, cluster.Name),
+				Kind: certv1.ClusterIssuerKind,
+			},
+		},
+	}
+	controllerutil.SetControllerReference(cluster, ca, scheme)
```
baluchicken: Please handle error properly

pkg/pki/certmanagerpki/certmanager_pki.go
```diff
@@ -0,0 +1,254 @@
...
+		ObjectMeta: clusterIssuerMeta,
+		Spec: certv1.IssuerSpec{
+			IssuerConfig: certv1.IssuerConfig{
+				CA: &certv1.CAIssuer{
+					SecretName: fmt.Sprintf(pkicommon.BrokerCACertTemplate, cluster.Name),
+				},
+			},
+		},
+	}
+	controllerutil.SetControllerReference(cluster, issuer, scheme)
```
baluchicken: Please handle error properly

pkg/pki/certmanagerpki/certmanager_user.go
```diff
@@ -0,0 +1,160 @@
...
+			IssuerRef: certv1.ObjectReference{
+				Name: caName,
+				Kind: caKind,
+			},
+		},
+	}
+	if user.Spec.DNSNames != nil && len(user.Spec.DNSNames) > 0 {
+		cert.Spec.DNSNames = user.Spec.DNSNames
+	}
+	controllerutil.SetControllerReference(user, cert, scheme)
```
baluchicken: Please handle error properly

pkg/pki/vaultpki/vault_pki.go
```diff
@@ -0,0 +1,287 @@
...
+				ObjectMeta: templates.ObjectMeta(fmt.Sprintf(pkicommon.BrokerControllerTemplate, v.cluster.Name), pkicommon.LabelsForKafkaPKI(v.cluster.Name), v.cluster),
+				Data: map[string][]byte{
+					corev1.TLSCertKey:       controllerCert.Certificate,
+					corev1.TLSPrivateKeyKey: controllerCert.Key,
+					v1alpha1.CoreCACertKey:  controllerCert.CA,
+					v1alpha1.TLSJKSKey:      controllerCert.JKS,
+					v1alpha1.PasswordKey:    controllerCert.Password,
+				},
+			}
+			controllerutil.SetControllerReference(v.cluster, clientCert, scheme)
```
baluchicken: Please handle error properly

pkg/pki/vaultpki/vault_pki.go
```diff
@@ -0,0 +1,287 @@
...
+				ObjectMeta: templates.ObjectMeta(fmt.Sprintf(pkicommon.BrokerServerCertTemplate, v.cluster.Name), pkicommon.LabelsForKafkaPKI(v.cluster.Name), v.cluster),
+				Data: map[string][]byte{
+					corev1.TLSCertKey:       brokerCert.Certificate,
+					corev1.TLSPrivateKeyKey: brokerCert.Key,
+					v1alpha1.CoreCACertKey:  brokerCert.CA,
+					v1alpha1.TLSJKSKey:      brokerCert.JKS,
+					v1alpha1.PasswordKey:    brokerCert.Password,
+				},
+			}
+			controllerutil.SetControllerReference(v.cluster, serverCert, scheme)
```
baluchicken: Please handle error properly

pkg/pki/vaultpki/vault_common.go
```diff
@@ -0,0 +1,335 @@
...
+// checkSecretPath checks if the provided path is a full vault path and
+// returns a modified path for the default backend if not
+func checkSecretPath(path string) string {
+	if len(strings.Split(path, "/")) == 1 {
+		return fmt.Sprintf("secret/%s", path)
+	}
+	return path
+}
+
+// getClient retrieves a vault client using the environment configuration
```
bonifaido: For this purpose (service account-based authentication) I would pull in the https://github.com/banzaicloud/bank-vaults/blob/master/pkg/sdk/vault/client.go#L106, it does it for you and renews the token automatically.

pkg/pki/vaultpki/vault_pki.go
```diff
@@ -0,0 +1,287 @@
...
+	return
+}
+
+func (v *vaultPKI) initVaultPKI(vault *vaultapi.Client) error {
+	caPath := v.getCAPath()
+	// intermediatePath := v.getIntermediatePath()
+	userPath := v.getUserStorePath()
+	var err error
+
+	// Setup a PKI mount for the ca
```
bonifaido: As a general comment for this PR (which is great in overall!) that I would leave out all Vault configuration parts from it because of these things require `sudo` Vault (admin) capabilities, for example manipulating mounts. I'm pretty sure that most people don't want to let the kafka-operator do things like that, so I would leave these parts to an exercise to the user, with vault-operator for example.

bonifaido: I'm happy to find out how to configure Vault with our vault-operator to make these PKI backends available.

pkg/pki/vaultpki/vault_common.go
tinyzimmer: Will take a look at that 😄

pkg/pki/vaultpki/vault_pki.go
tinyzimmer: This is done in the latest commit - I may need to update user docs more to reflect how vault needs to be setup. Also curious if there is a better way to handle passing vault's CA cert to the operator then what I'm doing right now in the helm chart with a secret/envvar.

PR #165, opened 2019-10-03
add exists check for __CruiseControlMetrics topic

rpahli:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?
after creating a new cluster I got error that the controller is not possible to create the __CruiseControlMetrics topic. After checking the boker logs I found out that the topic was already. I tryed to create new clusters multible times I got always the same error so I added a check if the topic already exists.

### Checklist

- [x] Implementation tested

### To Do

### DIFF
--- a/pkg/resources/cruisecontrol/topicManager.go
+++ b/pkg/resources/cruisecontrol/topicManager.go
@@ -32,6 +32,13 @@ func generateCCTopic(cluster *v1beta1.KafkaCluster, client client.Client, log lo
 	}
 	defer broker.Close()
 
+	topic, err := broker.GetTopic("__CruiseControlMetrics")
+
+	// topic exists
+	if err == nil && topic != nil {
+		return nil
+	}
+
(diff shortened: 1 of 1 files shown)


baluchicken:
Thanks @rpahli, for reporting it and thanks for this PR, we are aware of this problem. Just check #148 . We also have an open PR which fixes this behaviour, we are actively reviewing that, so stay tuned. Using KafkaTopic CR to create a Cruise Control topic is a better solution to this problem than  just checking if it is exists. Existing CC topic can be an error since, it can mean an already deleted Kafka Cluster left it's data present on ZK. Using that data in a new Kafka Cluster can cause errors.

baluchicken:
We just merged #157 which fixes this problem as well

baluchicken:
@rpahli we just released the new version of the operator 0.7.0 which tries to delete the `__CruiseControlMetrics` from ZK during shutdown.

PR #166, opened 2019-10-04
Add auto generated clients for the CRDs

nikore:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| API breaks?     | yes
| Deprecations?   | no

### What's in this PR?

I add the ability to generate the client code for the CRDs in this project, this did require moving the CRDs from `/api` to `/api/kafka` to conform to the tool requirements. I updated all imports to point to the new path. 

I recognize this looks like a large change, anything under `pkg/client` is auto generated.

### Why?

This is to make it easier to build tooling around these CRDs, namely to have a programmatic way to create CRDs. It will also allow for better testing. Typically this code and the `/api` code is in a different apimachinery repo. 

This client code is generated in other operators in banzaicloud and a fairly standard thing across operators. 

This change has no functional change other than the `api` package change. 

### Checklist
- [x] Implementation tested


pepov:
Hey @nikore have you considered using https://github.com/kubernetes-sigs/controller-runtime in your client project? Using `controller-runtime` and it's built-in informers/indexers and the client, you can consume the API in a generic way so that there is no need to generate a specific client. I beleive this is the reason why kubebuilder does not generate a client anymore.

nikore:
OK the `controller-runtime` does work but its more annoying to use and more annoying to test with. I get that `envtest.Environment` is better to test with, but it requires me to have the kubebuilder binary on anyone who wants to test my app and I have to make sure its in the CI image even though my project has no dependency on kubebuilder. 

Regardless, given that we can close this and make use of `controller-runtime` I'll miss the easy of use of the old generated clients.

pepov:
I didn't mean to close this out so fast, but instead to talk about the proposed changes in more detail. What does `client-gen` give you that you will miss so much? The fake implementations for the API types?

PR #203, opened 2019-11-18
Fix nil pointer dereference since the 0.11 cert manager does not popu…

baluchicken:
…late all secrets during creation

| Q               | A
| --------------- | ---
| Bug fix?        | yes|
| New feature?    | no|
| API breaks?     | no|
| Deprecations?   | no|

### What's in this PR?

This PR fixes a nil pointer dereference in the operator since, cert manager 0.11 does not populate 
secrets immediately. It creates a secret with empty values and updates it. Operator now checks if all required values are set in the secret and proceed accordingly.

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/pki/certmanagerpki/certmanager_user.go
+++ b/pkg/pki/certmanagerpki/certmanager_user.go
@@ -16,6 +16,7 @@ package certmanagerpki
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/banzaicloud/kafka-operator/api/v1alpha1"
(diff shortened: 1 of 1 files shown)


pkg/pki/certmanagerpki/certmanager_user.go
```diff
@@ -16,7 +16,9 @@ package certmanagerpki
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"github.com/go-logr/logr"
```
bonifaido: Unused import maybe?

baluchicken: Thanks, fixed.

PR #210, opened 2019-11-21
remove per-broker dns names and append wildcards to cluster dns names

tinyzimmer:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | n/a

### What's in this PR?

I didn't see a ticket for it, but @baluchicken brought it up on slack. There was an issue with cluster scaling and the DNS names for the broker certificates not including the new kafka nodes.

This PR gets rid of `brokerDNSNames` and instead adds a wildcard for each of the `clusterDNSName` formats. A quick test locally worked along with being able to scale the cluster and have a new pod join cleanly. On an existing cluster, it might trigger a rolling upgrade with new certificates, but we should probably verify that. 

### Checklist

- [X] Implementation tested

### DIFF
--- a/pkg/util/pki/common.go
+++ b/pkg/util/pki/common.go
@@ -122,7 +122,6 @@ func clusterDNSNames(cluster *v1beta1.KafkaCluster) (names []string) {
 		)
 
 		// service name only
-		// TODO (tinyzimmer): I'm not sure if this one is required
 		names = append(names,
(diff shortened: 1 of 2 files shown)


baluchicken:
Huge THX @tinyzimmer

PR #213, opened 2019-11-22
Check CC before process alerts

### LINKED ISSUE #199: Stop creating new pods based on alerts if CC is down.
When the CC is down and the kafka-operator reacts on alert, the newly created broker isn't put properly to the cluster. In this case, the number of pods will be flapping.

I think it would be better if we turn off the reactions on alert when the CC will be down.

pbalogh-sa:
Signed-off-by: Peter Balogh <p.balogh.sa@gmail.com>

| Q               | A
| --------------- | ---
| Bug fix?        |yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #199

### What's in this PR?
This PR fixes the issue of flapping brokers when the Cruise Control isn't reachable.

### DIFF
--- a/internal/alertmanager/currentalert/process.go
+++ b/internal/alertmanager/currentalert/process.go
@@ -25,12 +25,35 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func examineAlert(alert *currentAlertStruct, client client.Client, rollingUpgradeAlertCount int) error {
+type ccConfig struct {
+	Name                  string
+	Namespace             string
+	CruiseControlEndpoint string
+}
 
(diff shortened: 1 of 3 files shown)


internal/alertmanager/currentalert/process.go
```diff
@@ -141,3 +164,10 @@ func upScale(labels model.LabelSet, annotations model.LabelSet, client client.Cl
 	}
 	return nil
 }
+
+func (c *ccConfig) getCruiseControlStatus() error {
+	if err := scale.GetCruiseControlStatus(c.Namespace, c.CruiseControlEndpoint, c.Name); err != nil {
```
baluchicken: This can be simplified to `return scale.GetCruiseControlStatus(c.Namespace, c.CruiseControlEndpoint, c.Name)`

PR #233, opened 2019-12-12
Merge labels from KafkaCluster resource to child objects

sebdoido:
| Q               | A
| --------------- | ---
| Bug fix?        | no|yes
| New feature?    | no|yes
| API breaks?     | no|yes
| Deprecations?   | no|yes

### What's in this PR?

Update util.MergeLabels to a variadic Function to merge multiple map of labels.

### Why?

It could be great that "child" objects (eg. pods, services, cm, pvc) inherit labels from KafkaCluster resource.

### Additional context

This docker image is available if you want to test the feature
`sebdoido/kafka-operator:0.8.1-4`

```
helm upgrade -i kafka-operator --namespace=dbre banzaicloud-stable/kafka-operator --set operator.image.repository=sebdoido/kafka-operator --set operator.image.tag=0.8.1-4
```
### Checklist

- [ ] Implementation tested

### To Do


baluchicken:
We really like the idea that the child objects can inherit labels from the KafkaCluster resource. Although we think it should be optional. We recommend to add a new field to the `kafkacluster_types` something like `ClusterLabels` and if that one is set the operator should place the labels in all resources. What do you think?

sebdoido:
Thanks for your comment.
You're right, i'm going to continue to work on this PR on that way, adding a flag to feature-flip this behaviour.

sebdoido:
Hi there, 
image `sebdoido/kafka-operator:0.8.1-5` is available if you want to test the latest commit.

Need to add this flag to your manifest `propagateLabels: true` 
Example:
```
config/samples/simplekafkacluster.yaml
11:  propagateLabels: true
```

tinyzimmer:
From a quick look, it appears you need to do a `make generate manifests` and commit the result.

The `make manifests` will update the CRD in `config/base/crds`, and it will have to be manually added to the copy in the helm chart.

sebdoido:
Thanks, i updated the pr

api/v1beta1/kafkacluster_types.go
```diff
@@ -36,6 +36,7 @@ type KafkaClusterSpec struct {
 	Brokers                []Broker                `json:"brokers"`
 	RollingUpgradeConfig   RollingUpgradeConfig    `json:"rollingUpgradeConfig"`
 	OneBrokerPerNode       bool                    `json:"oneBrokerPerNode"`
+	PropagateLabels        bool                    `json:"propagateLabels"`
```
tinyzimmer: I think we should add an `omitempty` tag here. So it defaults to `false` without bothering the user.  Then the next `make manifests` will remove it from the required fields.

sebdoido:
Thanks, good catch

sebdoido:
`I see a few other places where we should probably do that too` 
> Let me know if you want to I complete some stuff through this PR ?

tinyzimmer:
Don't worry about that, for this I'd say just make sure the helm chart is updated.

sebdoido:
Hi there
any chance to get this PR merged? 
Thanks

baluchicken:
Sure will check it today

baluchicken:
Thanks for the contribution @sebdoido

PR #243, opened 2020-01-07
Add downScaleLimit and upScaleLimit

pbalogh-sa:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes

### What's in this PR?
This PR implement min/max broker count

### Why?
With this feature, the actions which are based on alerts check broker count before scale up and down the cluster.


config/base/crds/kafka.banzaicloud.io_kafkaclusters.yaml
```diff
@@ -32,6 +32,17 @@ spec:
         spec:
           description: KafkaClusterSpec defines the desired state of KafkaCluster
           properties:
+            alertManagerConfig:
+              description: AlertManagerConfig defines configuration for alert manager
+              properties:
+                maxBrokerCount:
```
stoader: Should we name it `upScaleLimit` and `downScaleLimit`. Also can we add description here for each field?

`upScaleLimit` - The upper broker count limit the cluster can be upscaled  to.  If the number of brokers is above this limit than auto upscaling triggered by alerts is disabled.

`downScaleLimit` - The lower broker count limit the cluster can be downscaled  to.  If the number of brokers is below this limit than auto downscaling triggered by alerts is disabled.

pbalogh-sa: I agree with you. I'll replace them.

internal/alertmanager/currentalert/process.go
```diff
@@ -93,17 +95,15 @@ func (e *examiner) processAlert(ds disableScaling) error {
 		}
 	case "downScale":
 		if ds.Down {
-			e.Log.Info("downscale is skipped due to minimum broker count")
-			return nil
+			return errors.New("downscaling is skipped due to downscale limit")
```
stoader: why do we return with error instead of just simply logging the fact the downscale is skipped? Not performing downscale when it's disabled it's not an error.

internal/alertmanager/currentalert/process.go
```diff
@@ -93,17 +95,15 @@ func (e *examiner) processAlert(ds disableScaling) error {
...
 		}
 		err := downScale(e.Alert.Labels, e.Client)
 		if err != nil {
 			return err
 		}
 	case "upScale":
 		if ds.Up {
-			e.Log.Info("upscale is skipped due to maximum broker count")
-			return nil
+			return errors.New("upscaling is skipped due to upscale limit")
```
stoader: why do we return with error instead of just simply logging the fact the upscale is skipped? Not performing upscale when it's disabled it's not an error.

internal/alertmanager/currentalert/process.go
pbalogh-sa: If we just log the skipping, the alert will be marked as processed because of the handler checks error returned by `examineAlert`.
https://github.com/banzaicloud/kafka-operator/blob/feature/min-max-broker/internal/alertmanager/currentalert/current_alerts.go#L129-L133
In that case, we log the skipping as info, the `examineAlert` should return the skipping state as well.

internal/alertmanager/currentalert/process.go
pbalogh-sa: https://github.com/banzaicloud/kafka-operator/pull/243#discussion_r364087820

internal/alertmanager/currentalert/current_alerts.go
```diff
@@ -102,18 +116,26 @@ func (a *currentAlerts) AlertGC(alert AlertState) error {
...
 	}
 	if a.alerts[alertFp].Processed != true {
-		err := examineAlert(a.alerts[alertFp], client, rollingUpgradeAlertCount)
+		e := &examiner{
+			Alert:          a.alerts[alertFp],
+			Client:         client,
+			IgnoreCCStatus: a.IgnoreCCStatus,
+			Log:            log,
+		}
+		skipProcess, err := e.examineAlert(rollingUpgradeAlertCount)
```
stoader:
```suggestion
		alertProcessed, err := e.examineAlert(rollingUpgradeAlertCount)
```

internal/alertmanager/currentalert/current_alerts.go
```diff
@@ -102,18 +116,26 @@ func (a *currentAlerts) AlertGC(alert AlertState) error {
...
 		if err != nil {
 			return nil, err
 		}
-		a.alerts[alertFp].Processed = true
+		if !skipProcess {
```
stoader:
```suggestion
		if ! alertProcessed {
```

charts/kafka-operator/templates/operator-kafka-crd.yaml
```diff
@@ -36,6 +36,18 @@ spec:
         spec:
           description: KafkaClusterSpec defines the desired state of KafkaCluster
           properties:
+            alertManagerConfig:
+              description: AlertManagerConfig defines configuration for alert manager
+              properties:
+                downScaleLimit:
+                  description: DownScaleLimit defines the minimum number of brokers,
```
stoader:
```suggestion
                  description: DownScaleLimit the limit for auto-downscaling the Kafka cluster. Once the size of the cluster (number of brokers) reaches or falls below this limit the auto-downscaling triggered by alerts is disabled until the cluster size exceeds this limit.  This limit is not enforced if this field is omitted or is <= 0.
```

charts/kafka-operator/templates/operator-kafka-crd.yaml
```diff
@@ -36,6 +36,18 @@ spec:
...
+                    the alertmanager doesn't scale down the brokers below the number
```
stoader:
```suggestion
                    
```

charts/kafka-operator/templates/operator-kafka-crd.yaml
```diff
@@ -36,6 +36,18 @@ spec:
...
+                  type: integer
+                upScaleLimit:
+                  description: UpScaleLimit defines the maximum number of brokers,
```
stoader:
```suggestion
                  description: UpScaleLimit the limit for auto-upscaling the Kafka cluster. Once the size of the cluster (number of brokers) reaches or exceeds this limit the auto-upscaling triggered by alerts is disabled until the cluster size falls below this limit.  This limit is not enforced if this field is omitted or is <= 0.
```

charts/kafka-operator/templates/operator-kafka-crd.yaml
```diff
@@ -36,6 +36,18 @@ spec:
...
+                    the alertmanager doesn't scale up the brokers above the number
```
stoader:
```suggestion
                    
```

PR #275, opened 2020-01-30
Use struct instead of map[string]interface and casting

baluchicken:
| Q               | A
| --------------- | ---
| Bug fix?        | yes|
| New feature?    | no|
| API breaks?     | no|
| Deprecations?   | no|

### What's in this PR?

This PR removes the unnecessary type casts in scale.go and uses structs instead.
Also it fixes the CheckIfCCTaskFinished function bad behaviour if a not known user id is used.

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/scale/scale.go
+++ b/pkg/scale/scale.go
@@ -112,13 +112,17 @@ func GetCruiseControlStatus(namespace, ccEndpoint, clusterName string) error {
 		return err
 	}
 
-	var response map[string]interface{}
+	var response struct {
+		AnalyzerState struct {
+			IsProposalReady bool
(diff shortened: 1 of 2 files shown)


pkg/scale/scale.go
```diff
@@ -150,7 +154,12 @@ func isKafkaBrokerReady(brokerId, namespace, ccEndpoint, clusterName string) (bo
 		return running, err
 	}
 
-	var response map[string]interface{}
+	var response struct {
```
stoader: I'd add a unit test for parsing an sample CC response into the typed struct

PR #277, opened 2020-01-30
Kafka downscale reconcile should search for broker pods only

amuraru:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Currently when the user initiates a downscale by removing broker ids from spec
the operator will also select the cruise-control pod as a pod delete candidate.
Limiting the filter to broker pods only.

### Why?
Avoid unnecessary invalid empty `remove_broker` calls to cruise-control.  

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/resources/kafka/kafka.go
+++ b/pkg/resources/kafka/kafka.go
@@ -158,9 +158,7 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
 	}
 	// Handle Pod delete
 	podList := &corev1.PodList{}
-	matchingLabels := client.MatchingLabels{
-		"kafka_cr": r.KafkaCluster.Name,
-	}
+	matchingLabels := client.MatchingLabels(labelsForKafka(r.KafkaCluster.Name))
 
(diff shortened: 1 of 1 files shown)


pkg/resources/kafka/kafka.go
```diff
@@ -160,6 +160,7 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
 	podList := &corev1.PodList{}
 	matchingLabels := client.MatchingLabels{
 		"kafka_cr": r.KafkaCluster.Name,
+		"app":      "kafka",
```
stoader: I think rather `labelsForKafka(r.KafkaCluster.Name)` should used here as `labelsForKafka(r.KafkaCluster.Name)` is used when the Pod is created: https://github.com/banzaicloud/kafka-operator/blob/master/pkg/resources/kafka/pod.go#L88

amuraru: yup - better, fixed!

PR #285, opened 2020-02-05
enhancement: Configurable resource pattern type of KafkaUser ACLs.

### LINKED ISSUE #269: Assign permissions to KafkaUser based on topicName regex match

Currently, the `topicGrants` for a `KafkaUser` need to assigned to each `topicName` explicitly. There is no method to assign a permission to a range of topics base on `topicName` regex match. This problem brings much inconvenience for user management. 

For example, I want to create a KafkaUser `test` that has write permission to all topics whose name has a `test` prefix.

```yaml
apiVersion: kafka.banzaicloud.io/v1alpha1
kind: KafkaUser
metadata:
  name: test
spec:
  clusterRef:
    name: kafka
  secretName: test
  topicGrants:
    - topicName: test.*
      accessType: write
```

So I don't need to change the configuration of the `KafkaUser` when I add a new topic with `test` prefix. Permissions will be automatically granted to the user.

Simple wildcard match is also welcome if regex match is not possible.

**Additional context**

YanzheL:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #269 

### What's in this PR?

Kafka 2.x supports [four types of `ResourcePatternType`](https://kafka.apache.org/20/javadoc/org/apache/kafka/common/resource/PatternType.html) when configuring user ACLs.
Currently, the `topicGrants` property in `KafkaUser` CRD requires user to provide `topicName` explicitly, because it uses `LITERAL` pattern type by default.

This PR makes the User ACL pattern type configurable. User can use other pattern types.

In the following example, the `prefixed` pattern types enables user to assign permissions on any topics with a `test.` prefix.
```yaml
apiVersion: kafka.banzaicloud.io/v1alpha1
kind: KafkaUser
...
spec:
  ...
  topicGrants:
    - topicName: test.
      accessType: write
      patternType: prefixed
```
`literal` will be default if `patternType` is not provided.

The detailed behavior and pattern syntax of these pattern types depends on kafka.

### Additional context

Actually, I don't understand the behavior of `MATCH` pattern type. According to [Kafka Documentation](https://kafka.apache.org/20/javadoc/org/apache/kafka/common/resource/PatternType.html#MATCH), it says:

>public static final PatternType MATCH
In a filter, will perform pattern matching. e.g. Given a filter of ResourcePatternFilter(TOPIC, "payments.received", MATCH)`, the filter match any ResourcePattern that matches topic 'payments.received'. This might include:
A Literal pattern with the same type and name, e.g. ResourcePattern(TOPIC, "payments.received", LITERAL)
A Wildcard pattern with the same type, e.g. ResourcePattern(TOPIC, "*", LITERAL)
A Prefixed pattern with the same type and where the name is a matching prefix, e.g. ResourcePattern(TOPIC, "payments.", PREFIXED)

This sounds counterintuitive. It's supposed to use regex matching when pattern type is `MATCH`, but it doesn't. Actually it matches other ACLs that will match this "payments.received".

As for `ANY`, it's also confusing. What does ["In a filter, matches any resource pattern type."](https://kafka.apache.org/20/javadoc/org/apache/kafka/common/resource/PatternType.html#ANY) mean? It means when I use `ANY` while deleting ACLs for a topic, it will delete all ACLs of the topic no matter what pattern type it has? So what about the behavior of creating ACLs with `ANY` type?

### Checklist

- [x] Implementation tested

### To Do

### DIFF
--- a/api/v1alpha1/kafkauser_types.go
+++ b/api/v1alpha1/kafkauser_types.go
@@ -33,6 +33,8 @@ type UserTopicGrant struct {
 	TopicName string `json:"topicName"`
 	// +kubebuilder:validation:Enum={"read","write"}
 	AccessType KafkaAccessType `json:"accessType"`
+	// +kubebuilder:validation:Enum={"literal","match","prefixed","any"}
+	PatternType KafkaPatternType `json:"patternType"`
 }
 
 // KafkaUserStatus defines the observed state of KafkaUser
--- a/api/v1alpha1/common_types.go
+++ b/api/v1alpha1/common_types.go
@@ -17,6 +17,9 @@ package v1alpha1
 // KafkaAccessType hold info about Kafka ACL
 type KafkaAccessType string
 
+// KafkaPatternType hold the Resource Pattern Type of kafka ACL
+type KafkaPatternType string
+
 // TopicState defines the state of a KafkaTopic
 type TopicState string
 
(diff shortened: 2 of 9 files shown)


baluchicken:
Thanks will check this soon.

api/v1alpha1/kafkauser_types.go
```diff
@@ -33,6 +33,8 @@ type UserTopicGrant struct {
 	TopicName string `json:"topicName"`
 	// +kubebuilder:validation:Enum={"read","write"}
 	AccessType KafkaAccessType `json:"accessType"`
+	// +kubebuilder:validation:Enum={"literal","match","prefixed","any"}
+	PatternType KafkaPatternType `json:"patternType"`
```
baluchicken: Can we make this omitempty, to maintain backward compatibility.

baluchicken:
Just a minor thing otherwise looks good. Thanks @YanzheL

baluchicken:
Will fix that omitempty in a different PR.

PR #293, opened 2020-02-10
Do not block on pending brokers during upscale

amuraru:
This fixes the following scenario:
1. Start with a 3 racks cluster (3 brokers per rack)
2. One AZ goes completely down
  - existing brokers on that AZ are lost/failed
  - new brokers/pod cannot be scheduled on that AZ

At this point, we assume topics in Kafka are still healthy (replication factor > 1 + rack-aware balanced replicas)
but under-replicated (offline partitions on lost AZ)

One of the possible actions now is to trigger an upscale in the other 2 healthy AZs with more brokers
to take the load and move offline replicas

3. This patch ensures that `Pending` broker pods are not blocking bootstrapping other pods that can be scheduled.
And not stuck serially on each pod to be scheduled.

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### Checklist

- [x] Implementation tested


baluchicken:
Fixed with #294

PR #294, opened 2020-02-10
Decommission only registered brokers in CC during downscale

amuraru:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Depends on https://github.com/banzaicloud/kafka-operator/pull/293

When downscale is initiated for brokers that are not yet part of Kafka cluster
we avoid now decommissioning the broker in CC (i.e. `remove_broker`) to avoid deadlock (CC returns 5XX when unknown brokers are removed)

Also - downscaling brokers not fully registered may miss ConfigMap
so we skip missing ConfigMaps in that case.

### Why?

When downscaling not-yet fully registered brokers in Cruise Control (this can happen when pod cannot be scheduled for example due to faulty az, etc) operator issues a invalid `remove_broker` request.

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/resources/kafka/kafka.go
+++ b/pkg/resources/kafka/kafka.go
@@ -525,7 +525,9 @@ func (r *Reconciler) reconcileKafkaPod(log logr.Logger, desiredPod *corev1.Pod) 
 		currentPod = podList.Items[0].DeepCopy()
 		brokerId := currentPod.Labels["brokerId"]
 		if _, ok := r.KafkaCluster.Status.BrokersState[brokerId]; ok {
-			if r.KafkaCluster.Spec.RackAwareness != nil {
+			if currentPod.Spec.NodeName == "" {
+				log.Info(fmt.Sprintf("pod for brokerId %s does not scheduled to node yet", brokerId))
+			} else if r.KafkaCluster.Spec.RackAwareness != nil {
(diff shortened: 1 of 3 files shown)


pkg/resources/kafka/kafka.go
```diff
@@ -213,7 +224,12 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
 			}
 			err = r.Client.Delete(context.TODO(), &corev1.ConfigMap{ObjectMeta: templates.ObjectMeta(fmt.Sprintf(brokerConfigTemplate+"-%s", r.KafkaCluster.Name, broker.Labels["brokerId"]), labelsForKafka(r.KafkaCluster.Name), r.KafkaCluster)})
 			if err != nil {
-				return errors.WrapIfWithDetails(err, "could not delete configmap for broker", "id", broker.Labels["brokerId"])
+				if apierrors.IsNotFound(err) {
```
amuraru: the same may apply to other broker pod dependent objects? `Service`, `Volumes/PVCs`

baluchicken: Yes I am in the middle of the review and that is what I thought that if CM affected then the service and the PVC as well I guess.

baluchicken:
This also contains #293 so I will close that once this goes in.

PR #295, opened 2020-02-11
Add pending pod health check during rolling upgrade

baluchicken:
| Q               | A
| --------------- | ---
| Bug fix?        | no|
| New feature?    | yes|
| API breaks?     | no|
| Deprecations?   | no|

### What's in this PR?

This PR adds an extra check to pod deletion during rolling upgrade. 
Now if there is any pending pod in the kafka cluster, the rolling upgrade won't proceed.

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/k8sutil/resource.go
+++ b/pkg/k8sutil/resource.go
@@ -158,3 +158,21 @@ func CheckIfObjectUpdated(log logr.Logger, desiredType reflect.Type, current, de
 		return true
 	}
 }
+
+func IsPodContainsTerminatedContainer(pod *corev1.Pod) bool {
+	for _, containerState := range pod.Status.ContainerStatuses {
+		if containerState.State.Terminated != nil {
+			return true
+		}
+	}
+	return false
+}
+
+func IsPodContainsPendingContainer(pod *corev1.Pod) bool {
+	for _, containerState := range pod.Status.ContainerStatuses {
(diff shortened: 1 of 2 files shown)


amuraru:
LGTM +1

pkg/k8sutil/resource.go
```diff
@@ -158,3 +158,25 @@ func CheckIfObjectUpdated(log logr.Logger, desiredType reflect.Type, current, de
 		return true
 	}
 }
+
+func IsPodContainsTerminatedContainer(pod *corev1.Pod) bool {
+	contains := false
+	for _, containerState := range pod.Status.ContainerStatuses {
+		if containerState.State.Terminated != nil {
+			contains = true
```
hi-im-aren: Just return true here without the break and return false at the end of the func?

pkg/k8sutil/resource.go
```diff
@@ -158,3 +158,25 @@ func CheckIfObjectUpdated(log logr.Logger, desiredType reflect.Type, current, de
...
+		}
+	}
+	return contains
+}
+
+func IsPodContainsPendingContainer(pod *corev1.Pod) bool {
+	contains := false
+	for _, containerState := range pod.Status.ContainerStatuses {
+		if containerState.State.Waiting != nil {
+			contains = true
```
hi-im-aren: Same as above

PR #299, opened 2020-02-13
Add rebalance after new pvc was added, add new cruisecontrol controller

hi-im-aren:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

[NEW] New Cruise Control controller is added that will handle cruise control related tasks.
[NEW] Rebalance request to Cruise Control after a new PVC is mounted.
[FIX] Do not mount new PVC if the node cannot mount more. 

### Why?
Even if we add a PVC when an alert is firing because of a PVC not having enough space, without a rebalance the alert will keep firing.
The new controller will separate concerns from the existing controllers.

### Checklist

- [X] Implementation tested


controllers/cruisecontrol_controller.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+
+	return nil
+}
+func (r *Reconciler) handlePodAddCCTask(brokerId string, brokerState kafkav1beta1.BrokerState, log logr.Logger) error {
+	podList := &corev1.PodList{}
+
+	matchingLabels := client.MatchingLabels(
+		util.MergeLabels(
+			map[string]string{"app": "kafka", "kafka_cr": r.KafkaCluster.Name},
```
stoader: Use `kafka.labelsForKafka(r.KafkaCluster.Name)`

controllers/cruisecontrol_controller.go
```diff
@@ -0,0 +1,364 @@
...
+	return nil
+}
+
+func (r *Reconciler) checkCCTaskState(brokerIds []string, brokerState v1beta1.BrokerState, cruiseControlState v1beta1.CruiseControlState, log logr.Logger) error {
+
+	// check cc task status
+	status, err := scale.GetCCTaskState(brokerState.GracefulActionState.CruiseControlTaskId,
+		r.KafkaCluster.Namespace, r.KafkaCluster.Spec.CruiseControlConfig.CruiseControlEndpoint, r.KafkaCluster.Name)
+	if err != nil {
+		log.Info(fmt.Sprintf("Cruise control communication error checking running task: %s", brokerState.GracefulActionState.CruiseControlTaskId))
```
stoader:
```suggestion
		log.Info("Cruise control communication error checking running task", "taskId", brokerState.GracefulActionState.CruiseControlTaskId)
```

pkg/resources/kafka/kafka_test.go
```diff
@@ -49,13 +49,13 @@ func TestGetBrokersWithPendingOrRunningCCTask(t *testing.T) {
 							GracefulActionState: v1beta1.GracefulActionState{CruiseControlState: v1beta1.GracefulUpscaleSucceeded},
 						},
 						"1": {
-							GracefulActionState: v1beta1.GracefulActionState{CruiseControlState: v1beta1.GracefulUpdateFailed},
+							GracefulActionState: v1beta1.GracefulActionState{CruiseControlState: v1beta1.GracefulUpscaleFailed},
```
stoader: add test cases that covers downscale and rebalance actions

pkg/scale/scale.go
```diff
@@ -108,6 +108,46 @@ func parseCCErrorFromResp(input io.Reader) (string, error) {
...
 }
 
+func isKafkaBrokerDiskReady(brokerId, mountPath, namespace, ccEndpoint, clusterName string) (bool, error) {
+	options := map[string]string{
+		"json": "true",
+	}
+
+	rsp, err := getCruiseControl(kafkaClusterStateAction, namespace, options, ccEndpoint, clusterName)
+	if err != nil {
+		log.Error(err, "can't work with cruise-control because it is not ready")
```
stoader: add context e.g.
```suggestion
               keyVals :=[]interface{}{
                 "namespace", namespace,
                 "clusterName", clusterName,
                 "brokerId", brokerId,
                 "path", mountPath, 
               }
		log.Error(err, "can't check if broker disk is ready as Cruise Control not ready", keyVals...)
```

controllers/cruisecontrol_controller.go
```diff
@@ -0,0 +1,420 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
+
+	kafkav1beta1 "github.com/banzaicloud/kafka-operator/api/v1beta1"
+)
+
+// CruiseControlReconciler reconciles a kafka cluster object
+type CruiseControlReconciler struct {
```
amuraru: Would it make sense to rename this to `CruiseControlTasksReconciler`?
It reconciles tasks in CC not CC itself

pkg/resources/kafka/kafka.go
```diff
@@ -359,6 +258,128 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
 	return nil
 }
 
+func (r *Reconciler) reconcileKafkaPodDelete(log logr.Logger) error {
```
amuraru: I would rename this "handleBrokersDownscale" as it's checks the  brokers list in the spec.
I am thinking we can add in a future PR a check here to limit the dowscale under a threshold also

pkg/resources/cruisecontrol/configmap.go
```diff
@@ -92,3 +94,86 @@ func generateBootstrapServer(headlessEnabled bool, clusterName string) string {
...
+
+func parseMountPathWithSize(brokerConfigGroup v1beta1.BrokerConfig, log logr.Logger, brokerState v1beta1.Broker, brokerDisks map[string]string) {
+	for _, storageConfig := range brokerConfigGroup.StorageConfigs {
+		int64Value, isConvertible := util.QuantityPointer(storageConfig.PvcSpec.Resources.Requests["storage"]).AsInt64()
+		if isConvertible == false {
+			log.Info("Could not convert 'storage' quantity to Int64 in brokerConfig for broker",
+				"brokerId", brokerState.Id)
+		}
+
+		brokerDisks[storageConfig.MountPath] = strconv.FormatInt(int64Value, 10)
```
amuraru: In https://github.com/banzaicloud/kafka-operator/blob/376d13ac9336c739cc793a3b9375e311070d6802/pkg/resources/kafka/configmap.go "/kafka" is appended to mountpath when building kafka log.dirs. CC is expecting exact path in capacity config iiuc. I think the suffix needs to be added here as well

hi-im-aren: Thanks, good catch. In the future it would be great to move this into a single common place so we would avoid this bug in the future when we are working with mountpaths.

pkg/resources/kafka/kafka.go
baluchicken: Good idea, we are going to address this on a different commit, since all our reconcile method name follows this pattern, we are going to rename all to handleBrokersUpscale,Downscale

PR #305, opened 2020-02-19
Add support to configure operator to watch resources in a set of namespaces

### LINKED ISSUE #300: Add operator support to listen for CR in certain namespace
Currently, Kafka-operator has no support to listen for CR in certain namespace(s).
This is a required if you have a single Kubernetes cluster that is shared among multiple installs (e.g.: use a single non-prod Kubernetes cluster to install multiple Kafka clusters and operator to test different scenarios at the same time).

Add the config in the Kafka-operator to allow listening to a specific set of namespace, similar to `WATCH_NAMESPACE` property.

We might be able to use Kubernetes RBAC to limit access to specific namespaces, but this is a cumbersome workaround compared to `WATCH_NAMESPACE` option.

amuraru:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #300

### What's in this PR?
In constrained environments where operator cannot
be granted cluster level roles, users can configure
the operator to watch CRs only in specific set of
configurable k8s namespaces.
In this scenario, users can replace the default
ClusterRole and ClusterRoleBinding to Role and RoleBinding respectively
For further information see the kubernetes documentation about
Using [RBAC Authorization](https://kubernetes.io/docs/reference/access-authn-authz/rbac/).

Admission webhooks can also be disabled in these environments as
they require more rights than the user can grant to operator

### Why?
Enable kafka-operator in sandboxed/rbac restricted environments  

### Checklist

- [x] Implementation tested

### DIFF
--- a/main.go
+++ b/main.go
@@ -127,9 +127,10 @@ func main() {
 	}
 
 	kafkaClusterReconciler := &controllers.KafkaClusterReconciler{
-		Client: mgr.GetClient(),
-		Scheme: mgr.GetScheme(),
-		Log:    ctrl.Log.WithName("controllers").WithName("KafkaCluster"),
+		Client:       mgr.GetClient(),
+		DirectClient: mgr.GetAPIReader(),
+		Scheme:       mgr.GetScheme(),
+		Log:          ctrl.Log.WithName("controllers").WithName("KafkaCluster"),
 	}
 
 	if err = controllers.SetupKafkaClusterWithManager(mgr, kafkaClusterReconciler.Log).Complete(kafkaClusterReconciler); err != nil {
--- a/main.go
+++ b/main.go
@@ -32,6 +32,9 @@ package main
 import (
 	"flag"
 	"os"
+	"strings"
+
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 
 	istioclientv1alpha3 "github.com/banzaicloud/istio-client-go/pkg/networking/v1alpha3"
 	banzaiistiov1beta1 "github.com/banzaicloud/istio-operator/pkg/apis/istio/v1beta1"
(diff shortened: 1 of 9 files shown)


amuraru:
When `namespaces` parameter is configured node level labels cannot be retrieved.
I am still testing this
```
2020-02-19T13:25:29.906Z	INFO	controllers.KafkaCluster	updating cr with rack awareness info failed: unable to get: /kafka-worker because of unknown namespace for the cache	{"Request.Namespace": "kafka/kafka", "Request.Name": "kafka"}
```

amuraru:
@stoader @baluchicken I think this can be reviewed. Let me know if you think it there is anything missing.

main.go
```diff
@@ -82,11 +87,28 @@ func main() {
...
+	//Using [RBAC Authorization](https://kubernetes.io/docs/reference/access-authn-authz/rbac/).
+	managerWatchCache := (cache.NewCacheFunc)(nil)
+	if namespaces != "" {
+		ns := strings.Split(namespaces, ",")
+		for i := range ns {
+			ns[i] = strings.TrimSpace(ns[i])
+		}
+		// always append empty namespace to watch cluster-scoped resources
+		// e.g. Nodes' labels are queried as part of broker rack-awareness setup
+		managerWatchCache = cache.MultiNamespacedCacheBuilder(append(ns, ""))
```
stoader: If we append empty string wouldn't the operator watch all namespaces regardless of what has been passed in through the `--namespaces` flag?

amuraru: I'll double check that @stoader. My understanding is that empty ns in the controller cache stands for cluster scoped resources. But will double check

amuraru:
Also relevant for this PR is https://github.com/kubernetes/kubernetes/issues/40610 - having the downwardAPI support Node labels propagation would make the life of Kafka Operator much easier

main.go
stoader: Yes this can work. I'd also add to the note that of rack info is set manually than it's advisable to set pod affinity in such a way that the manually set rack info is in line with where the broker pod is scheduled to.

amuraru: Correct ^ - brokerGroup includes nodeAffinity to schedule the kafka pods in that group in particular failure-domain. Do you want me to add this in the docs here?

stoader: yes please

amuraru:
I think I found an workaround to avoid cluster-scope resources watching in the controller cache and instead use the k8s direct client to get the Node labels:
https://github.com/amuraru/kafka-operator/commit/313ca4ab85dcc357afac434f270c6898b6029825

Let me know if this would work or you see other potential issues with this

PR #311, opened 2020-02-26
Make webhooks optional

amuraru:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### Why?

In constrained environments when Operator is running with least privileges creating the
ValidatingWebhookConfiguration CR may be an option from a RBAC perspective.

Allowing these webhooks to be disabled allows using the operator in these environments.

Default: true (enabled)

### Checklist

- [x] Implementation tested


amuraru:
Hi guys - wanted to get your feedback on this patch. The reason for flag is to optionally lower the privs requirements.

Also - the topic admission webhook seems a bit too heavy as it involves talking with kafka. Wondering if we should push that check to controller and let the webhook do syntactic validations only (kafkaCluster name exists, etc)

PR #340, opened 2020-04-15
Fix indentation for JMX exporter config ADDENDUM

amuraru:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Remove leading spaces from the whole jmx config file so the generated file/config map is more clear.
Also replaced tab indentation with spaces in the cm/file

### DIFF
--- a/pkg/sdk/v1beta1/kafkacluster_types.go
+++ b/pkg/sdk/v1beta1/kafkacluster_types.go
@@ -536,110 +536,108 @@ func (mConfig *MonitoringConfig) GetKafkaJMXExporterConfig() string {
 		return mConfig.KafkaJMXExporterConfig
 	}
 	// Use upstream defined rules https://github.com/prometheus/jmx_exporter/blob/master/example_configs/kafka-2_0_0.yml
-	return `
-    lowercaseOutputName: true
-    rules:
-    # Special cases and very specific rules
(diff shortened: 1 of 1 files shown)


api/v1beta1/kafkacluster_types.go
```diff
@@ -539,110 +539,110 @@ func (mConfig *MonitoringConfig) GetKafkaJMXExporterConfig() string {
...
-    - pattern: kafka.(\w+)<type=(.+), name=(.+)><>Count
-      name: kafka_$1_$2_$3_count
-      type: COUNTER
-    - pattern: kafka.(\w+)<type=(.+), name=(.+)><>(\d+)thPercentile
-      name: kafka_$1_$2_$3
-      type: GAUGE
-      labels:
-        quantile: "0.$4"
-`
+	return strings.TrimSpace(`
```
stoader, line 542: I think we can achieve the same through
```
`lowercaseOutputName: true
```
instead of 
```
`
lowercaseOutputName: true
```

amuraru: I can to do that as well - but prefered to have the file snippet separate on its own lines

amuraru: implemented your suggestion @stoader

amuraru:
rebased on master. Apparently when I use force push in a PR branch all reviews are discarded :D

baluchicken:
No worries:)

PR #342, opened 2020-04-15
Update dependency

baluchicken:
| Q               | A
| --------------- | ---
| Bug fix?        | no|
| New feature?    |yes|
| API breaks?     | no|
| Deprecations?   | no|

### What's in this PR?

- This PR updates the operators dependencies to the latest available.
- Moves the API to a different module to make easier to use it as a lib.
- Fixes controller gen error, by moving it to the bin directory instead of using the system wide one.
- Fixes watcher error which kills the controller runtime if the resource is not available on the cluster

### Checklist

- [x] Implementation tested


config/base/crds/kafka.banzaicloud.io_kafkaclusters.yaml
```diff
@@ -997,12 +529,8 @@ spec:
                 cruiseControlEndpoint:
                   type: string
                 cruiseControlTaskSpec:
-                  description: CruiseControlTaskSpec specifies the configuration of
-                    the CC Tasks
                   properties:
                     RetryDurationMinutes:
-                      description: RetryDurationMinutes describes the amount of time
```
amuraru, line 1004: is the removal of description fields on-purpose or a side effect of version upgrade?

baluchicken: Good catch, I will check this this is a side effect of upgrading to the 0.2.5 generator tool. Will check.

baluchicken: I found the problem will fix it on the next commit.

amuraru: awesome - thanks Balint!

PR #347, opened 2020-04-16
Bump all reasources to use kubernetes 0.17.3 ADDENDUM

amuraru:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Sync helm charts CRDs definitions following the upgrade to k8s 0.17 libs and controller-gen upgrade

### DIFF
--- a/charts/kafka-operator/templates/operator-kafka-crd.yaml
+++ b/charts/kafka-operator/templates/operator-kafka-crd.yaml
@@ -1,20 +1,21 @@
+
+---
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
-  labels:
-    app.kubernetes.io/name: {{ include "kafka-operator.name" . }}
-    helm.sh/chart: {{ include "kafka-operator.chart" . }}
-    app.kubernetes.io/instance: {{ .Release.Name }}
-    app.kubernetes.io/managed-by: {{ .Release.Service }}
-    app.kubernetes.io/version: {{ .Chart.AppVersion }}
-    app.kubernetes.io/component: operator
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.2.5
+  creationTimestamp: null
   name: kafkaclusters.kafka.banzaicloud.io
 spec:
   group: kafka.banzaicloud.io
   names:
     kind: KafkaCluster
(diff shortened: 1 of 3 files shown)


charts/kafka-operator/templates/operator-kafka-crd.yaml
```diff
@@ -1,20 +1,21 @@
+
+---
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
-  labels:
-    app.kubernetes.io/name: {{ include "kafka-operator.name" . }}
```
stoader, line 5: These helm specific labels shouldn't be removed. Unfortunately the gen tool used by kubebuilder is not helm aware thus we need to maintain these labels manually to make the helm charts to be proper

amuraru: Do you feel strong for keeping helm labels in crd @stoader?
The reason I am asking is that we also have an internal patch to move to helm3 where crds are not managed as templates anymore. 

https://geeksocket.in/posts/helm-2-3-migration/ goes at lenght on explaining this. Basically to keep compat between helm2 amd helm3 helm is recommending to move out crds in a separate dir in the chart

amuraru: I am talking specifically about this change https://github.com/adobe/kafka-operator/commit/9537fce209934f9fed9076a3d36c22288a40a303 we wanna propose

stoader: Thinking it more I think you're right.

charts/kafka-operator/templates/operator-kafka-topic-crd.yaml
```diff
@@ -1,39 +1,36 @@
+
+---
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    certmanager.k8s.io/inject-ca-from: {{ .Release.Namespace }}/{{ include "kafka-operator.fullname" . }}-server-cert
```
stoader, line 5: I believe this is needed by cert-manager.

PR #352, opened 2020-04-18
feat(jmx): add config option for KAFKA_JMX_OPTS

kirecek:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | https://github.com/banzaicloud/kafka-operator/issues/351

### What's in this PR?
This PR allows users to configure JMX_PORT and JMX_OPTS environment variables used in https://github.com/kafka-dev/kafka/blob/master/bin/kafka-run-class.sh#L38

### Why?

we would like to collect JMX metrics ourselves since we are using different monitoring stack (DD). See https://github.com/banzaicloud/kafka-operator/issues/351 for more context.

### Checklist

- [x] Implementation tested

I am totally unsure about the approach and what is the best option here.

NOTE: also set exporter options (jmxImage and pathToBin) as optional since they have hardcoded defaults.


baluchicken:
Thanks for opening this PR. It is a great improvement to add. I suggest to make it more generic by adding the ability to specify any Env var in the KafkaCluster spec. Moreover, I guess you need to modify the container port used for metrics as well. I think we should make it configurable as well.
Let me know if you have any more questions. I am happy to help.

kirecek:
> I suggest to make it more generic by adding the ability to specify any Env var in the KafkaCluster spec

Actually that was my initial idea, however, then I saw kafkaHeapOpts and JVMOps which are also env variables so I thought this would be conventional approach? 

> need to modify the container port

not required but indeed useful

baluchicken:
Thanks @kirecek LGTM overall, one minor thing, can you please rename the Env field to Envs.

kirecek:
@baluchicken thx, I have updated the field name. I assume json tag should stay the same as the one in container spec, right? :D

baluchicken:
@kirecek I would make the json tag as `envs` as well. Thanks:)

pkg/resources/kafka/pod.go
```diff
@@ -156,7 +174,7 @@ fi
 								},
 							},
 						},
-					},
+					}, r.KafkaCluster.Spec.Envs...),
```
stoader, line 177: If the user specifies one of the env vars from above (e.g. KAFKA_HEAP_OPTS) in `r.KafkaCluster.Spec.Envs` than the env var list passed to Kafka will contain duplicate env vars.  We should either not allow users to override those env vars via `r.KafkaCluster.Spec.Envs` as for those there is already field in Spec and that should be used or handle duplicates.

I'd opt for the former catching duplicate env vars at validation.

kirecek: Thank you.

I had this in my mind actually and that's one of the reasons why I put those settings to monitoringConfig in the first commit https://github.com/banzaicloud/kafka-operator/pull/352/commits/b6accf4a64fc80f0c2530a0134e92d2447ac6ba0 :D

But to be honest, I am not sure how to catch this at validation. Can you please give me some guidance or point me in the right direction before I write something nasty? :D I have checked the kubebuilder validation spec but I don't see there anything useful that I could use for []corev1.EnvVar?

Sorry, I am not experienced much with this :/

stoader: for now let's ensure that there are no duplicate env vars and `brokerConfig.GetKafkaHeapOpts()`  has precedence over `Spec.Envs` for `KAFKA_HEAP_OPTS` respectively `brokerConfig.GetKafkaPerfJmvOpts()` for  `KAFKA_JVM_PERFORMANCE_OPTS`.

baluchicken:
Thanks @kirecek

PR #361, opened 2020-04-24
Allow override cluster domain

lukemassa:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no (I added a new field that defaults to the existing behavior, not sure if that counts)
| Deprecations?   | no
| Related tickets | 

### What's in this PR?
I've added a field to the KafkaCluster for `clusterDomain`, which defaults to `cluster.local` but can now be overridden.

### Why?
My clusters do not use `cluster.local`, we have a different cluster domain for each cluster. This PR allows me to set the cluster domain, but keeps the existing functionality for everyone else.

### Checklist

- [x] Implementation tested
 (I couldn't find anywhere where the settings in the KafkaCluster were documented, so I just added a description field in the CRD)

### To Do

1) Is KafkaCluster.Spec the right place for this? It's likely this will be the same for all clusters in a given deployment, I couldn't find an easy place to put that.
2) I recognize there's potential for confusion regarding the k8s term "cluster" and the kafka term "cluster". I could make the setting something more like `kubernetesClusterDomain` if that helps?
3) I noticed that most of the functions in `pkg/scale/scale.go` took the same arguments, so I had to copy them around many times. Would it be worth it to encapsulate the parameters into a `struct` and change these functions into methods?


pkg/sdk/v1beta1/kafkacluster_types.go
```diff
@@ -310,6 +311,15 @@ func (kSpec *KafkaClusterSpec) GetIngressController() string {
 	return kSpec.IngressController
 }
 
+// GetDomain returns the default domain if not specified otherwise
+func (kSpec *KafkaClusterSpec) GetClusterDomain() string {
+	if kSpec.ClusterDomain == "" {
+		return "cluster.local"
+		//return "kub.n.tripadvisor.com"
```
baluchicken, line 318: I guess this left here accidentally. :)

lukemassa: haha whoops!

lukemassa:
@baluchicken Great sounds good!

Do you want me to do the refactor of `scale.go` as part of this PR or in a separate one?

baluchicken:
Please do in this PR. Thanks!

baluchicken:
Hi, @lukemassa any update on this? Do you need any further help?

lukemassa:
Hi sorry I got distracted with other things, I should be able to make those edits by the end of the week

lukemassa:
OK I made some progress. I created a `CruiseControlScaler` struct and converted all the functions there to methods. 

I'm not totally happy about the state now, though, since a lot of places especially in `cruisecontroltask_controller.go` I create an object and immediately use it. I was thinking of making CruiseControlScaler a property on KafkaCluster, but that felt a bit strange for such a central object. Any suggestions where to go from here?

baluchicken:
Thanks for the "refactor". I see what you mean, but I think let's address that in a different PR. Also we need some time to think this through. I don't like making a CruiseControlScaler a property on KafkaCluster. Let's leave this one as is, I don't want to delay merging this PR any more:) Can you please create an issue about this, so we can follow up? Thanks.

lukemassa:
I added the issue (linked above).

Thanks!

baluchicken:
Thanks @lukemassa for this great contribution.

Issue #348, opened 2020-04-17
Support for transaction id ACLs


qbast:
**Is your feature request related to a problem? Please describe.**

Using transactional producers requires authorising transactional.id in Kafka. Unfortunately there is no way of adding required ACLs using KafkaUser resource - it only supports topic-level acls.

**Describe the solution you'd like to see**

KafkaUser CRD should be extended to add a list of authorised transactional ids, similar to topicGrants. The only operations supported are WRITE and DESCRIBE, so it there is no need to specify accessType  in CRD.
For example:
```
 - transactionalIdGrants:
   - transactionalId: txid
     patternType: literal
```

This would create two ACLs for the user: 
```
 User:CN=XXX,O=cert-manager,TransactionalId,LITERAL,txid,Describe,Allow,*
 User:CN=XXX,O=cert-manager,TransactionalId,LITERAL,txid,Write,Allow,*
```
**Describe alternatives you've considered**

Simpler solution would be to automatically create wildcard ACLs if a user has any topic write permissions. Similar to how any kafka user with read permissions gets automatic 'Group','*','Read' ACL.
```
User:CN=XXX,O=cert-manager,TransactionalId,LITERAL,*,Describe,Allow,*
User:CN=XXX,O=cert-manager,TransactionalId,LITERAL,*,Write,Allow,*
```

baluchicken:
There are not intention to support more ACLs with Koperator so closing this.

PR #368, opened 2020-05-07
Create catch-all ACLs for transactional and idempotent producer

### LINKED ISSUE #365: Idempotent producer does not work
**Describe the bug**
Idempotent producer fails with "org.apache.kafka.common.errors.ClusterAuthorizationException: Cluster authorization failed" error.

Create a new cluster, a topic and a KafkaUser. Give the new user write access to the topic.
Try sending a message to the topic using idempotent producer. An authorization exception will be thrown.

**Expected behavior**
I would expect that giving write access to a topic also allows idempotent producer to work.

**Additional context**
I realise that idempotent producer also requires IDEMPOTENT_WRITE cluster-level ACL, but I think that if an user is authorised to write to any topic, this ACL should be added automatically by the operator.

qbast:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #365 and #348

### What's in this PR?
During creation of write ACLs also create two more:
 - cluster level ACL enabling idempotent writes
 - TransactionId level ACL enabling write for all transaction ids 

### Additional context
I don't see any harm in enabling idempotent writes in all cases. However support for transaction Ids is quite simplistic - it just allows all of them. If this is good enough or if there should be ability to enable specific transactionIds in KafkaUser CRD is something for discussion.

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/kafkaclient/users.go
+++ b/pkg/kafkaclient/users.go
@@ -149,7 +149,7 @@ func (k *kafkaClient) createWriteACLs(dn string, topic string, patternType saram
 func (k *kafkaClient) createCommonWriteACLs(dn string) (err error) {
 	if err = k.admin.CreateACL(sarama.Resource{
 		ResourceType:        sarama.AclResourceCluster,
-		ResourceName:        "*",
+		ResourceName:        "kafka-cluster",
 		ResourcePatternType: sarama.AclPatternLiteral,
 	}, sarama.Acl{
 		Principal:      dn,
--- a/pkg/kafkaclient/users.go
+++ b/pkg/kafkaclient/users.go
@@ -113,6 +113,9 @@ func (k *kafkaClient) createWriteACLs(dn string, topic string, patternType saram
 	if err = k.createCommonACLs(dn, topic, patternType); err != nil {
 		return
 	}
+	if err = k.createCommonWriteACLs(dn); err != nil {
+		return
+	}
 
 	// WRITE on topic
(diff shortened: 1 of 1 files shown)


baluchicken:
Thanks for your PR, We don't think it is a good idea to enable idempotent writes in all cases as user should have control whether they want to enable it or not. The same true for transaction Ids as well.
It allows specific access to Kafka which should not be granted to all producers by default. Since there is no changes made to the CRD, the users will lack the knowledge that this right is provided.
To support it properly we need to rework the complete KafkaTopic CRD. Which is not on the near future roadmap of the operator. We are happy to get a PR or discuss the ideas how to do it.  The next Supertubes release will ship advanced ACL handling that allows access to all Kafka ACLs declaratively through custom resources.

qbast:
I agree about transaction ids, however I don't understand why idempotent writes need any special ACLs in the first place and what kind of possible attack scenario it allows.

As for your remark about supertube: does it mean that management of Kafka ACLs will move from KafkaUser CRD to something else handled by Supertubes? If yes, is it going to be still possible to use kafka-operator without supertubes?

baluchicken:
We don't want to grant any right to the user automatically which is not required specifically.

To answer your second question:
No KafkaUser CRD will stay and will support all the features which are available right now in the operator. All I was saying is Supertubes will provides a more fine grained ACLs support with Kubernetes RBAC integration.

PR #370, opened 2020-05-11
Allow disabling the "preserveUnknownFields" option in CRD 

aaron-trout:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | n/a

### What's in this PR?
This allows installing the operator on Kubernetes 1.14 by optionally disabling the `preserveUnknownFields` flag in the CRD manifests.

### Why?
Because I am still running 1.14 and want to use the operator 😆

### Checklist

- [x] Implementation tested

### DIFF
--- a/charts/kafka-operator/README.md
+++ b/charts/kafka-operator/README.md
@@ -61,3 +61,4 @@ Parameter | Description | Default
 `webhook.enabled` | Operator will activate the admission webhooks for custom resources | `true`
 `webhook.certs.generate` | Helm chart will generate cert for the webhook | `true`
 `webhook.certs.secret` | Helm chart will use the secret name applied here for the cert | `kafka-operator-serving-cert`
+`crds.preserveUnknownFieldsSupported` | Don't set the `preserveUnknownFields` option on CRDs (k8s 1.14) | `true`
--- a/charts/kafka-operator/templates/operator-kafka-crd.yaml
+++ b/charts/kafka-operator/templates/operator-kafka-crd.yaml
@@ -14,7 +14,9 @@ spec:
     listKind: KafkaClusterList
     plural: kafkaclusters
     singular: kafkacluster
+  {{ if .Values.crds.preserveUnknownFieldsSupported }}
   preserveUnknownFields: false
+  {{ end }}
   scope: Namespaced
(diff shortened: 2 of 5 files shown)


baluchicken:
Thanks for the contribution, can we please use helm feature called `Capabilities`. 
Using that, helm can dynamically decide whether to use this field or not.
https://helm.sh/docs/topics/charts/

aaron-trout:
> Thanks for the contribution, can we please use helm feature called `Capabilities`.
> Using that, helm can dynamically decide whether to use this field or not.
> https://helm.sh/docs/topics/charts/

Ooh cool, I didn't know that was a thing. Will update the PR!

aaron-trout:
Sorry, I completely forgot about this! I did have a go at using `.Capabilities.KubeVersion` but I couldn't find a nice way of comparing the version string in the template. Any thoughts on how to do that cleanly?

baluchicken:
I just checked the official helm chart repo and I think a nice thing to do is like here in the Traefik chart:
https://github.com/helm/charts/blob/24df78eb1835f1b6a605c0cccb487990d6f8dbf4/stable/traefik/templates/rbac.yaml#L8

aaron-trout:
Huh, interesting. `semverCompare` does not appear in the helm docs at all, or even the go template docs!

Screenshot 2020-05-28 at 10 56 50

https://golang.org/pkg/text/template/

Will give that a go though! Will be some time next week as I am moving house this weekend.

baluchicken:
Yeah because helm using not just the go template functions but the Masterminds sprig lib as well.
http://masterminds.github.io/sprig/

baluchicken:
Hi @aaron-trout any update on this? Do you need any help from our side?

aaron-trout:
Hey, sorry for the slowness on this! It's been a busy few weeks and we have actually upgraded to 1.15 now so less of an issue for us. Do you still want to add this or will you drop support for kubernetes <= 1.14?

baluchicken:
Okey then I think we can close this, as kubernetes 1.14 is so old, so we can safely drop the support.

PR #375, opened 2020-05-21
Fix: recreate evicted pod as well

### LINKED ISSUE #326: Evicted pods are not recreated
**Describe the bug**
Two of three kafka broker pods have been evicted by kubernetes. Operator leaves the cluster in this state (one broker running, two down) instead of recreating the evicted pods.

Create a cluster with 3 brokers. Force eviction of any pod.

**Expected behavior**
I would expect that operator notices evicted pods and creates replacements.

**Additional context**
Observed with operator version 0.10.0 , kubernetes 1.15 on GKE.

baluchicken:
| Q               | A
| --------------- | ---
| Bug fix?        |yes|
| New feature?    | no|
| API breaks?     | no|
| Deprecations?   | no|
| Related tickets | fixes #326

### What's in this PR?

Without this fix the operator wont recognise an evicted pod. It leads to states where all broker pods are evicted and the operator does not restarts any of it.

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/k8sutil/resource.go
+++ b/pkg/k8sutil/resource.go
@@ -168,6 +168,14 @@ func IsPodContainsTerminatedContainer(pod *corev1.Pod) bool {
 	return false
 }
 
+// IsPodContainsEvictedContainer returns true if pod status has an evicted reason false otherwise
+func IsPodContainsEvictedContainer(pod *corev1.Pod) bool {
+	if pod.Status.Phase == corev1.PodFailed && pod.Status.Reason == "Evicted" {
+		return true
+	}
+	return false
+}
+
 func IsPodContainsPendingContainer(pod *corev1.Pod) bool {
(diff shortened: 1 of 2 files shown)


pkg/k8sutil/resource.go
```diff
@@ -168,6 +168,14 @@ func IsPodContainsTerminatedContainer(pod *corev1.Pod) bool {
 	return false
 }
 
+// IsPodContainsEvictedContainer returns true if pod status has an evicted reason false otherwise
+func IsPodContainsEvictedContainer(pod *corev1.Pod) bool {
```
stoader, line 172: Since `pod.Status.Reason` is not an enum I's use `strings.Contains(pod.Status.Reason, "Evicted")`

baluchicken: Good point:)

PR #379, opened 2020-05-27
Fix downscale error, do not update cc capacity config when downscale …

baluchicken:
…in progress

| Q               | A
| --------------- | ---
| Bug fix?        |yes|
| New feature?    | no|
| API breaks?     | no|
| Deprecations?   | no|

### What's in this PR?

This PR fixes the bug which leads to infinite reconcile loop when downscaling cluster with Cruise Control Capacity config generated by the operator.

Operator removes the downscaled broker to early from the CC config which leads to nullpointer exception. To fix this, we are going to reuse the old config during downscale.

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/resources/cruisecontrol/configmap.go
+++ b/pkg/resources/cruisecontrol/configmap.go
@@ -31,7 +31,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-func (r *Reconciler) configMap(log logr.Logger, clientPass string) runtime.Object {
+func (r *Reconciler) configMap(log logr.Logger, clientPass string, config *corev1.ConfigMap) runtime.Object {
 	configMap := &corev1.ConfigMap{
 		ObjectMeta: templates.ObjectMeta(
 			fmt.Sprintf(configAndVolumeNameTemplate, r.KafkaCluster.Name),
@@ -46,7 +46,7 @@ func (r *Reconciler) configMap(log logr.Logger, clientPass string) runtime.Objec
     zookeeper.connect=%s
 `, generateBootstrapServer(r.KafkaCluster.Spec.HeadlessServiceEnabled, r.KafkaCluster.Name), r.KafkaCluster.Spec.ListenersConfig.InternalListeners[0].ContainerPort, zookeeperutils.PrepareConnectionAddress(r.KafkaCluster.Spec.ZKAddresses, r.KafkaCluster.Spec.GetZkPath())) +
 				generateSSLConfig(&r.KafkaCluster.Spec.ListenersConfig, clientPass),
-			"capacity.json":       GenerateCapacityConfig(r.KafkaCluster, log),
+			"capacity.json":       GenerateCapacityConfig(r.KafkaCluster, log, config),
 			"clusterConfigs.json": r.KafkaCluster.Spec.CruiseControlConfig.ClusterConfig,
 			"log4j.properties": `
 log4j.rootLogger = INFO, FILE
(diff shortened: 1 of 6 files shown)


pkg/resources/cruisecontrol/configmap.go
```diff
@@ -117,13 +117,20 @@ type Capacity struct {
...
 // generateCapacityConfig generates a CC capacity config with default values or returns the manually overridden value if it exists
-func GenerateCapacityConfig(kafkaCluster *v1beta1.KafkaCluster, log logr.Logger) string {
+func GenerateCapacityConfig(kafkaCluster *v1beta1.KafkaCluster, log logr.Logger, config *corev1.ConfigMap) string {
 	log.Info("Generating capacity config")
 
 	// If there is already a config added manually, use that one
 	if kafkaCluster.Spec.CruiseControlConfig.CapacityConfig != "" {
 		return kafkaCluster.Spec.CruiseControlConfig.CapacityConfig
 	}
+	// During cluster downscale the CR does not contain the required data to generate
```
stoader, line 127:
```suggestion
	// During cluster downscale the CR does not contain data for brokers being downscaled which is required to generate
```

pkg/resources/cruisecontrol/configmap.go
```diff
@@ -117,13 +117,20 @@ type Capacity struct {
...
+	// the proper capacity json for CC so we are reusing the old one.
```
stoader, line 128:
```suggestion
	// the proper capacity json for CC so we are reusing the old one. We can only remove brokers from capacity config when they were removed (pods deleted) from CC as well.
```

pkg/resources/cruisecontrol/configmap.go
```diff
@@ -31,7 +31,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-func (r *Reconciler) configMap(log logr.Logger, clientPass string) runtime.Object {
+func (r *Reconciler) configMap(log logr.Logger, clientPass string, config *corev1.ConfigMap) runtime.Object {
```
stoader, line 34: Inject the capacity config from outside:
```
func (r *Reconciler) configMap(log logr.Logger, clientPass string, capacityConfig string)
```

pkg/resources/cruisecontrol/cruisecontrol.go
```diff
@@ -90,17 +90,42 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
...
+					return errorfactory.New(
+						errorfactory.APIFailure{},
+						err,
+						"getting cruise control configmap failed",
+						"name", key.Name,
+					)
 				}
 			}
+
+			o = r.configMap(log, clientPass, config)
```
stoader, line 118: Invoke `GenerateCapacityConfig` here and pass generated capacity config to `r.configMap` instead of passing the whole `config`

pkg/resources/cruisecontrol/cruisecontrol.go
```diff
@@ -125,3 +150,15 @@ func (r *Reconciler) getClientPassword() (string, error) {
 
 	return clientPass, nil
 }
+
+func isBrokerDeletionInProgress(brokers []v1beta1.Broker, brokerState map[string]v1beta1.BrokerState) bool {
+	if len(brokers) < len(brokerState) {
```
stoader, line 155: Do we need this check here or the `for` loop below should be enough?

Could it happen during upscale that the number of brokers is different from the number of broker states?

pkg/resources/cruisecontrol/deployment.go
```diff
@@ -174,8 +174,8 @@ fi`},
 	}
 }
 
-func generatePodAnnotations(kafkaCluster *v1beta1.KafkaCluster, log logr.Logger) map[string]string {
-	hashedCruiseControlCapacityJson := sha256.Sum256([]byte(GenerateCapacityConfig(kafkaCluster, log)))
+func generatePodAnnotations(kafkaCluster *v1beta1.KafkaCluster, log logr.Logger, config *corev1.ConfigMap) map[string]string {
+	hashedCruiseControlCapacityJson := sha256.Sum256([]byte(GenerateCapacityConfig(kafkaCluster, log, config)))
```
stoader, line 178: Inject capacityConfig as parameter instead of invoking again `GenerateCapacityConfig`

pkg/resources/cruisecontrol/deployment.go
```diff
@@ -31,7 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
-func (r *Reconciler) deployment(log logr.Logger, clientPass string) runtime.Object {
+func (r *Reconciler) deployment(log logr.Logger, config *corev1.ConfigMap) runtime.Object {
```
stoader, line 34: Inject podAnnotations value instead of the config map.

pkg/resources/cruisecontrol/cruisecontrol.go
```diff
@@ -90,17 +90,42 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
...
 				}
 			}
+
+			o = r.configMap(log, clientPass, config)
+			err = k8sutil.Reconcile(log, r.Client, o, r.KafkaCluster)
+			if err != nil {
+				return errors.WrapIfWithDetails(err, "failed to reconcile resource", "resource", o.GetObjectKind().GroupVersionKind())
+			}
+
+			o = r.deployment(log, config)
```
stoader, line 124: Invoke `generatePodAnnotations()` here and pass the generated pod annotations as parameter to `r.deployment()`

PR #385, opened 2020-06-02
Make CRDs installation and ServiceAccount name configurable (#1)

dobrerazvan:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
We needed to have a way to choose if we want to install kafka operator CRDs or not and to name the service account differently. We also started the work to upgrade the chart to helm3 by moving the crds in the crds dedicated folder. This chart maintains compatibility with helm2.

### Why?
It adds more flexibility during deployment where the k8s environment is a restricted and cluster level resources are created by the cluster admin and not the namespace admin. 

### Checklist

- [x] Implementation tested


charts/kafka-operator/values.yaml
```diff
@@ -48,13 +51,20 @@ prometheusMetrics:
...
+
 
 nameOverride: ""
 fullnameOverride: ""
 
 rbac:
   enabled: true
 
+crd:
+  enabled: false
```
amuraru, line 66: let's make sure this is not breaking with previous version. should this be `true` ?

dobrerazvan: fair, yeah, by default should be true, fixing

amuraru:
@dobrerazvan the CRDs should also define 
```
annotations:
    "helm.sh/hook": crd-install
```
so they are both helm3 and helm2 compatible (see https://geeksocket.in/posts/helm-2-3-migration/)

dobrerazvan:
@amuraru Fixed the crds by adding that annotation.

amuraru:
@dobrerazvan i split this in #404 and #405 to ease the review.
Can you please close this and move the discussion in the new PRs?

PR #394, opened 2020-06-18
Do not create service account if default

lukemassa:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Do not attempt to create a ServiceAccount if the user has specified that they want the "default" service account.

### Why?
In my cluster setup, we typically prefer that operators do not create permissions for themselves, instead we like to set all that up by a human operator, then let the code do its thing. Specifically we don't want them to create ServiceAccounts.

### Additional context

The service account code is a little tricky, since you are able to specify in the spec what you want the service account to be for the brokers, cruise control, and envoy, but that doesn't change what service account is actually created.

To get around this, I encapsulated "does the user want to use the default k8s service account" and decided if that was the case, they wouldn't want/need that SA to be managed by the operator.

I'm open to other suggestions of how I might go about this.

### Checklist
- [x] Implementation tested

### To Do


baluchicken:
Thanks for opening a PR. We discussed this internally with the team, and we agreed that is not a good practice to give such rights to the operator to create a serviceAccount. My suggestion is the following:
- We should remove the ServiceAccount creation from the operator and deal with it e.g.: inside the helm chart.
- Also modify the operator to by default use the `default` service account if not specified otherwise.

What are your thoughts about this?

lukemassa:
That sounds great to me!

I think it's a good pattern to have the installation, in this case helm, be responsible for ServiceAccounts, (Cluster)Roles, and (Cluster)RoleBindings, then have the operator/running code use those permissions, so it is easier for human operators to reason about the scope of the k8s operator, and also to prevent accidental privilege escalation.

I know there is debate within k8s about whether the default SA should be used, or if each application should make its own SA. I personally take the approach that the default SA should have very basic privilege (for us we do `view`), and if that's all you need you can use default, but if not you should create a new SA and grant *that* privileges (instead of altering `default`).

So for this case, I think as long as the service account that the brokers/cruise control/envoy run as don't need any k8s privileges (I think they don't?), defaulting to default makes perfect sense. However, if you start needing privileges there, my suggestion would be to change the default to an SA that you explicitly grant RoleBindings to, that way you know exactly what you're getting.

Thanks!

baluchicken:
I agree. You are right brokers and CC and envoy does not require any special right now.
One more question do you want to modify this PR based on our discussion?:)

lukemassa:
Sure! So this would consist of:
1. Removing the reconciliation loop for the SA
2. Changing the three GetServiceAccount()s in `kafkacluster_types.go` to default to `default` instead of `<cluster-name>-cluster`

I was going to add a third part where I put the SA into helm, but now that I've thought about it more that has some issues.

Firstly, how would we know what namespace to put the SA into?

Another thought I had was, won't this break people who are already specifying another SA in the KafkaCluster spec? Then I realized the operator isn't guaranteeing that as it is anyway (that is, it is always creating service accounts of the form `<cluster-name>-cluster`). Maybe, given that, we don't put this in helm, and still allow people to override the ServiceAccount in the Spec if they want to, but leave it up to them to create the ServiceAccount (which is the current behavior).

Again, at least until specific permissions are needed, at which point RoleBindings have to come into play.

baluchicken:
You are right about the helm chart, since it installs only the operator it does not know where to put the SA. Also it is bad behaviour to create the SA in the form `<cluster-name>-cluster` despite of what the user specifies in the spec.
So yeah allow them to specify in the spec but leave it up to them to create the SA.

lukemassa:
OK great, I'll take a look!

lukemassa:
ok @baluchicken that change is made, let me know if there are any issues

lukemassa:
Done. I also grepped through for `serviceaccount` I think the rest are legitimate uses of the operator's service account.

PR #402, opened 2020-06-25
Refine logging for cruise-control interactions

amuraru:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Refactored the logging for calls to cruise-control to:
-  reduce code duplication
- always report the cruise-control error message in logs where there is a non-200 return code

### Why?
This was done while investigating https://github.com/banzaicloud/kafka-operator/issues/400

### Checklist

- [x] Implementation tested

### DIFF
--- a/controllers/cruisecontroltask_controller.go
+++ b/controllers/cruisecontroltask_controller.go
@@ -196,7 +196,7 @@ func (r *CruiseControlTaskReconciler) handlePodAddCCTask(kafkaCluster *v1beta1.K
 	cc := scale.NewCruiseControlScaler(kafkaCluster.Namespace, kafkaCluster.Spec.GetKubernetesClusterDomain(), kafkaCluster.Spec.CruiseControlConfig.CruiseControlEndpoint, kafkaCluster.Name)
 	uTaskId, taskStartTime, scaleErr := cc.UpScaleCluster(brokerIds)
 	if scaleErr != nil {
-		log.Info("cruise control communication error during upscaling broker(s)", "brokerId(s)", brokerIds)
+		log.Info("Cannot upscale broker(s)", "brokerId(s)", brokerIds, "error", scaleErr.Error())
 		return errorfactory.New(errorfactory.CruiseControlNotReady{}, scaleErr, fmt.Sprintf("broker id(s): %s", brokerIds))
 	}
 	statusErr := k8sutil.UpdateBrokerStatus(r.Client, brokerIds, kafkaCluster,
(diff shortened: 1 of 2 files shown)


pkg/scale/scale.go
```diff
@@ -84,7 +84,16 @@ func (cc *CruiseControlScaler) postCruiseControl(action string, options map[stri
 		return nil, err
 	}
 	if rsp.StatusCode != 200 && rsp.StatusCode != 202 {
-		log.Error(errors.New("Non 200 response from cruise-control: "+rsp.Status), "error during talking to cruise-control")
+		ccErr, parseErr := parseCCErrorFromResp(rsp.Body)
```
amuraru, line 87: this was duplicated in multiple cc methods

pkg/scale/scale.go
```diff
@@ -84,7 +84,16 @@ func (cc *CruiseControlScaler) postCruiseControl(action string, options map[stri
...
+		if parseErr != nil {
+			return nil, parseErr
+		}
+		closeErr := rsp.Body.Close()
+		if closeErr != nil {
+			return nil, closeErr
+		}
+		log.Error(errCruiseControlNotReturned200, "non-200 response returned by cruise-control",
```
amuraru, line 95: this is new - for any non-200 response we log an error in logs

controllers/cruisecontroltask_controller.go
```diff
@@ -196,7 +196,7 @@ func (r *CruiseControlTaskReconciler) handlePodAddCCTask(kafkaCluster *v1beta1.K
 	cc := scale.NewCruiseControlScaler(kafkaCluster.Namespace, kafkaCluster.Spec.GetKubernetesClusterDomain(), kafkaCluster.Spec.CruiseControlConfig.CruiseControlEndpoint, kafkaCluster.Name)
 	uTaskId, taskStartTime, scaleErr := cc.UpScaleCluster(brokerIds)
 	if scaleErr != nil {
-		log.Info("cruise control communication error during upscaling broker(s)", "brokerId(s)", brokerIds)
+		log.Info("Cannot upscale broker(s)", "brokerId(s)", brokerIds, "error", scaleErr.Error())
```
amuraru, line 199: updated the wording here - as this can fail not only when there is a communication error with but also when say the broker is not yet ready and not yet registered in cc.
logging the error message as well for better clarity.

pkg/scale/scale.go
hi-im-aren: I see the same in line 112, maybe we can change that as well for the sake of completeness.

amuraru: Good catch @hi-im-aren - will change that as well

amuraru: Done - added a similar code in GET method

baluchicken:
THX @amuraru

PR #416, opened 2020-07-08
Remove Prometheus scraping hardcoded annotations

alungu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes https://github.com/banzaicloud/kafka-operator/issues/398

### What's in this PR?
In https://github.com/banzaicloud/kafka-operator/issues/398 it was reported that adding a Service Monitor for metric scraping would generate metric duplicates since Kafka brokers are already scraped using (hard coded) Prometheus annotations. These Prometheus annotations were removed to allow users to use Service Monitors instead

### Additional context
Prometheus annotations were used to scrape both Broker metrics and Cruise Control metrics.
Prometheus recommends the use of Service Monitors for metrics scraping (this would increase scraping flexibility as well): https://github.com/banzaicloud/kafka-operator/blob/master/config/samples/kafkacluster-prometheus.yaml

For broker metrics, there is still a workaround to use Prometheus annotations: `brokerAnnotations` property.
For Cruise Control, at this point, the only option is to create a service monitor.

### Checklist

- [x] Implementation tested

### To Do
- [ ] Should we also add a property for Cruise Control, similar to `brokerAnnotations`, to offer an alternative for the Service Monitor?
- [ ] Should we update the documentation with more information about the metrics scraping using a Service Monitor?

### DIFF
--- a/charts/kafka-operator/crds/operator-kafka-crd.yaml
+++ b/charts/kafka-operator/crds/operator-kafka-crd.yaml
@@ -64,6 +64,11 @@ spec:
                     additionalProperties:
                       type: string
                     type: object
+                    description: 'Custom annotations for the broker pods:
+                      e.g.: Prometheus scraping annotations:
+                        prometheus.io/scrape: "true"
+                        prometheus.io/port: "9020"
+                      '
                   config:
                     type: string
                   image:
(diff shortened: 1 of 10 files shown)


charts/kafka-operator/crds/operator-kafka-crd.yaml
```diff
@@ -514,6 +519,11 @@ spec:
                         additionalProperties:
                           type: string
                         type: object
+                        description: 'Custom annotations for the broker pods:
```
amuraru, line 522: can we actually add this description in types file so kubebuilder automatically generates this field description?

alungu:
@baluchicken @amuraru impremented review for the auto-generated descriptions. Could you please take a look? Thanks!

PR #424, opened 2020-07-31
Add resizePvc command into alert manager

nikore:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
This implements the resizePvc command from alertmanager. This will add the additional space set in the alert to the existing pvc for the specific broker it alerts for.

### Why?
I have a setup where we want to retain data by time and not size, we tried the addPvc but that is slow and more disruptive than just resizing the pvc. I have tested this with GKE and EKS and it works as expected. 

### Checklist

- [x] Implementation tested

### DIFF
--- a/internal/alertmanager/currentalert/process.go
+++ b/internal/alertmanager/currentalert/process.go
@@ -259,11 +259,16 @@ func resizePvc(log logr.Logger, labels model.LabelSet, annotiations model.LabelS
 				storageConfigs = broker.BrokerConfig.StorageConfigs
 			}
 
-			for i, c := range storageConfigs {
+			for _, c := range storageConfigs {
 				if c.MountPath == pvc.Annotations["mountPath"] {
 					size := *c.PvcSpec.Resources.Requests.Storage()
 
-					size.Add(resource.MustParse(string(annotiations["diskSize"])))
+					incrementBy, err := resource.ParseQuantity(string(annotiations["incrementBy"]))
+					if err != nil {
+						return err
+					}
+
+					size.Add(incrementBy)
 
 					c.PvcSpec.Resources.Requests = corev1.ResourceList{
 						"storage": size,
@@ -278,7 +283,7 @@ func resizePvc(log logr.Logger, labels model.LabelSet, annotiations model.LabelS
 					}
 				}
 			}
-			cr.Spec.Brokers[i] = broker
+			cr.Spec.Brokers[i].BrokerConfig = broker.BrokerConfig
 		}
 	}
 
@@ -287,7 +292,7 @@ func resizePvc(log logr.Logger, labels model.LabelSet, annotiations model.LabelS
 		return err
 	}
 
-	log.Info(fmt.Sprintf("PVC for %s successfully resized to broker %s", pvc.Annotations["mountPath"], pvc.Labels["brokerId"]))
+	log.Info("successfully resized broker pvc", "mount path", pvc.Annotations["mountPath"], "broker id", pvc.Labels["brokerId"])
 
 	return nil
 }
--- a/internal/alertmanager/currentalert/process.go
+++ b/internal/alertmanager/currentalert/process.go
@@ -282,7 +282,7 @@ func resizePvc(log logr.Logger, labels model.LabelSet, annotiations model.LabelS
 		}
 	}
 
-	err = k8sutil.UpdateCrWithRollingUpgrade(0, cr, client)
+	err = k8sutil.UpdateCr(cr, client)
 	if err != nil {
 		return err
 	}
--- a/internal/alertmanager/currentalert/process.go
+++ b/internal/alertmanager/currentalert/process.go
@@ -139,7 +139,7 @@ func (e *examiner) processAlert(ds disableScaling) (bool, error) {
 		if err := validators.ValidateAlert(); err != nil {
 			return false, err
 		}
-		err := resizePvc(e.Alert.Labels, e.Alert.Annotations, e.Client)
+		err := resizePvc(e.Log, e.Alert.Labels, e.Alert.Annotations, e.Client)
 		if err != nil {
 			return false, err
 		}
@@ -238,8 +238,56 @@ func addPvc(log logr.Logger, alertLabels model.LabelSet, alertAnnotations model.
 	return nil
 }
 
-func resizePvc(labels model.LabelSet, annotiations model.LabelSet, client client.Client) error {
-	//TODO
+func resizePvc(log logr.Logger, labels model.LabelSet, annotiations model.LabelSet, client client.Client) error {
+	cr, err := k8sutil.GetCr(string(labels["kafka_cr"]), string(labels["namespace"]), client)
+	if err != nil {
+		return err
+	}
+
+	pvc, err := getPvc(string(labels["persistentvolumeclaim"]), string(labels["namespace"]), client)
+	if err != nil {
+		return err
+	}
+
+	for i, broker := range cr.Spec.Brokers {
+		if strconv.Itoa(int(broker.Id)) == pvc.Labels["brokerId"] {
+			var storageConfigs []v1beta1.StorageConfig
+
+			if broker.BrokerConfig == nil {
+				storageConfigs = cr.Spec.BrokerConfigGroups[cr.Spec.Brokers[i].BrokerConfigGroup].StorageConfigs
+			} else {
+				storageConfigs = broker.BrokerConfig.StorageConfigs
+			}
+
+			for i, c := range storageConfigs {
+				if c.MountPath == pvc.Annotations["mountPath"] {
+					size := *c.PvcSpec.Resources.Requests.Storage()
+
+					size.Add(resource.MustParse(string(annotiations["diskSize"])))
+
+					c.PvcSpec.Resources.Requests = corev1.ResourceList{
+						"storage": size,
+					}
+
+					if broker.BrokerConfig == nil {
+						broker.BrokerConfig = &v1beta1.BrokerConfig{
+							StorageConfigs: []v1beta1.StorageConfig{c},
+						}
+					} else {
+						broker.BrokerConfig.StorageConfigs[i] = c
+					}
+				}
+			}
+			cr.Spec.Brokers[i] = broker
+		}
+	}
+
+	err = k8sutil.UpdateCrWithRollingUpgrade(0, cr, client)
+	if err != nil {
+		return err
+	}
+
+	log.Info(fmt.Sprintf("PVC for %s successfully resized to broker %s", pvc.Annotations["mountPath"], pvc.Labels["brokerId"]))
 
 	return nil
 }
--- a/internal/alertmanager/currentalert/process_test.go
+++ b/internal/alertmanager/currentalert/process_test.go
@@ -16,6 +16,7 @@ package currentalert
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
 	"github.com/banzaicloud/kafka-operator/api/v1beta1"
@@ -54,9 +55,8 @@ func Test_resizePvc(t *testing.T) {
 						"node":                  "test-node",
 					},
 					Annotations: model.LabelSet{
-						"command":         "resizePvc",
-						"mountPathPrefix": "/kafka-logs",
-						"diskSize":        "2G",
+						"command":     "resizePvc",
+						"incrementBy": "2G",
 					},
 				},
 			},
@@ -125,9 +125,8 @@ func Test_resizePvc(t *testing.T) {
 						"node":                  "test-node",
 					},
 					Annotations: model.LabelSet{
-						"command":         "resizePvc",
-						"mountPathPrefix": "/kafka-logs",
-						"diskSize":        "2G",
+						"command":     "resizePvc",
+						"incrementBy": "2G",
 					},
 				},
 			},
@@ -229,6 +228,7 @@ func Test_resizePvc(t *testing.T) {
 				t.Errorf("kafka cr was not found, error = %v", err)
 			}
 
+			fmt.Println(kafkaCluster.Spec.Brokers)
 			brokerStorageConfig := &kafkaCluster.Spec.Brokers[0].BrokerConfig.StorageConfigs[0]
 
 			if brokerStorageConfig.PvcSpec.Resources.Requests.Storage().Value() != 6294967296 {
--- a/internal/alertmanager/currentalert/process_test.go
+++ b/internal/alertmanager/currentalert/process_test.go
@@ -31,6 +31,216 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+func Test_resizePvc(t *testing.T) {
+	testClient := fake.NewFakeClientWithScheme(scheme.Scheme)
+
+	//Setup test kafka cluster and pvc
+	setupEnvironment(t, testClient)
+
+	testCase := []struct {
+		name      string
+		alertList []model.Alert
+		pvc       corev1.PersistentVolumeClaim
+		cluster   v1beta1.KafkaCluster
+	}{
+		{
+			name: "Resize pvc created by default group",
+			alertList: []model.Alert{
+				{
+					Labels: model.LabelSet{
+						"kafka_cr":              "test-cluster",
+						"namespace":             "kafka",
+						"persistentvolumeclaim": "testPvc",
+						"node":                  "test-node",
+					},
+					Annotations: model.LabelSet{
+						"command":         "resizePvc",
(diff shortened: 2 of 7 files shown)


internal/alertmanager/currentalert/process.go
```diff
@@ -238,8 +238,56 @@ func addPvc(log logr.Logger, alertLabels model.LabelSet, alertAnnotations model.
...
+					} else {
+						broker.BrokerConfig.StorageConfigs[i] = c
+					}
+				}
+			}
+			cr.Spec.Brokers[i] = broker
+		}
+	}
+
+	err = k8sutil.UpdateCrWithRollingUpgrade(0, cr, client)
```
amuraru, line 285: Is this needed so the kafka pods are bounced to pick up the new resized PV? Is that's the case we may want to make this configurable as newer versions of CSI plugins support hot disk resize

amuraru: Also in my environment Insaw that when resizing the pvc (albeit manually manipulating the kafka cr spec and triggering a rolling restart) it may take some time until the underlying disk is resized. We may need to poll the status of the pv and only when fully resized trigger a rolling here

nikore: The main reason why I'm updating the KafkaCR is in the next reconciliation loop it will attempt to lower the PVC value which isn't supported. I wanted to originally just change the value in the PVC and in the kafka CR without triggering a restart but the update CR method is private and I figured not every cluster supports dynamic resizing. 

In my setup the restart isn't necessary either; I would assume pretty much any CSI device wouldn't need to restart. I figured there was little harm in restarting, in the current proposal im using the reconciliation loop to increase the size of the PVC which I think is safer.

baluchicken: I would make the `updateCR` public or create an another wrapper around it instead of using `k8sutil.UpdateCrWithRollingUpgrade. It is meant to update the errorCount value on the CR. Setting that value to zero will cause problems. That should not be changed from its value, or at least not during a resize alert.

We are aware that rolling upgrading a cluster after a pvc resize might not be handled correctly. I suggest the following. I create an Issue something like `Investigate whether rolling upgrade is required in case of PVC resize` and we can move the discussion there. I think that issue is not part of this PR and should not be addressed here.
How does this sound to you?

nikore:
I changed `UpdateCR` to be public and used it instead of the other method. 

I also added a 3rd commit that makes the `StorageConfig` unique since we are appending the ConfigGroups to the end. I can make this a different PR if we think it exceeds the scope of this. I noticed they where not unique in testing my `resizePvc` command in dev

baluchicken:
Thanks, I don't think the dedup is the right thing to do here. I think we need to merge these two configs like we do in case of other config values. That way we can set precedence. broker -> configgroups.

nikore:
This is post the config merge

it currently does:

```
       err := mergo.Merge(bConfig, clusterSpec.BrokerConfigGroups[broker.BrokerConfigGroup], mergo.WithAppendSlice)
	if err != nil {
		return nil, errors.WrapIf(err, "could not merge brokerConfig with ConfigGroup")
	}
```

which appends the storage configs from the BrokerConfigGroup to the end of the ones from the broker specific config. It currently inflate `bConfig` via:

```
	bConfig := &v1beta1.BrokerConfig{}
	if broker.BrokerConfigGroup == "" {
		return broker.BrokerConfig, nil
	} else if broker.BrokerConfig != nil {
		bConfig = broker.BrokerConfig.DeepCopy()
	}
```

so before we use `mergo.Merge` we already have the `[]StorageConfig` initialize with with the broker specific `[]StorageConfig` 

The "dedup" logic makes broker specific configs take precedence over the ConfigGroup.

Unless I misunderstanding something; this seems to be the desired behavior. Without the reconciliation loop fails trying to reduce the PVC size (which no platform I know of outside of NFS allows for lowering the pvc size) because the `[]StorageConfig` has both the `StorageConfig` from the `brokerConfig` and from the `brokerConfigGroup` I can get rid of the deduping but we would need to solve the reconciliation failure.

baluchicken:
Sorry for the delayed response. I tried and you are right there will be duplicate values in the []StorageConfig slice. The main reason this happens is because of how mergo handles slice merges. By default mergo merge policy regarding to slice is to overwrite it with the value which is coming from the second parameter which our case is `clusterSpec.BrokerConfigGroups[broker.BrokerConfigGroup]`. However we added a merge option which tells mergo to append slices instead of overwriting it. This causes the duplication problem. 

I think we have an ultimate option here is to change the `[]StorageConfig` value from slice to `map[string]*corev1.PersistentVolumeClaimSpec` because that way mergo will take care of the merging properly. 
But in the meantime maybe this dedubing can work as well. Thanks for pointing this out:)

schultheiszn:
I would find a way to check the underlying StorageClass since not all SCs support volume expansion by default e.g. amazon's gp2 support for expansion is disabled by default

internal/alertmanager/currentalert/process.go
```diff
@@ -238,8 +238,56 @@ func addPvc(log logr.Logger, alertLabels model.LabelSet, alertAnnotations model.
...
+				storageConfigs = cr.Spec.BrokerConfigGroups[cr.Spec.Brokers[i].BrokerConfigGroup].StorageConfigs
+			} else {
+				storageConfigs = broker.BrokerConfig.StorageConfigs
+			}
+
+			for i, c := range storageConfigs {
+				if c.MountPath == pvc.Annotations["mountPath"] {
+					size := *c.PvcSpec.Resources.Requests.Storage()
+
+					size.Add(resource.MustParse(string(annotiations["diskSize"])))
```
stoader, line 266: Extend `resizePvcValidator` or implement a new one  to validate that there is a `diskSize` annotation on the alert and is in the correct format. We should not use `MustParse` as the can exit with `panic` but rather ensure that the execution reaches this part only if the received alert is valid for processing.

The `diskSize` is misleading as in fact it's an increment, thus `diskSizeIncrement` or `incrementBy` would be a better name.

internal/alertmanager/currentalert/process.go
```diff
@@ -238,8 +238,56 @@ func addPvc(log logr.Logger, alertLabels model.LabelSet, alertAnnotations model.
...
+					c.PvcSpec.Resources.Requests = corev1.ResourceList{
+						"storage": size,
+					}
+
+					if broker.BrokerConfig == nil {
+						broker.BrokerConfig = &v1beta1.BrokerConfig{
+							StorageConfigs: []v1beta1.StorageConfig{c},
+						}
+					} else {
+						broker.BrokerConfig.StorageConfigs[i] = c
```
stoader, line 277: Wouldn't this wipe out the fields of existing storage config other than `storage` field (e.g. `mountPath`)? `mountPath` doesn't have `omitempty` specified in the json tag so it may wipe out `mountPath` if no value is set.

Instead of  `broker.BrokerConfig.StorageConfigs[i] = c` it's safer to update the `storage` field of the existing storage config instance if exists.

internal/alertmanager/currentalert/process.go
```diff
@@ -238,8 +238,56 @@ func addPvc(log logr.Logger, alertLabels model.LabelSet, alertAnnotations model.
...
+			cr.Spec.Brokers[i] = broker
+		}
+	}
+
+	err = k8sutil.UpdateCr(cr, client)
+	if err != nil {
+		return err
+	}
+
+	log.Info(fmt.Sprintf("PVC for %s successfully resized to broker %s", pvc.Annotations["mountPath"], pvc.Labels["brokerId"]))
```
stoader, line 290: Please use structured logging.
e.g.:
```
	log.Info("successfully resized broker pvc",  "mount path", pvc.Annotations["mountPath"], "broker id", pvc.Labels["brokerId"], "new size", size)
```

internal/alertmanager/currentalert/process_test.go
```diff
@@ -31,6 +31,216 @@ import (
...
+				{
+					Labels: model.LabelSet{
+						"kafka_cr":              "test-cluster",
+						"namespace":             "kafka",
+						"persistentvolumeclaim": "testPvc",
+						"node":                  "test-node",
+					},
+					Annotations: model.LabelSet{
+						"command":         "resizePvc",
+						"mountPathPrefix": "/kafka-logs",
```
stoader, line 58: `mountPathPrefix` not used.

internal/alertmanager/currentalert/process_test.go
```diff
@@ -31,6 +31,216 @@ import (
...
```
stoader, line 129: `mountPathPrefix` not used

internal/alertmanager/currentalert/process.go
nikore: I changed it to `incrementBy` and I implemented some basic checking logic in `resizePvcValidator`. 

I will say you guys should consider opening a ticket to do the same thing to `addPvc` as well as `upScale` they do not do any validation on `diskSize` in the `*Validator` and they all implement `resources.MustParse`. I basically followed `addPvc` as close as I could to not change the style of the code or make it look too different.

internal/alertmanager/currentalert/process.go
nikore: I played around with this a lot, and I think I got it simpler and less possibly dangerous but to explain why it looks the way it does:

First I have to resolve what is the `brokerConfig`/`[]storageConfig` for the the broker in question which is what I do here https://github.com/banzaicloud/kafka-operator/blob/101bc86d972cd59983cce899a8bcee81b3e73f17/internal/alertmanager/currentalert/process.go#L256

namely:
```
			if broker.BrokerConfig == nil {
				storageConfigs = cr.Spec.BrokerConfigGroups[cr.Spec.Brokers[i].BrokerConfigGroup].StorageConfigs
			} else {
				storageConfigs = broker.BrokerConfig.StorageConfigs
			}
```

Now that I have some storageConfig slice I go through them looking for a `mountPath` that matches (https://github.com/banzaicloud/kafka-operator/blob/101bc86d972cd59983cce899a8bcee81b3e73f17/internal/alertmanager/currentalert/process.go#L263) if I find it I resize the PVC based on the `incrementBy` set in the alert. 

Where it gets tricky is in the original CR request the brokerConfig could still be null or not so I try to handle the case here where I create a `brokerConfig` with just a `storageConfig` in it or I update the specific `storageConfig` if one existed. I update it from the copy the `range` made of object in the slice + my modification to the pvc size (https://github.com/banzaicloud/kafka-operator/blob/101bc86d972cd59983cce899a8bcee81b3e73f17/internal/alertmanager/currentalert/process.go#L277)

```
if broker.BrokerConfig == nil {
  broker.BrokerConfig = &v1beta1.BrokerConfig{
    StorageConfigs: []v1beta1.StorageConfig{c},
  }
} else {
  broker.BrokerConfig.StorageConfigs[i] = c
}
```

since its a copy of the object was in the slice previously it should contain all that was there previously such as `mountPath` or anything else (I think the storage config only has `mountPath` and `PvcSpec` 

then in https://github.com/banzaicloud/kafka-operator/blob/101bc86d972cd59983cce899a8bcee81b3e73f17/internal/alertmanager/currentalert/process.go#L286 I update the original CR I got, I wanted to just update the storageConfig but since a brokerConfig might not exist I have to update the whole brokerConfig object. So because the config might not exist I can't safely do something like `cr.Spec.Brokers[i].BrokerConfig.StorageConfig[ii].PvcSpec.Resources.Requests = newPvcConfig{}` as some or none of that whole path might exist. I could `if/else` it until I find the lowest level of the path and update just that, but I think my current solution balances safety and readability. 

I do cover some of the cases in the unit test I wrote, making sure we don't loose a storage config during the update. 

If you have other suggestions I'm happy to try.

nikore:
Note the unit test build is failing due to https://github.com/banzaicloud/kafka-operator/pull/428

internal/alertmanager/currentalert/process.go
```diff
@@ -238,8 +238,61 @@ func addPvc(log logr.Logger, alertLabels model.LabelSet, alertAnnotations model.
...
+
+					size.Add(incrementBy)
+
+					c.PvcSpec.Resources.Requests = corev1.ResourceList{
+						"storage": size,
+					}
+
+					if broker.BrokerConfig == nil {
+						broker.BrokerConfig = &v1beta1.BrokerConfig{
+							StorageConfigs: []v1beta1.StorageConfig{c},
```
stoader, line 279: Let's assume that the storage configuration for the broker comes from a configuration group (thus `broker.BrokerConfig` is nil) and there are multiple log dirs (multiple PVCs) configured in the configuration group.
 
This means that we are going to have a slice of storage configs in the configuration group and a slice of storage configs with one element at the broker level. Since the two configs are merged using `WithAppendSlice` (https://github.com/banzaicloud/kafka-operator/blob/master/pkg/util/util.go#L193) this will lead to incorrect storage configuration.

```
...
spec:
  brokerConfigGroups:
    default:
      storageConfigs:
        - mountPath: "/kafka-logs-1"
          pvcSpec:
            accessModes:
              - ReadWriteOnce
            resources:
              requests:
                storage: 10Gi
        - mountPath: "/kafka-logs-2"
          pvcSpec:
            accessModes:
              - ReadWriteOnce
            resources:
              requests:
                storage: 10Gi
  brokers:
    - id: 0
      brokerConfigGroup: "default"
    - id: 1
      brokerConfigGroup: "default"
    - id: 2
      brokerConfigGroup: "default"
```

When an alert is triggered for let 's say the PVC that holds `/kafka-logs-2` this results
```
spec:
  brokerConfigGroups:
    default:
      storageConfigs:
        - mountPath: "/kafka-logs-1"
          pvcSpec:
            accessModes:
              - ReadWriteOnce
            resources:
              requests:
                storage: 10Gi
        - mountPath: "/kafka-logs-2"
          pvcSpec:
            accessModes:
              - ReadWriteOnce
            resources:
              requests:
                storage: 10Gi
  brokers:
    - id: 0
      brokerConfigGroup: "default"
      brokerConfig:
        storageConfigs:
          - mountPath: "/kafka-logs-2"
            pvcSpec:
              accessModes:
                - ReadWriteOnce
               resources:
                 requests:
                    storage: 25Gi
    - id: 1
      brokerConfigGroup: "default"
      brokerConfig:
        storageConfigs:
          - mountPath: "/kafka-logs-2"
            pvcSpec:
              accessModes:
                - ReadWriteOnce
               resources:
                 requests:
                    storage: 25Gi
    - id: 2
      brokerConfigGroup: "default"  
      brokerConfig:
        storageConfigs:
          - mountPath: "/kafka-logs-2"
            pvcSpec:
              accessModes:
                - ReadWriteOnce
               resources:
                 requests:
                    storage: 25Gi 
```

The `util.GetBrokerConfig` method will return a storage configuration slice that has `/kafka-logs-2` twice here https://github.com/banzaicloud/kafka-operator/blob/master/pkg/resources/kafka/kafka.go#L221 . This leads to non-deterministic behaviour.  In this case we should copy the entire storage configuration slice from the config group  and update the appropriate item in the slice.

Have you tested this scenario?

nikore: Yes I have, there was discussion on this topic earlier on in this PR.

I updated the `utils.GetBrokerConfig` to dedup the storageConfigs based on the `mountPath` https://github.com/banzaicloud/kafka-operator/blob/101bc86d972cd59983cce899a8bcee81b3e73f17/pkg/util/util.go#L198

The broker specific configs take precedence over the `brokerConfigGroups` so in your example you would only end up reconciling one`/kafka-logs-2`. 

when we do: `err := mergo.Merge(bConfig, clusterSpec.BrokerConfigGroups[broker.BrokerConfigGroup], mergo.WithAppendSlice)` the bConfig is already inflated via a `deepCopy` if it is there. This will be in the first part of the new slice that `mergo` creates. 

I created a unit test to specific test this dedup/merge logic here: https://github.com/banzaicloud/kafka-operator/blob/101bc86d972cd59983cce899a8bcee81b3e73f17/pkg/util/util_test.go#L94

which covers the case of the path being defined in both the `brokerConfigGroup` and the `brokerConfig`. 

I also have this version of the operator deployed to a cluster managing 26 kafka clusters all with resized enabled so far 16 of them have had at least one volume resized. No issues as of yet (~1 week of running it).

stoader:
@nikore can you rebase your PR such as it picks up https://github.com/banzaicloud/kafka-operator/commit/bae79b91ecba085fe4426e66d3b06c04eef13ec6. It should address the license check issue.

hi-im-aren:
LGTM, thanks @nikore

stoader:
LGTM, thanks @nikore

PR #457, opened 2020-09-24
Clean removal of brokers from envoy ingress controller on cluster scale-down

### LINKED ISSUE #440: Clean removal of brokers from envoy ingress controller on cluster scale-down
While implementing https://github.com/banzaicloud/kafka-operator/issues/438 I realised we can improve the way ingress
controller manages the configuration for upstream kafka broker proxying.

Currently the envoy config is generated exclusively based of `KafkaCluster.Spec.Brokers` field. 
This works in practice for initial setup and scale-up operations as the newly added brokers are added in ingress but won't be "advertised" as partitions owners by kafka internal metadata mechanism until they actually start serving partitions.
So it doesn't really require to synchronise the ingress config with actual kafka brokers being operational. (i.e. having a broker exposed in ingress but not yet returned in Metadata Responses from Kafka is a non-issue)

When scaling down the cluster though, the sequence of removing the brokers from ingress matters.
Here is an example of events:
1. Start with a cluster of say 6 brokers
2. Envoy ingress is configured to proxy all of them
3. Remove a broker from ``KafkaCluster.Spec.Brokers`
kafka and envoy reconcilers are both notified of this change and they async operate the corresponding changes:
(3.1) envoy reconciler is refreshing the envoy.yaml configuration removing the broker
(3.2) kafka reconciler starts draining the broker and waiting for cruise-control to complete decomissioning

As (3.1) is faster than (3.2) - the ingress basically makes the pending decomissiong broker unavailable for an external Kafka client.

The operator also maintains the `status` sub-resource on KafkaCluster.
This is a more reliable source to drive the ingress configuration as it tracks all states for a kafka broker.
The envoy reconciler should be updated to watch this Status field and update the configuration accordingly.
 

N/A

**Additional context**
N/A

baluchicken:
| Q               | A
| --------------- | ---
| Bug fix?        |yes|
| New feature?    | no|
| API breaks?     | no|
| Deprecations?   | no|
| Related tickets | fixes #440 

### What's in this PR?

Like described in issue 440, this PR instead of using spec to determine which broker is available at the moment. It uses the Status field which is a more reliable source for this information.

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -187,6 +187,7 @@ func GetBrokerIdsFromStatus(brokerStatuses map[string]v1beta1.BrokerState, log l
 		id, err := strconv.Atoi(brokerId)
 		if err != nil {
 			log.Error(err, "could not parse brokerId properly")
+			continue
 		}
 		brokerIds = append(brokerIds, id)
 	}
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -180,6 +180,16 @@ func AreStringSlicesIdentical(a, b []string) bool {
 	return reflect.DeepEqual(a, b)
 }
 
+func GetBrokerIdsFromStatus(brokerStatuses map[string]v1beta1.BrokerState) []int {
+	brokerIds := make([]int, 0, len(brokerStatuses))
+	for brokerId, _ := range brokerStatuses {
+		id, _ := strconv.Atoi(brokerId)
+		brokerIds = append(brokerIds, id)
+	}
+	sort.Ints(brokerIds)
+	return brokerIds
+}
+
 // GetBrokerConfig compose the brokerConfig for a given broker
 func GetBrokerConfig(broker v1beta1.Broker, clusterSpec v1beta1.KafkaClusterSpec) (*v1beta1.BrokerConfig, error) {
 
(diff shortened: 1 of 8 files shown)


pkg/util/util.go
```diff
@@ -180,6 +180,16 @@ func AreStringSlicesIdentical(a, b []string) bool {
 	return reflect.DeepEqual(a, b)
 }
 
+func GetBrokerIdsFromStatus(brokerStatuses map[string]v1beta1.BrokerState) []int {
+	brokerIds := make([]int, 0, len(brokerStatuses))
+	for brokerId, _ := range brokerStatuses {
```
stoader, line 185:
```suggestion
	for brokerId := range brokerStatuses {
```

pkg/util/util.go
```diff
@@ -180,6 +180,16 @@ func AreStringSlicesIdentical(a, b []string) bool {
...
+		id, _ := strconv.Atoi(brokerId)
```
stoader, line 186: should we treat the error to be on the safe side?

stoader (changes requested):

stoader (dismissed):

stoader (approved):

schultheiszn (approved):

Issue #448, opened 2020-09-08
Add support for Disruption Budgets


ecojan:
**Is your feature request related to a problem? Please describe.**
When running on a kubernetes cluster not managed by your team and have `oneBrokerPerNode: false`, you are vulnerable to disruptions in the cluster due to rolling upgrades of the K8s cluster. A second problem is allowing the replicas to be moved to a new Kafka broker (for an `acks=all` producer).

**Describe the solution you'd like to see**
Creation of a PDB managed by the operator, to alleviate the disruption of a  K8s cluster rolling upgrade.
Support for checking the URP before taking down a second pod.

**Describe alternatives you've considered**
Disabling autoscaling and supporting PDB from helm chart (since we know the number of brokers) but this will solve only half of the problem.

ecojan:
Hi @baluchicken ,
I got around today to implement this and test it out. Please have a look at the PR when you have time.
Thank you

baluchicken:
Thanks @ecojan implementing this.

PR #460, opened 2020-09-28
Added support for PodDisruptionBudget

ecojan:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | implements #448 

### What's in this PR?
Added a podDisruptionBudget managed by the kafka-operator.

### Why?
To make a KafkaCluster deployment more resilient to disruptions. When running on a kubernetes cluster not managed by your team, the KafkaCluster is vulnerable to disruptions in the K8s cluster (eg. rolling upgrades, node drains). PodDisruptionBudgets allow us to reduce controlled disruptions on a KafkaCluster by only allowing 1 broker to be down at a time (via .spec.minAvailable)

### Checklist

- [X] Implementation tested
- [X] Error handling code meets the [guideline](https://github.com/banzaicloud/developer-guide/blob/master/docs/coding-style/error-handling-guide.md)


pkg/resources/kafka/poddisruptionbudget.go
```diff
@@ -0,0 +1,57 @@
...
+				MatchLabels: LabelsForKafka(r.KafkaCluster.Name),
+			},
+		},
+	}
+
+}
+
+// Calculate maxUnavailable as max between brokerCount - 1 (so we only allow 1 broker to be disrupted)
+// and 1 (to cover for 1 broker clusters)
+func (r *Reconciler) computeMinAvailable() intstr.IntOrString {
```
amuraru, line 55: @cojan minAvailable needs to be configurable. Value 1 works for cases when oneBrokerPerNode is configured but not when this is disabled. In the latter case there may be multiple brokers per node and a value 1 of minAvailable will imped the k8s cluster upgrade if there are multiple broker pods scheduled on the same node.

My proposal would be to put this pbd under a feature flag and let the user configure the actual value of minAvailable

ecojan: @amuraru I agree that minAvailable might be worth configuring to some degree, I actually was planning on asking for some guidance where this configuration should be stay.

In the latter case that you describe, the PDB will act the same as for the "oneBrokerPerNode", the only difference would be in the time required for the node to be drained (as 2 or more brokers need to be moved before finishing the node drain). In the case of multiple brokers on a single node, it's even more critical in my opinion.

Agreed to put this under a feature flag though.

If I'm missing anything please let me know.

baluchicken (commented):
Please also update the annotation before the kafkacluster controller reconcile method to allow the controller to Get,Update,Create poddistruption budget resources. Also please do not forget to update the chart as well.

pkg/resources/kafka/poddisruptionbudget.go
amuraru: I take my comment back - it seems that that k8s descheduler/drain can work with 1  minUnvailable replica and multiple kafka pods on the same node.
I don't think that making this configurable unless the banzaicloud maintainers feel so

ecojan:
Hi @baluchicken , 
Updated the PR with the missing RBAC, please have another look

pkg/resources/kafka/poddisruptionbudget.go
baluchicken: I like the idea of putting this under a flag and make it configurable. Also what do you think instead of using a number we should use percentage. Using percentage we can support scenarios where the cluster can tolerate more than one broker outage e.g.: when all your topics has three replication factors.
I know we cannot use percentage on the PDB side but the operator can calculate the number based on the percentage specified in the KafkaCluster spec.

ecojan: All for putting it under a feature flag in the KafkaCluster spec, was thinking about:
```yaml
disruptionBudget:
 create: true
 budget: "33%"
```
Let me know what are your thoughts.

I was also wondering if we could support both percentage and static number, wdyt?

I was looking at a case with replication factor 2 and multi AZ. When switching to the next VM from an AZ to another, there is a chance to have 2 brokers down at the same time and both of them to hold one of the replicas. I know it's a corner-case  but I would like to have the flexibility of using a number as well.

baluchicken: I like it. Yeah it is reasonable to support both percentage and static number.

ecojan: Had a bit of time today and added the extra work to implement the above, please have another look.

pkg/resources/kafka/poddisruptionbudget.go
```diff
@@ -0,0 +1,101 @@
...
+	} else {
+		// treat static number budget
+		staticBudget, err := strconv.ParseInt(disruptionBudget, 10, 0)
+		if err != nil {
+			log.Error(err, "error occurred during parsing the disruption budget")
+		} else {
+			budget = int(staticBudget)
+		}
+
+	}
```
amuraru, line 98: what if `Spec.DisruptionBudget.Budget` is not set?

config/samples/simplekafkacluster.yaml
```diff
@@ -35,6 +35,8 @@ spec:
       brokerConfigGroup: "default"
   rollingUpgradeConfig:
     failureThreshold: 1
+  disruptionBudget:
+    create: false
```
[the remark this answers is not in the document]
ecojan: The CRD schema validation for v1beta1 doesn't allow defaulting out of the box (available in v1).
For now allowing it to be missing solves this.

pkg/resources/kafka/poddisruptionbudget.go
ecojan: Thanks for pointing this out. I was unsure how the error was treated, did some testing around, found undesired behavior and now should cover this.

config/samples/simplekafkacluster.yaml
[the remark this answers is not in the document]
amuraru: thanks @ecojan - lgtm

pkg/resources/kafka/poddisruptionbudget.go
amuraru: thanks @ecojan - lgtm

config/base/crds/kafka.banzaicloud.io_kafkaclusters.yaml
```diff
@@ -1146,6 +1146,18 @@ spec:
                   - replicationFactor
                   type: object
               type: object
+            disruptionBudget:
```
baluchicken, line 1149: Could you please place this to the charts/kafka-operator/crds.yaml as well.

baluchicken (commented):
Just some minor thing otherwise LGTM thx @ecojan

baluchicken (approved):
LGTM thx @ecojan

pkg/resources/kafka/poddisruptionbudget.go
```diff
@@ -0,0 +1,108 @@
...
+	budget := 0
+
+	// treat percentage budget
+	if strings.HasSuffix(disruptionBudget, "%") {
+		percentage, err := strconv.ParseFloat(disruptionBudget[:len(disruptionBudget)-1], 4)
+		if err != nil {
+			log.Error(err, "error occurred during parsing the disruption budget")
+			return intstr.FromInt(-1), err
+		} else {
+			budget = int(math.Floor((percentage * float64(brokers)) / 100))
```
stoader, line 93: Since PDB allows percentage as well should just return `intstr.FromString(disruptionBudget)` here and use that in `PodDisruptionBudgetSpec`.

baluchicken: Sadly no, we have to compute that since it is only supported when the scale resource is implemented in the CRD.
https://kubernetes.io/docs/tasks/run-application/configure-pdb/#arbitrary-controllers-and-selectors

stoader (changes requested):

charts/kafka-operator/templates/crds.yaml
```diff
@@ -3493,6 +3493,18 @@ spec:
                     type: object
                   type: array
               type: object
+            disruptionBudget:
+              description: DisruptionBudget defines the configuration for PodDisruptionBudget
+              properties:
+                budget:
+                  description: The budget to set for the PDB, can either be static
+                    number or a percentage
+                  pattern: '[0-9]+(\.[0-9]+)?(%)?'
```
stoader, line 3502: How about this regex: `^[0-9]+$|^[0-9]{1,2}%$|^100%$` to not allow invalid percentage values such as 1000%

config/base/crds/kafka.banzaicloud.io_kafkaclusters.yaml
```diff
@@ -3491,6 +3491,18 @@ spec:
                     type: object
                   type: array
               type: object
+            disruptionBudget:
+              description: DisruptionBudget defines the configuration for PodDisruptionBudget
+              properties:
+                budget:
+                  description: The budget to set for the PDB, can either be static
+                    number or a percentage
+                  pattern: '[0-9]+(\.[0-9]+)?(%)?'
```
stoader, line 3500: How about this regex: `^[0-9]+$|^[0-9]{1,2}%$|^100%$` to not allow invalid percentage values such as 1000%

charts/kafka-operator/templates/crds.yaml
ecojan: All for not allowing invalid percentage, do we want to invalidate "33.33%" though?

stoader: Do you see any use case when fractional percentage would be useful?

ecojan: How about
`^[0-9]+$|^[0-9]{1,2}(\.[0-9]+%)?$|^100%$` ?

ecojan: You might be right, in terms of use cases, fractional percentages would only be helpful here when going over 1000, otherwise they are rendered useless by the floor function

ecojan: Let's go without fractional and change it if we need to.

stoader: that's what I was thinking as well. The fractional percentage would have a noticeable impact only in case of broker count over 1000, even than compared to the total number of brokers the diff would negligible

stoader (approved):

baluchicken (approved):

PR #495, opened 2020-11-07
[COSMETIC] - Remove leading spaces from cruise-control config files and update log4j log format

amuraru:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

- Remove leadings spaces from generated CruiseControl `cruisecontrol.properties` and `log4j.properties` files
- Update cruise-control log4j log format to more canonical `[%d] %p %m (%c)%n` 

### Why?
Cosmetic changes

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/sdk/v1beta1/kafkacluster_types.go
+++ b/pkg/sdk/v1beta1/kafkacluster_types.go
@@ -601,10 +601,10 @@ func (cConfig *CruiseControlConfig) GetCCLog4jConfig() string {
 		return cConfig.Log4jConfig
 	}
 	return `log4j.rootLogger = INFO, FILE
-    log4j.appender.FILE=org.apache.log4j.FileAppender
-    log4j.appender.FILE.File=/dev/stdout
-    log4j.appender.FILE.layout=org.apache.log4j.PatternLayout
-    log4j.appender.FILE.layout.conversionPattern=%-6r [%15.15t] %-5p %30.30c %x - %m%n`
+log4j.appender.FILE=org.apache.log4j.FileAppender
+log4j.appender.FILE.File=/dev/stdout
+log4j.appender.FILE.layout=org.apache.log4j.PatternLayout
+log4j.appender.FILE.layout.conversionPattern=%-6r [%15.15t] %-5p %30.30c %x - %m%n`
 }
 
 // GetImage returns the used image for Prometheus JMX exporter
--- a/pkg/resources/cruisecontrol/configmap.go
+++ b/pkg/resources/cruisecontrol/configmap.go
@@ -41,10 +41,10 @@ func (r *Reconciler) configMap(clientPass, capacityConfig string) runtime.Object
 		),
 		Data: map[string]string{
 			"cruisecontrol.properties": r.KafkaCluster.Spec.CruiseControlConfig.Config + fmt.Sprintf(`
-    # The Kafka cluster to control.
-    bootstrap.servers=%s:%d
-    # The zookeeper connect of the Kafka cluster
-    zookeeper.connect=%s
+# The Kafka cluster to control.
+bootstrap.servers=%s:%d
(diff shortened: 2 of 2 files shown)


pkg/sdk/v1beta1/kafkacluster_types.go
```diff
@@ -601,10 +601,10 @@ func (cConfig *CruiseControlConfig) GetCCLog4jConfig() string {
 		return cConfig.Log4jConfig
 	}
 	return `log4j.rootLogger = INFO, FILE
-    log4j.appender.FILE=org.apache.log4j.FileAppender
-    log4j.appender.FILE.File=/dev/stdout
-    log4j.appender.FILE.layout=org.apache.log4j.PatternLayout
-    log4j.appender.FILE.layout.conversionPattern=%-6r [%15.15t] %-5p %30.30c %x - %m%n`
+log4j.appender.FILE=org.apache.log4j.FileAppender
```
adamantal, line 604: Even though this is not what this PR is for, this log4j.properties seems odd to me.
Combining a file appender with `/dev/stdout` is suboptimal as log4j has a `ConsoleAppender` which is built for this particular purpose (supposedly containing more optimisation regarding pushing data to stdout).

pkg/sdk/v1beta1/kafkacluster_types.go
```diff
@@ -601,10 +601,10 @@ func (cConfig *CruiseControlConfig) GetCCLog4jConfig() string {
...
+log4j.appender.FILE.File=/dev/stdout
+log4j.appender.FILE.layout=org.apache.log4j.PatternLayout
+log4j.appender.FILE.layout.conversionPattern=[%d] %p %m (%c)%n`
```
adamantal, line 607: Could you please add a sample line with this pattern (before/after)?

adamantal (commented):
Hi! 
I had a two minor remarks regarding this pull request, if you don't mind (find them inline).

pkg/sdk/v1beta1/kafkacluster_types.go
amuraru: This is fair - it's probably better to move this change is a separate pr

amuraru: Added https://github.com/banzaicloud/kafka-operator/issues/496 to track this @adamantal

pkg/sdk/v1beta1/kafkacluster_types.go
adamantal: I was fine changing the pattern to `[%d] %p %m (%c)%n`, I was just curious how the new pattern actually looks like.

Based on the [log4j docs](https://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/PatternLayout.html) logs are displayed like this:
```
[1999-11-27 15:49:37,459] INFO Log entry (org.apache.log4j.examples.Sort)
``` 

While the currently used pattern is:
```
256   [main          ] INFO  org.apache.log4j.examples.Sort      - Log entry
```

I think the one you originally wanted to use is more human readable, so I'd suggest to use that one, instead of keeping the existing one.

pkg/sdk/v1beta1/kafkacluster_types.go
adamantal: Thanks!

pkg/sdk/v1beta1/kafkacluster_types.go
amuraru: I'll do that in a different PR

adamantal: Okay, thanks!

adamantal (approved):

baluchicken (approved):

PR #502, opened 2020-11-19
Let CC create all its topics

schultheiszn:
| Q               | A
| --------------- | ---
| New feature?   | yes|
| API breaks?     | no|
| Deprecations?   | no|

### What's in this PR?
Let CC create its own topics if cruise.control.metrics.topic.auto.create=true is present
If property isn't configured act like before

### Checklist

- [x] Implementation tested

### DIFF
--- a/config/base/crds/kafka.banzaicloud.io_kafkaclusters.yaml
+++ b/config/base/crds/kafka.banzaicloud.io_kafkaclusters.yaml
@@ -1644,7 +1644,6 @@ spec:
                         type: array
                         x-kubernetes-list-map-keys:
                         - containerPort
-                        - protocol
                         x-kubernetes-list-type: map
                       readinessProbe:
                         description: 'Periodic probe of container service readiness.
--- a/config/base/crds/kafka.banzaicloud.io_kafkaclusters.yaml
+++ b/config/base/crds/kafka.banzaicloud.io_kafkaclusters.yaml
@@ -1644,6 +1644,7 @@ spec:
                         type: array
                         x-kubernetes-list-map-keys:
                         - containerPort
+                        - protocol
                         x-kubernetes-list-type: map
                       readinessProbe:
                         description: 'Periodic probe of container service readiness.
--- a/config/samples/kafkacluster_with_internal_ssl_vault.yaml
+++ b/config/samples/kafkacluster_with_internal_ssl_vault.yaml
@@ -12,6 +12,7 @@ spec:
   clusterImage: "wurstmeister/kafka:2.12-2.1.0"
   readOnlyConfig: |
     auto.create.topics.enable=false
+    cruise.control.metrics.topic.auto.create=true
     authorizer.class.name=kafka.security.auth.SimpleAclAuthorizer
     allow.everyone.if.no.acl.found=false
     ssl.secure.random.implementation=SHA1PRNG
--- a/config/samples/kafkacluster_with_internal_ssl_vault.yaml
+++ b/config/samples/kafkacluster_with_internal_ssl_vault.yaml
@@ -12,7 +12,13 @@ spec:
   clusterImage: "banzaicloud/kafka:2.13-2.6.0-bzc.1"
   readOnlyConfig: |
     auto.create.topics.enable=false
+    metric.reporters=com.linkedin.kafka.cruisecontrol.metricsreporter.CruiseControlMetricsReporter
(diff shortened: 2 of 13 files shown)


amuraru:
these topics are already created by CC itself IIUC. Is there any strong reason for doing this from the operator?

stoader:
> these topics are already created by CC itself IIUC. Is there any strong reason for doing this from the operator?

Currently CC only creates the topics, however it doesn't update the topics when changes are made to the settings of these topics via the CC config. The intent is to apply config changes made to CC topics.

amuraru:
> Currently CC only creates the topics, however it doesn't update the topics when changes are made to the settings of these topics via the CC config. The intent is to apply config changes made to CC topics.

that makes sense! thanks

schultheiszn:
> > Currently CC only creates the topics, however it doesn't update the topics when changes are made to the settings of these topics via the CC config. The intent is to apply config changes made to CC topics.
> 
> that makes sense! thanks

Have just discussed with the whole team the topic and after some code digging we decided to go with CC's built-in topic management and an adjusted operator setup

config/samples/kafkacluster_with_internal_ssl_vault.yaml
```diff
@@ -12,6 +12,7 @@ spec:
   clusterImage: "wurstmeister/kafka:2.12-2.1.0"
```
stoader, line 12:
```suggestion
  clusterImage: "banzaicloud/kafka:2.13-2.6.0-bzc.1"
```

pkg/resources/cruisecontrol/topicManager.go
```diff
@@ -68,6 +71,17 @@ func newCruiseControlTopic(cluster *v1beta1.KafkaCluster) *v1alpha1.KafkaTopic {
 }
 
 func generateCCTopic(cluster *v1beta1.KafkaCluster, client client.Client, log logr.Logger) error {
+	readOnlyConfigProperties := util.ParsePropertiesFormat(cluster.Spec.ReadOnlyConfig)
+
+	// for compatibility reasons the only case when we let CC to create its own kafka topics is
+	// when we enable the creation explicitly
+	if propertyValue, present := readOnlyConfigProperties[ccMetricTopicAutoCreate]; present {
+		if autoCreate, _ := strconv.ParseBool(propertyValue); autoCreate {
```
stoader, line 79: Don't ignore the error returned by `strconv.ParseBool` but rather pass it to the caller.

stoader (changes requested):

config/base/crds/kafka.banzaicloud.io_kafkaclusters.yaml
```diff
@@ -1644,6 +1644,7 @@ spec:
                         type: array
                         x-kubernetes-list-map-keys:
                         - containerPort
+                        - protocol
```
baluchicken, line 1647:
```suggestion
```
Due to a bug in the generator tool this should be removed, otherwise the kubernetes api server will not allow to install this to a k8s cluster v 1.18+

pkg/resources/cruisecontrol/cruisecontrol.go
```diff
@@ -84,9 +84,9 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
 
 	if r.KafkaCluster.Spec.CruiseControlConfig.CruiseControlEndpoint == "" {
 
-		err := generateCCTopic(r.KafkaCluster, r.Client, log)
+		err := generateCCTopic(r.KafkaCluster, r.Client, log.WithName("generateCCTopic"))
 		if err != nil {
-			k8sutil.UpdateCRStatus(r.Client, r.KafkaCluster, v1beta1.CruiseControlTopicNotReady, log)
+			_ = k8sutil.UpdateCRStatus(r.Client, r.KafkaCluster, v1beta1.CruiseControlTopicNotReady, log)
```
stoader, line 89: If `k8sutil.UpdateCRStatus` returns an error than return a combined error that contains both errors.

```
errors.Combine(err, ...)
```

stoader (changes requested):

stoader (dismissed):

baluchicken (approved):
LGTM thx @schultheiszn

bonifaido (approved):

adamantal (approved):
LGTM, thanks!

PR #503, opened 2020-11-20
Update SSL and listener related configs dynamically

adamantal:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

This PR moves changes specific already existing per-boker Kafka configurations to be updated dynamically instead of doing a broker restart.
According to the 2.6.0 Kafka [documentation](https://kafka.apache.org/documentation/#dynamicbrokerconfigs) these configs are labelled as per-broker:
- listener related configs
`listener.security.protocol.map`, `listeners` can be set dynamically (but `inter.broker.listener.name` could be not!)
- SSL related configs
`ssl.keystore.location`, `ssl.truststore.location`, `ssl.keystore.password`, `ssl.truststore.password`, `ssl.client.auth` can all be set dynamically
- advertised listener
`advertised.listeners` can be set dynamically
We should move these to the dynamically updated configurations in the controller.

### Why?

This is a faster way to update the configuration: less downtime, no unnecessary broker restarts etc.

### Checklist

- [ ] Implementation tested
- [x] Unit tests added and passing

### DIFF
--- a/pkg/resources/kafka/configmap.go
+++ b/pkg/resources/kafka/configmap.go
@@ -34,7 +34,7 @@ import (
 )
 
 var kafkaConfigTemplate = `
-{{ .ListenerConfig }}
+{{ .ListenerReadOnlyConfig }}
 {{ if .ControlPlaneListener }}
 control.plane.listener.name={{ .ControlPlaneListener }}
 {{ end }}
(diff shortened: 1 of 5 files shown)


adamantal:
After thorough testing turned out that this approach has multiple problems:
- since dynamic configs are treated totally separately, the static configs do not contain any listener and SSL related configs
- this causes Kafka not to come alive due to invalid configs
- to fix this error we need to give brokers an initial valid configs, but do not update them later, which requires some additional complex logic, like introducing an "init" phase in the brokers where per broker configs are also given through the ConfigMap

I'll come up with another approach, but this just does not seem to be the right direction.

PR #504, opened 2020-11-22
Accepting existing serviceAccountName for both AuthProxy and Operator

ghahramani:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Accepting existing `serviceAccountName` for both `operator` and `authProxy`

### Why?
This gives the advantages of adding more permissions or apiGroups in case you have a sidecar container and you need to add those to your pods, for instance, Istio injection.

### DIFF
--- a/charts/kafka-operator/templates/_helpers.tpl
+++ b/charts/kafka-operator/templates/_helpers.tpl
@@ -46,9 +46,9 @@ Compute operator deployment serviceAccountName key
 Compute operator prometheus metrics auth proxy service account
 */}}
 {{- define "operator.metricsAuthProxy.serviceAccountName" -}}
-{{- if .Values.prometheusMetrics.authProxy.serviceAccount.name -}}
+{{- if .Values.prometheusMetrics.authProxy.serviceAccount.create -}}
 {{ default "default" .Values.prometheusMetrics.authProxy.serviceAccount.name }}
 {{- else -}}
-{{- printf "%s" "default" }}
+{{ default "default" .Values.prometheusMetrics.authProxy.serviceAccount.name }}
 {{- end -}}
 {{- end -}}
--- a/charts/kafka-operator/templates/_helpers.tpl
+++ b/charts/kafka-operator/templates/_helpers.tpl
@@ -35,10 +35,10 @@ Create chart name and version as used by the chart label.
 Compute operator deployment serviceAccountName key
 */}}
 {{- define "operator.serviceAccountName" -}}
-{{- if .Values.operator.serviceAccount.name -}}
+{{- if .Values.operator.serviceAccount.create -}}
 {{ default "default" .Values.operator.serviceAccount.name }}
 {{- else -}}
-{{- printf "%s" "default" }}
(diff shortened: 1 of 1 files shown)


baluchicken:
Hi, @ghahramani thanks for the PR. Can you please elaborate how this PR solves your issue? On the other hand what do you mean by `this gives  advantages of adding more permissions or apiGroups in case you have a sidecar container and you need to add those to your pods, for instance, Istio injection.` Thanks.

ghahramani:
Yeah, no worries, basically, when you use Istio injection, it needs some permission like privileged container, network admin, and the rest, and if your cluster uses RBAC, it prevents the Istio sidecar container to start up as they don't have that permission (Pod does not have that permission), so to allow to inject the container in that pods and works with no problem, you need to give some specific permissions. That's where the service account comes to play. This is just one example that you want to use your own service account rather than using the created one via chart.

charts/kafka-operator/templates/_helpers.tpl
```diff
@@ -35,7 +35,7 @@ Create chart name and version as used by the chart label.
 Compute operator deployment serviceAccountName key
 */}}
 {{- define "operator.serviceAccountName" -}}
-{{- if .Values.operator.serviceAccount.create -}}
+{{- if .Values.operator.serviceAccount.name -}}
 {{ default "default" .Values.operator.serviceAccount.name }}
 {{- else -}}
 {{- printf "%s" "default" }}
```
baluchicken, line 41:
```suggestion
{{- if .Values.operator.serviceAccount.create -}}
{{ default "default" .Values.operator.serviceAccount.name }}
{{- else -}}
{{ default "default" .Values.operator.serviceAccount.name }}
```

baluchicken (commented):
Sorry for the delayed response I just reviewed your PR.

charts/kafka-operator/templates/_helpers.tpl
```diff
@@ -38,15 +38,15 @@ Compute operator deployment serviceAccountName key
...
 
 {{/*
 Compute operator prometheus metrics auth proxy service account
 */}}
 {{- define "operator.metricsAuthProxy.serviceAccountName" -}}
-{{- if .Values.prometheusMetrics.authProxy.serviceAccount.create -}}
+{{- if .Values.prometheusMetrics.authProxy.serviceAccount.name -}}
 {{ default "default" .Values.prometheusMetrics.authProxy.serviceAccount.name }}
 {{- else -}}
 {{- printf "%s" "default" }}
```
baluchicken, line 52:
```suggestion
{{- if .Values.prometheusMetrics.authProxy.serviceAccount.create -}}
{{ default "default" .Values.prometheusMetrics.authProxy.serviceAccount.name }}
{{- else -}}
{{ default "default" .Values.prometheusMetrics.authProxy.serviceAccount.name }}
```

baluchicken (approved):

stoader (approved):

PR #505, opened 2020-11-23
Multiple external listener

baluchicken:
| Q               | A
| --------------- | ---
| Bug fix?        | no|
| New feature?    |yes|
| API breaks?     | no|
| Deprecations?   | no|

### What's in this PR?

This PR adds support for multiple external listener endpoints for Kafka.
This is a long limitation of the operator which allowed only one external listener per cluster.
Since Kafka is not handling advertised.listeners config as a read-only config any more this listener setting will only work when creating a new cluster with the operator. Adding a new listener to a running cluster will not work since it will trigger a rolling update which will fail. We are working on a different PR which treats these settings as a dynamic config to overcome this limitation. Also we know that we should allow to support multiple settings per listener, that will come soon as well.

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/resources/kafka/kafka.go
+++ b/pkg/resources/kafka/kafka.go
@@ -652,7 +652,7 @@ func (r *Reconciler) reconcileKafkaPod(log logr.Logger, desiredPod *corev1.Pod) 
 		//Since toleration does not support patchStrategy:"merge,retainKeys", we need to add all toleration from the current pod if the toleration is set in the CR
 		if len(desiredPod.Spec.Tolerations) > 0 {
 			desiredPod.Spec.Tolerations = append(desiredPod.Spec.Tolerations, currentPod.Spec.Tolerations...)
-			uniqueTolerations := make([]corev1.Toleration, len(desiredPod.Spec.Tolerations), 0)
+			uniqueTolerations := make([]corev1.Toleration, 0, len(desiredPod.Spec.Tolerations))
 			keys := make(map[corev1.Toleration]bool)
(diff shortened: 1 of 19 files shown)


stoader (approved):

pkg/resources/kafka/kafka.go
```diff
@@ -655,7 +652,7 @@ func (r *Reconciler) reconcileKafkaPod(log logr.Logger, desiredPod *corev1.Pod)
 		//Since toleration does not support patchStrategy:"merge,retainKeys", we need to add all toleration from the current pod if the toleration is set in the CR
 		if len(desiredPod.Spec.Tolerations) > 0 {
 			desiredPod.Spec.Tolerations = append(desiredPod.Spec.Tolerations, currentPod.Spec.Tolerations...)
-			uniqueTolerations := []corev1.Toleration{}
+			uniqueTolerations := make([]corev1.Toleration, len(desiredPod.Spec.Tolerations), 0)
```
adamantal, line 655:
```suggestion
			uniqueTolerations := make([]corev1.Toleration, 0, len(desiredPod.Spec.Tolerations))
```

adamantal (commented):
LGTM, one suggestion inline.

pkg/resources/kafka/kafka.go
baluchicken: Thanks nice catch this is a serious error.

adamantal (approved):

schultheiszn (approved):

PR #506, opened 2020-11-24
Per broker configs should not trigger rolling upgrades

adamantal:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

- Introduced `PerBrokerConfigurationState` in the `BrokerStatus` field
- If ConfigMap is updated and only per-broker configs are changed, rolling upgrade is not triggered

### Why?

Per broker configs can be updated through an API call which is faster and does less interruption in the cluster than restarting a broker.

### Additional context

The following scenarios should pass:
 - [x] Changing `Config` field in brokerConfigGroup (per-broker config) does not trigger rolling upgrade (no regression of existing behaviour)
 - [x] Changing `InternalListener` or `ExternalListener` config which does not alter `PodSpec` should not trigger rolling upgrade
 - [x] Changing SSL related configs which does not alter `PodSpec` should not trigger rolling upgrade
 - [x] In case of each valid per broker config change the `PerBrokerConfigurationState` status should change from out of sync to in sync.
 - [x] If user attempts to dynamically update to an invalid config, the status should stay in out of sync state (and the reconciler should not stuck in a loop) \*

\* Only provide validation from Kafka

### Checklist

- [x] Implementation tested


pkg/k8sutil/perbrokerconfig.go
```diff
@@ -0,0 +1,40 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package k8sutil
+
+// these configurations will not trigger rolling upgrade when updated
+var PerBrokerConfigs = []string{
```
stoader, line 18: Not sure if the `k8sutil` package is the right place for this list as it is specific to Kafka brokers. I'd place it in an internal broker config package an make it private. Since this list should have no duplicates probable is better to store it in a `map[string]struct{}`

stoader (changes requested):

pkg/k8sutil/perbrokerconfig.go
adamantal: Reconciling k8s resources are performed through `k8sutil/resources.go` 's `Reconcile` function. This function depends on the Kafka controller package, because this function sets the config in sync/out of sync states. This causes a circular dependency when I moved it into the kafka package. Where should it be put then?

stoader: I'd move out the updating of the broker status when the config map changes into `kafka.go` where the config map change originates from.

adamantal: Moved the related parts to the kafka package.

adamantal:
Updated the PR and tested it against the scenarios mentioned above.
Though the scope has shrunk since I started working on this, I think this feature provides user substantial benefits.

This patch roughly includes:
- some configurations are marked as dynamically updatable (like `listener.security.protocol.map`, `ssl.client.auth`, `listeners`, `advertised.listeners`). Updating these would not trigger rolling upgrade by itself (but if the spec of the broker pods have been changed, it will, which is 90% of the cases)
- these configs are still reflected in the configmap, but the pod will not be recreated unless a read-only config has been changed
- added a status to track the per-broker config status: in sync/out of sync
- added pre-validation before updating the per-broker configs: 
  - though honestly it is not very effective, in a lot of cases validation passes, but the broker will remain out of sync per-broker config state as broker will throw error upon the change
  - if the validation fails, the per-broker config status will be changed to error
- added special case for `listener.security.protocol.map`: we will trigger rolling upgrade based on the change, but not always
- moved post configmap update handling back to kafka package
- moved the code responsible for dynamically updating the broker config to a separate go file
- added some unit tests

Please feel free to provide your opinion the patch. This is a medium sized change, so let's be as cautious as possible.

pkg/k8sutil/status.go
```diff
@@ -94,6 +94,8 @@ func generateBrokerState(brokerIds []string, cluster *banzaicloudv1beta1.KafkaCl
 				cluster.Status.BrokersState = map[string]banzaicloudv1beta1.BrokerState{brokerId: {GracefulActionState: state}}
 			case banzaicloudv1beta1.ConfigurationState:
 				cluster.Status.BrokersState = map[string]banzaicloudv1beta1.BrokerState{brokerId: {ConfigurationState: s}}
+			case banzaicloudv1beta1.PerBrokerConfigurationState:
```
stoader, line 97: I believe this code block
```
for _, brokerId := range brokerIds {
    if cluster.Status.BrokersState == nil {
        ...
    } else if val, ok := cluster.Status.BrokersState[brokerId]; ok {
        ...
    } else {
        ...
    }
}
```

can be simplified as follows:
```
brokersState := cluster.Status.BrokersState
if brokersState == nil {
    brokersState = make(map[string]banzaicloudv1beta1.BrokerState, len(brokerIds))
}

for _, brokerId := range brokerIds {
    brokerState := brokersState[brokerId]
    switch s := state.(type) {
			case banzaicloudv1beta1.RackAwarenessState:
				brokerState[brokerId] = banzaicloudv1beta1.BrokerState{RackAwarenessState: s}
			case banzaicloudv1beta1.GracefulActionState:
				brokerState[brokerId] = banzaicloudv1beta1.BrokerState{GracefulActionState: s}
			case map[string]banzaicloudv1beta1.GracefulActionState:
				state := s[brokerId]
				brokerState[brokerId] = banzaicloudv1beta1.BrokerState{GracefulActionState: state}
			case banzaicloudv1beta1.ConfigurationState:
				brokerState[brokerId] = banzaicloudv1beta1.BrokerState{ConfigurationState: s}
			case banzaicloudv1beta1.PerBrokerConfigurationState:
				brokerState[brokerId] = banzaicloudv1beta1.BrokerState{PerBrokerConfigurationState: s}
			case map[string]banzaicloudv1beta1.VolumeState:
				gracefulActionState := banzaicloudv1beta1.GracefulActionState{
					VolumeStates: s,
				}
				brokerState[brokerId] = banzaicloudv1beta1.BrokerState{GracefulActionState: gracefulActionState}
			case map[string]map[string]banzaicloudv1beta1.VolumeState:
				state := s[brokerId]
				gracefulActionState := banzaicloudv1beta1.GracefulActionState{
					VolumeStates: state,
				}
				brokerState[brokerId] = banzaicloudv1beta1.BrokerState{GracefulActionState: gracefulActionState}
            default: errors.New("unsupported broker state type")
    }
    brokersState[brokerId] = brokerState
}

cluster.Status.BrokersState = brokersState
```

pkg/resources/kafka/configmap.go
```diff
@@ -250,3 +255,116 @@ func (r Reconciler) generateBrokerConfig(id int32, brokerConfig *v1beta1.BrokerC
 
 	return strings.Join(completeConfig, "\n")
 }
+
+func (r *Reconciler) postConfigMapReconcile(log logr.Logger, desired *corev1.ConfigMap) error {
```
stoader, line 259: This method updates broker status so updateBrokerStatus might be a more descriptive name

pkg/resources/kafka/configs.go
```diff
@@ -0,0 +1,205 @@
...
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/banzaicloud/kafka-operator/api/v1beta1"
+	"github.com/banzaicloud/kafka-operator/pkg/errorfactory"
+	"github.com/banzaicloud/kafka-operator/pkg/k8sutil"
+	"github.com/banzaicloud/kafka-operator/pkg/kafkaclient"
+	"github.com/banzaicloud/kafka-operator/pkg/util"
+)
+
+const securityProtocolMap = "listener.security.protocol.map"
```
stoader, line 33:
```suggestion
const securityProtocolMapConfig = "listener.security.protocol.map"
```

stoader (changes requested):

pkg/resources/kafka/configs.go
```diff
@@ -0,0 +1,205 @@
...
+		err = kClient.AlterClusterWideConfig(util.ConvertMapStringToMapStringPointer(parsedClusterWideConfig), false)
+		if err != nil {
+			return errors.WrapIf(err, "could not alter cluster wide broker config")
+		}
+	}
+
+	return nil
+}
+
+func getBrokerConfigsFromConfigMap(configMap *corev1.ConfigMap) map[string]string {
```
stoader, line 181: This function seems to perform the same thing as `util.ParsePropertiesFormat`.

pkg/resources/kafka/configmap.go
```diff
@@ -250,3 +255,116 @@ func (r Reconciler) generateBrokerConfig(id int32, brokerConfig *v1beta1.BrokerC
...
+OUTERLOOP:
+	for _, config := range touchedConfigs {
+		// Security protocol cannot be updated for existing listener
+		// a rolling upgrade should be triggered in this case
+		if config == securityProtocolMap {
+			// added or deleted config is ok
+			if currentConfigs[securityProtocolMap] == "" || desiredConfigs[securityProtocolMap] == "" {
+				continue
+			}
+			currentListenerProtocolMap := make(map[string]string, 0)
```
stoader, line 309:
```suggestion
			currentListenerProtocolMap := make(map[string]string)
```

pkg/resources/kafka/configmap.go
```diff
@@ -250,3 +255,116 @@ func (r Reconciler) generateBrokerConfig(id int32, brokerConfig *v1beta1.BrokerC
...
+			desiredListenerProtocolMap := make(map[string]string, 0)
```
stoader, line 310:
```suggestion
			desiredListenerProtocolMap := make(map[string]string)
```

pkg/resources/kafka/configmap.go
```diff
@@ -250,3 +255,116 @@ func (r Reconciler) generateBrokerConfig(id int32, brokerConfig *v1beta1.BrokerC
...
+			for _, listenerConfig := range strings.Split(currentConfigs[securityProtocolMap], ",") {
+				listenerKeyValue := strings.Split(listenerConfig, ":")
```
stoader, line 312: I'd rename `securityProtocolMap` to `securityProtocolMapConfig` and `listenerKeyValue` to `listenerProtocolMap" to make clearer what does variables are used for.

pkg/resources/kafka/configmap.go
```diff
@@ -250,3 +255,116 @@ func (r Reconciler) generateBrokerConfig(id int32, brokerConfig *v1beta1.BrokerC
...
+				if len(listenerKeyValue) != 2 {
+					continue
+				}
+				currentListenerProtocolMap[listenerKeyValue[0]] = listenerKeyValue[1]
```
stoader, line 316: Trim spaces on key and value.

pkg/resources/kafka/configmap.go
```diff
@@ -250,3 +255,116 @@ func (r Reconciler) generateBrokerConfig(id int32, brokerConfig *v1beta1.BrokerC
...
+			}
+			for _, listenerConfig := range strings.Split(desiredConfigs[securityProtocolMap], ",") {
+				listenerKeyValue := strings.Split(listenerConfig, ":")
+				if len(listenerKeyValue) != 2 {
+					continue
+				}
+				desiredListenerProtocolMap[listenerKeyValue[0]] = listenerKeyValue[1]
+			}
+
+			for protocolName, desiredListenerProtocol := range desiredListenerProtocolMap {
```
stoader, line 326: There is no need for the `for` loop. The same verification can be done in the `for` loop above instead of populating `desiredListenerProtocolMap`

```
```

stoader (changes requested):

pkg/resources/kafka/configmap.go
```diff
@@ -250,3 +255,116 @@ func (r Reconciler) generateBrokerConfig(id int32, brokerConfig *v1beta1.BrokerC
...
+	}
+
+	return nil
+}
+
+func shouldRefreshOnlyPerBrokerConfigs(currentConfigs, desiredConfigs map[string]string, log logr.Logger) bool {
+	touchedConfigs := collectTouchedConfigs(currentConfigs, desiredConfigs, log)
+
+OUTERLOOP:
+	for _, config := range touchedConfigs {
```
stoader, line 301: This part could be refactored as follows to make it more readable:
```
touchedConfigs := collectTouchedConfigs(currentConfigs, desiredConfigs, log)

for _, perBrokerConfig := range perBrokerConfigs {
  for _, config := range touchedConfigs {
           if config != perBrokerConfig {
              return false // broker config that can not be updated dynamicaaly has been modified
           }

 		   if listenersSecurityProtocolChanged(...) {
                return false
           }
  }
}

return true
```

stoader (changes requested):

pkg/resources/kafka/configs.go
adamantal: Indeed, did not see this function. Remove the function and the corresponding tests.

pkg/resources/kafka/configmap.go
adamantal: Renamed

pkg/resources/kafka/configmap.go
adamantal: Actually if we return with a set data structure from `collectTouchedConfigs` we can do this much simpler. I come up with the following solution:
```go
	touchedConfigs := collectTouchedConfigs(currentConfigs, desiredConfigs, log)
	if _, ok := touchedConfigs[securityProtocolMapConfigName]; ok {
		if listenersSecurityProtocolChanged(currentConfigs, desiredConfigs) {
			return false
		}
	}
	for _, perBrokerConfig := range perBrokerConfigs {
		delete(touchedConfigs, perBrokerConfig)
	}
	return len(touchedConfigs) == 0
```
Assuming `collectTouchedConfigs` returns with `map[string]struct{}`

pkg/resources/kafka/configmap.go
```diff
@@ -275,3 +280,113 @@ func (r Reconciler) generateBrokerConfig(id int32, brokerConfig *v1beta1.BrokerC
 
 	return strings.Join(completeConfig, "\n")
 }
+
+func (r *Reconciler) updateBrokerStatus(log logr.Logger, desired *corev1.ConfigMap) error {
```
baluchicken, line 284: I think this should be inside the k8sreconciler method. We already identified the broker related configmaps during an update, all these things can be simplified with ~4 lines of code if we place this inside that method. Moreover, it removes the k8s client usage from a resource related file. If we check all other resource related file their scope is to create an object only, without any k8s interaction.

baluchicken (commented):
Huge thanks for the tests.

pkg/resources/kafka/configmap.go
adamantal: Agreed. Resolve this by putting that code back, and moving config related functions to `common.go` in `kafka/util` package to prevent import cycle being created.

stoader (dismissed):

stoader (approved):

baluchicken (approved):

PR #509, opened 2020-11-27
Add nodeport support for externalListener

baluchicken:
| Q               | A
| --------------- | ---
| Bug fix?        | no|
| New feature?    |yes|
| API breaks?     | no|
| Deprecations?   | no|

### What's in this PR?

This PR adds support to use nodeport as an external Listener option.
Adds a new cluster config which describes how this can be used.

### Checklist

- [x] Implementation tested


stoader (approved):

chrisgacsal (approved):
LGTM!

config/base/crds/kafka.banzaicloud.io_kafkaclusters.yaml
```diff
@@ -3916,13 +3942,41 @@ spec:
                     description: ExternalListenerConfig defines the external listener
                       config for Kafka
                     properties:
+                      accessMethod:
```
amuraru, line 3945: would it make sense to rename this to `externalAccessMethod` to denote that it's about exposing kafka externally through this parameter? as opposed to internal access method which already exists via headless service

config/base/crds/kafka.banzaicloud.io_kafkaclusters.yaml
```diff
@@ -3916,13 +3942,41 @@ spec:
...
+                        description: accessMethod defines the method which the external
+                          listener is exposed through. Two types are supported LoadBalancer
+                          and NodePort. The recommended and default is the LoadBalancer.
+                          NodePort should be used in Kubernetes environments with
+                          no support for provisioning Load Balancers.
+                        enum:
+                        - LoadBalancer
+                        - NodePort
```
amuraru, line 3953: would it make sense to also add `None` as an option here? to disable external access altogether?

stoader: Removing all external listeners would disable external access altogether. kafka-operator should remove all K8s that correspond to an external listener when that is removed from the CR.

config/base/crds/kafka.banzaicloud.io_kafkaclusters.yaml
stoader: The `ExternalListenerConfig` already denotes that controls the config of exposing Kafka outside the Kubernetes clusters. The `accessMethod` property defines the `how` to expose Kafka outside the Kubernetes cluster.

amuraru: makes sense! I see now. this is actually under externalListener

adamantal:
@baluchicken I see that this PR adds the following entry to the go.mod file:
```
github.com/banzaicloud/kafka-operator/pkg/sdk v0.4.1
```

I think we shouldn't use that. Should we fix this in master?

PR #516, opened 2020-12-03
fix go.mod remove operator sdk

baluchicken:
Remove unnecessary `github.com/banzaicloud/kafka-operator/pkg/sdk` from the go.mod file. It is left there by accident. IDE added that automatically.


adamantal (approved):
Thanks for fixing this!

schultheiszn (approved):

PR #519, opened 2020-12-10
Introduce listener status field

adamantal:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

Introducing listener status field which contains listener addresses. Client application can use this to connect to the cluster using either an internal or and external listener.

### Why?

There is a wide variety of configurations that you can set and currently it is nontrivial for a client application to obtain listener information based on the configs. Writing the addresses to the status field is a handy choice and ease the integration of the clients to the cluster.

### Checklist

- [x] Implementation tested

### TODO
- [x] Implement the actual status update
- [x] Test it on a remote cluster


pkg/sdk/v1beta1/kafkacluster_types.go
```diff
@@ -353,6 +354,21 @@ type CommonListenerSpec struct {
...
 }
 
+// ListenerStatuses defines the internal and external listener addresses
+type ListenerStatuses struct {
+	InternalListeners ListenerStatusMap `json:"internalListeners,omitempty"`
+	ExternalListeners ListenerStatusMap `json:"externalListeners,omitempty"`
+}
+
+// ListenerStatus objects can be looked up in ListenerStatusMap by listener name
+type ListenerStatusMap map[string]ListenerStatus
```
stoader, line 364: If there are no methods implemented for `ListenerStatusMap` than there is no need for introducing a new type.

```
type ListenerStatuses struct {
	InternalListeners map[string] ListenerStatus `json:"internalListeners,omitempty"`
	ExternalListeners map[string] ListenerStatus `json:"externalListeners,omitempty"`
}
```

adamantal:
Tested on a remote cluster.
CR:
```yaml
spec:
  ...
  ingressController: envoy
  listenersConfig:
    externalListeners:
      - type: "plaintext"
        name: "external"
        externalStartingPort: 19090
        containerPort: 9094
    internalListeners:
      - type: "plaintext"
        name: "internal"
        containerPort: 29092
        usedForInnerBrokerCommunication: true
      - type: "plaintext"
        name: "controller"
        containerPort: 29093
        usedForInnerBrokerCommunication: false
        usedForControllerCommunication: true
```
and the status field:
```yaml
status:
  ...
  listenerStatuses:
    externalListeners:
      external:
        host: a3c22a3a170184e2c834b0e1f4eb3b29-711704928.eu-central-1.elb.amazonaws.com
        port: 9094
    internalListeners:
      internal:
        host: kafka-headless.kafka.svc.cluster.local
        port: 29092
```

pkg/k8sutil/status.go
```diff
@@ -254,3 +256,62 @@ func UpdateRollingUpgradeState(c client.Client, cluster *v1beta1.KafkaCluster, t
...
+		}
+	}
+	// update loses the typeMeta of the config that's used later when setting ownerrefs
+	cluster.TypeMeta = typeMeta
+	logger.Info("updated listener statuses")
+	return nil
+}
+
+func createInternalListenerStatuses(kafkaCluster *v1beta1.KafkaCluster) map[string]v1beta1.ListenerStatus {
+	intListenerStatuses := make(map[string]v1beta1.ListenerStatus)
```
stoader, line 306:
```suggestion
	intListenerStatuses := make(map[string]v1beta1.ListenerStatus,len(kafkaCluster.Spec.ListenersConfig.InternalListeners))
```

Even though this may pre-allocate a map that is larger by one IHMO it's still worth it.

pkg/k8sutil/status.go
```diff
@@ -254,3 +256,62 @@ func UpdateRollingUpgradeState(c client.Client, cluster *v1beta1.KafkaCluster, t
 	logger.Info("Rolling upgrade status updated", "status", timeStamp)
 	return nil
 }
+
+func UpdateListenerStatuses(c client.Client, cluster *v1beta1.KafkaCluster, logger logr.Logger, extListenerStatuses map[string]v1beta1.ListenerStatus) error {
+	typeMeta := cluster.TypeMeta
+
+	intListenerStatuses := createInternalListenerStatuses(cluster)
```
stoader, line 263: Inject `intListenerStatuses` as a parameter instead of computing inside this method.

pkg/resources/kafka/kafka.go
```diff
@@ -804,3 +798,25 @@ func GetBrokersWithPendingOrRunningCCTask(kafkaCluster *v1beta1.KafkaCluster) []
 func isDesiredStorageValueInvalid(desired, current *corev1.PersistentVolumeClaim) bool {
 	return desired.Spec.Resources.Requests.Storage().Value() < current.Spec.Resources.Requests.Storage().Value()
 }
+
+func (r *Reconciler) createExternalListenerStatuses() (map[string]v1beta1.ListenerStatus, error) {
+	extListenerStatuses := make(map[string]v1beta1.ListenerStatus, len(r.KafkaCluster.Spec.ListenersConfig.ExternalListeners))
+	for _, eListener := range r.KafkaCluster.Spec.ListenersConfig.ExternalListeners {
```
stoader, line 804: Doesn't cover the case when `NodePort` is used to expose the Kafka cluster. For examples see:
https://github.com/banzaicloud/kafka-operator/blob/master/config/samples/simplekafkacluster-with-nodeport-external.yaml#L36

https://github.com/banzaicloud/kafka-operator/blob/master/config/samples/simplekafkacluster-with-nodeport-external.yaml#L55

pkg/resources/kafka/kafka.go
```diff
@@ -804,3 +798,25 @@ func GetBrokersWithPendingOrRunningCCTask(kafkaCluster *v1beta1.KafkaCluster) []
...
+			lbIP, err := getLoadBalancerIP(r.Client, r.KafkaCluster.GetNamespace(),
+				r.KafkaCluster.Spec.GetIngressController(), r.KafkaCluster.GetName(), eListener.Name)
+			if err != nil {
+				return nil, err
+			}
+			host = lbIP
+		}
+		extListenerStatuses[eListener.Name] = v1beta1.ListenerStatus{
+			Host: host,
+			Port: eListener.ContainerPort,
```
stoader, line 818: `ContainerPort` should be used only with `NodePort` when specif external IP is provided.

stoader (changes requested):

pkg/k8sutil/status.go
adamantal: Agreed. That was my first thought as well.

pkg/k8sutil/status.go
adamantal: Fixed

stoader (dismissed):

pkg/k8sutil/status.go
```diff
@@ -254,3 +256,92 @@ func UpdateRollingUpgradeState(c client.Client, cluster *v1beta1.KafkaCluster, t
...
+
+	err := c.Status().Update(context.Background(), cluster)
+	if apierrors.IsNotFound(err) {
+		err = c.Update(context.Background(), cluster)
+	}
+	if err != nil {
+		if !apierrors.IsConflict(err) {
+			return errors.WrapIf(err, "could not update listener status")
+		}
+		err := c.Get(context.TODO(), types.NamespacedName{
```
stoader, line 276:
```
  err := c.Get(context.TODO(), types.NamespacedName{
```

Rather create a 
```
ctx := context.Background()
```
at the beginning of the methods and use `ctx` for all the calls that requires a context. Also, the context could be injected as arg by the caller.

stoader (changes requested):

adamantal:
Fixed all review comments, tested on a remote cluster using envoy ingress controller with headless service turned on and off.
Will test istio ingress downstream.

baluchicken (approved):

pkg/kafkaclient/config_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright © 2019 Banzai Cloud
+// Copyright © 2020 Banzai Cloud
```
schultheiszn, line 1: don't change year values in license! with this change you would void any legal claims for the year 2019

pkg/k8sutil/status.go
```diff
@@ -254,3 +256,92 @@ func UpdateRollingUpgradeState(c client.Client, cluster *v1beta1.KafkaCluster, t
...
+	internalAddress := clientutil.GenerateKafkaAddressWithoutPort(kafkaCluster)
+	for _, iListener := range kafkaCluster.Spec.ListenersConfig.InternalListeners {
+		listenerStatusList := v1beta1.ListenerStatusList{}
+
+		// add headless or any broker address
+		var name string
+		if kafkaCluster.Spec.HeadlessServiceEnabled {
+			name = "headless"
+		} else {
+			name = "any-broker"
```
schultheiszn, line 316: you can use `any-broker` as a default values instead of `else`

baluchicken (approved):

adamantal:
Thanks for the reviews @baluchicken @schultheiszn @stoader 
Let me know if there's anything else we should discuss.

schultheiszn (approved):

stoader (approved):

PR #536, opened 2021-01-12
feat: Add HostnamePrefix in broker configuration

alungu:
This would allow to customize the format of the HostnameOverride

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | Related to https://github.com/banzaicloud/kafka-operator/pull/519

### What's in this PR?
In https://github.com/banzaicloud/kafka-operator/pull/519 , the `hostnameOverride` parameter behavior was changed. Now, `hostnameOverride` is just a suffix (and no longer the full fqdn).
This PR adds more flexibility to the `hostnameOverride`, adding the option to set a custom prefix in the broker config.

### Why?
Add custom prefix for broker external name, configurable at broker (or group) level.

### Checklist

- [x] Implementation tested

### DIFF
--- a/charts/kafka-operator/templates/crds.yaml
+++ b/charts/kafka-operator/templates/crds.yaml
@@ -705,6 +705,11 @@ spec:
                     type: object
                   config:
                     type: string
+                  hostnamePrefix:
+                    description: Hostname prefix it is used together with HostnameOverride
+                      to advertise a specific format for the broker address. The default
+                      hostname prefix format is <kafka-cluster-name>-<broker-id>-<listener-name>.<namespace>
+                    type: string
                   image:
                     type: string
                   imagePullSecrets:
@@ -1884,6 +1889,11 @@ spec:
                           prometheus.io/port: "9020"'
                         type: object
                       config:
+                        type: string
+                      hostnamePrefix:
+                        description: Hostname prefix it is used together with HostnameOverride
+                          to advertise a specific format for the broker address. The
+                          default hostname prefix format is <kafka-cluster-name>-<broker-id>-<listener-name>.<namespace>
                         type: string
                       image:
                         type: string
@@ -5590,7 +5600,9 @@ spec:
                           instead of public IP). In case of external listeners using
                           NodePort access method the broker instead of node public
                           IP (see "brokerConfig.nodePortExternalIP") is advertised
-                          on the address having the following format: <kafka-cluster-name>-<broker-id>.<namespace><value-specified-in-hostnameOverride-field>'
+                          on the address having the following format: <value-specified-in-hostnamePrefix-field><value-specified-in-hostnameOverride-field>
+                          If hostnamePrefix is missing (or empty) the advertised address
+                          will have the following format: <kafka-cluster-name>-<broker-id>-<listener-name>.<namespace><value-specified-in-hostnameOverride-field>'
                         type: string
                       name:
                         type: string
--- a/config/base/crds/kafka.banzaicloud.io_kafkaclusters.yaml
+++ b/config/base/crds/kafka.banzaicloud.io_kafkaclusters.yaml
@@ -706,6 +706,11 @@ spec:
                     type: object
                   config:
                     type: string
+                  hostnamePrefix:
+                    description: Hostname prefix it is used together with HostnameOverride
+                      to advertise a specific format for the broker address. The default
+                      hostname prefix format is <kafka-cluster-name>-<broker-id>-<listener-name>.<namespace>
+                    type: string
                   image:
                     type: string
                   imagePullSecrets:
@@ -1885,6 +1890,11 @@ spec:
                           prometheus.io/port: "9020"'
                         type: object
                       config:
+                        type: string
+                      hostnamePrefix:
+                        description: Hostname prefix it is used together with HostnameOverride
+                          to advertise a specific format for the broker address. The
+                          default hostname prefix format is <kafka-cluster-name>-<broker-id>-<listener-name>.<namespace>
                         type: string
                       image:
                         type: string
@@ -5591,7 +5601,9 @@ spec:
                           instead of public IP). In case of external listeners using
                           NodePort access method the broker instead of node public
                           IP (see "brokerConfig.nodePortExternalIP") is advertised
-                          on the address having the following format: <kafka-cluster-name>-<broker-id>.<namespace><value-specified-in-hostnameOverride-field>'
+                          on the address having the following format: <value-specified-in-hostnamePrefix-field><value-specified-in-hostnameOverride-field>
+                          If hostnamePrefix is missing (or empty) the advertised address
+                          will have the following format: <kafka-cluster-name>-<broker-id>-<listener-name>.<namespace><value-specified-in-hostnameOverride-field>'
                         type: string
                       name:
                         type: string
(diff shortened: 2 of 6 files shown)


adamantal:
Thanks for submitting the pull request @alungu.

I checked #519 with regards to this issue and at first glance it seems that the behaviour has been preserved. The code piece responsible for assembling the hostnames has been moved along, but essentially is the same.

For reference:
Current implementation (master) in `kafka.go:795` (in `createExternalListenerStatuses()`):
```go
		if eListener.HostnameOverride != "" {
			host = eListener.HostnameOverride
		} else if eListener.GetAccessMethod() == corev1.ServiceTypeLoadBalancer {
			foundLBService, err = getServiceFromExternalListener(r.Client, r.KafkaCluster, eListener.Name)
			if err != nil {
				return nil, ...
			}
			lbIP, err := getLoadBalancerIP(foundLBService)
			if err != nil {
				return nil, ...
			}
			host = lbIP
		}
```
Before the patch (9b48a9f858ff) the hostnames are created in `kafka.go:197`:
```go
		if eListener.HostnameOverride != "" {
			lbIPs[eListener.Name] = eListener.HostnameOverride
		} else if eListener.GetAccessMethod() == corev1.ServiceTypeLoadBalancer {
			lbIP, err := getLoadBalancerIP(r.Client, r.KafkaCluster.GetNamespace(),
				r.KafkaCluster.Spec.GetIngressController(), r.KafkaCluster.GetName(), eListener.Name)
			if err != nil {
				return err
			}
			lbIPs[eListener.Name] = lbIP
		}
```


Could you share a bit more what configuration did you use, when you encountered a different behaviour?

On a side note I think the feature you suggested would make sense, but I'd suggest a more restrictive approach. Something like use the the `hostnameOverride` as suffix or as a fqdn based on that it ends with a dot or not. What do you think about this suggestion?

alungu:
Hi @adamantal ! Thank you for checking the patch.
I am talking about https://github.com/banzaicloud/kafka-operator/blob/master/pkg/resources/kafka/kafka.go#L854 
The `hostnameOverride` was used in the external listener as it was declared. Now, it is a suffix, and several information is prepended (broker Id, namespace, external listener name, cluster name).

Our use case is the following: We are using NodePorts to access the brokers externally, but we don't want to use one NodePort / broker. Instead, we want to use a NodePort per Availability Zone (rack / DC). We are OK to manage the NodePorts on our own (there is not too much to manage since the NodePorts are preserved during scaling operations if there is a NodePort service per AZ). But we still need the help of the operator to set the proper externalListener address for each broker. As such, we need to have to option to customize the hostname at broker level or config group level.

This is a first PR from a set:
1. We need to be able to set override the hostname at broker / config group level
2. We need a new ExternalListener AccessMode, because we want to use self-managed NodePorts with operator managed Ingress
3. We would like to have to option to split the ingress deployment per Availability Zone, to ensure that the ingress route will never cross the AZ bounds.

The full picture is here: https://github.com/banzaicloud/kafka-operator/issues/487

> use the the hostnameOverride as suffix or as a fqdn based on that it ends with a dot or not
As I described above, we would like to be able to set the hostnameOverrride at broker level or config group level.
`hostnameOverride` is a global (cluster level) parameter.

baluchicken:
Hi @alungu, we are actively working on the patch which solves your first and your third problem. And with a slight modification we can deal with the number two as well.
The patch will introduce some changes in the Kafka cluster CRD.
We want to introduce a new field inside the BrokerConfig struct called `BrokerIdBindings` which allows setting specific loadbalancer to a specific broker mappings.
```yaml
// BrokerConfig defines the broker configuration
type BrokerConfig struct {
       	Image              string                        `json:"image,omitempty"`
	Config             string                        `json:"config,omitempty"`
        ...
	// BrokerIdBindings allows to set specific loadbalancer to a specific broker mappings.
	// If left empty, all broker will inherit this loadbalancer
	BrokerIdBindings []string `json:"brokerIdBindings,omitempty"`
}
```
```yaml
type LoadBalancerSettings struct {
	// In case of external listeners using LoadBalancer access method the value of this field is used to advertise the
	// Kafka broker external listener instead of the public IP of the provisioned LoadBalancer service (e.g. can be used to
	// advertise the listener using a URL recorded in DNS instead of public IP).
	// In case of external listeners using NodePort access method the broker instead of node public IP (see "brokerConfig.nodePortExternalIP")
	// is advertised on the address having the following format: <kafka-cluster-name>-<broker-id>.<namespace><value-specified-in-hostnameOverride-field>
	HostnameOverride string `json:"hostnameOverride,omitempty"`
	// ServiceAnnotations defines annotations which will
	// be placed to the service or services created for the external listener
	ServiceAnnotations map[string]string `json:"serviceAnnotations,omitempty"`
	// externalTrafficPolicy denotes if this Service desires to route external
	// traffic to node-local or cluster-wide endpoints. "Local" preserves the
	// client source IP and avoids a second hop for LoadBalancer and Nodeport
	// type services, but risks potentially imbalanced traffic spreading.
	// "Cluster" obscures the client source IP and may cause a second hop to
	// another node, but should have good overall load-spreading.
	// +optional
	ExternalTrafficPolicy corev1.ServiceExternalTrafficPolicyType `json:"externalTrafficPolicy,omitempty"`
}

// ExternalListenerConfig defines the external listener config for Kafka
type ExternalListenerConfig struct {
	CommonListenerSpec   `json:",inline"`
	LoadBalancerSettings `json:",inline"`
	// configuring AnyCastPort allows kafka cluster access without specifying the exact broker
	AnyCastPort *int32 `json:"anyCastPort,omitempty"`
	// +kubebuilder:validation:Enum=LoadBalancer;NodePort
	// accessMethod defines the method which the external listener is exposed through.
	// Two types are supported LoadBalancer and NodePort.
	// The recommended and default is the LoadBalancer.
	// NodePort should be used in Kubernetes environments with no support for provisioning Load Balancers.
	// +optional
	AccessMethod corev1.ServiceType `json:"accessMethod,omitempty"`
	// configuring Config allows to specify ingress controller configuration per external listener
	// if not set the default IstioIngressConfig or EnvoyConfig will be used from the KafkaClusterSpec
	Config *Config `json:"config,omitempty"`
}
// Config defines the external access ingress controller configuration
type Config struct {
	IstioIngressConfig map[string]IngressConfig `json:"ingressConfig,omitempty"`
}
type IngressConfig struct {
	LoadBalancerSettings `json:",inline"`
	IstioIngressConfig *IstioIngressConfig `json:"istioIngressConfig,omitempty"`
	EnvoyConfig        *EnvoyConfig        `json:"envoyConfig,omitempty"`
}
```
We are introducing a new field called `Config` where external listeners can be uniquely configured. Also this field allows users to add multiple ingress controller for one external listener. With this you can achieve:
`We would like to have to option to split the ingress deployment per Availability Zone, to ensure that the ingress route will never cross the AZ bounds.` 
By moving the LoadBalancerSettings inside the `IngressConfig` struct we can achieve:
`We need to be able to set override the hostname at broker / config group level`

About your second requirement:
`We need a new ExternalListener AccessMode, because we want to use self-managed NodePorts with operator managed Ingress`
We can allow to change service type inside the `LoadBalancerSettings` struct so instead of creating a loadbalancer type service we can do a nodeport one backed with an ingress controller.

>Note: Naming are in early stage:)
We know these are expert use cases so we are going to still support the basic settings by just omitting the Config struct from the externalListener configuration.

How all these sounds for you?

alungu:
Hi @baluchicken ,

I like your proposal. It's clear and it would bring even more flexibility than our implementation.

I do have one question regarding the `IngressConfig` behavior:
- Today, `BrokerConfig` can be set either at broker level, or at broker config group level - e.g: define a set of configurations for all the brokers sharing some properties, like the DataCenter / AvailabilityZone.
- In the new implementation, if I would define an `IngressConfig` at **broker group** level, would it result in a single (Envoy and/or NodePort) deployment for the entire group? My concern is that we might end up with 1 ingress deployment per broker, which would add extra complexity and (probably) extra costs.

Also, is there a tentative release date for these features?

Thanks!

PR #542, opened 2021-01-28
refactor: use properties lib for handlig configs

chrisgacsal:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Improve the way of handling Java Properties configuration by using `github.com/banzaicloud/kafka-operator/properties` library.

### Why?
The `github.com/banzaicloud/kafka-operator/properties` is a more sophisticated solution for handling Java Properties.

- [x] Implementation tested

### To Do
- [x] Update PR after #540 got merged
- [x] Update PR after #543 got merged
- [x] Implement `Sort` and `Diff` methods/functions in `github.com/banzaicloud/kafka-operator/properties`


go.mod
```diff
@@ -3,13 +3,14 @@ module github.com/banzaicloud/kafka-operator
...
 require (
-	emperror.dev/errors v0.7.0
+	emperror.dev/errors v0.8.0
 	github.com/Shopify/sarama v1.27.1
 	github.com/banzaicloud/bank-vaults/pkg/sdk v0.3.1
 	github.com/banzaicloud/istio-client-go v0.0.9
 	github.com/banzaicloud/istio-operator v0.0.0-20200407070503-3f7dc6953a7b
 	github.com/banzaicloud/k8s-objectmatcher v1.4.1
 	github.com/banzaicloud/kafka-operator/api v0.0.0
+	github.com/banzaicloud/kafka-operator/properties v0.0.0-20210128180953-6e7201a76903
```
baluchicken, line 13: I think we should handle this as above. It means we should use api v0.0.0 and use replace which points to the dir where the properties lives.
WIth that approach we can update the properties dir without any go.mod modification.

chrisgacsal: Agreed. Will update the PR as soon as the #543 got merged.

hi-im-aren (dismissed):

pkg/resources/cruisecontrol/configmap.go
```diff
@@ -28,46 +28,81 @@ import (
...
+	// Add base Cruise Control configuration
+	conf, err := properties.NewFromString(r.KafkaCluster.Spec.CruiseControlConfig.Config)
+	if err != nil {
+		log.Error(err, "parsing Cruise Control configuration failed", "config", r.KafkaCluster.Spec.CruiseControlConfig.Config)
+	}
+	ccConfig.Merge(conf)
+
+	// Set bootstrap servers
+	bootstrapServers := fmt.Sprintf("%s:%d",
+		generateBootstrapServer(r.KafkaCluster.Spec.HeadlessServiceEnabled, r.KafkaCluster.Name),
```
adamantal, line 48: [this is not related to the patch]

Here we manually generate the server address from the available information. 

@baluchicken I think we should remove these places from the code base where we manually assemble the server address: we should use the listener status field wherever possible - or at least from the components that depend on Kafka. By the time the reconcile gets there, the listener statuses should have been populated and we can use that field in CC.

pkg/util/util.go
```diff
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"emperror.dev/errors"
+	samara "github.com/Shopify/sarama"
```
adamantal, line 26: There's a typo here
```suggestion
	"github.com/Shopify/sarama"
```
It `sarama`, not `samara`.

pkg/util/util.go
```diff
@@ -320,3 +309,17 @@ func CreateLogger(debug bool, development bool) logr.Logger {
 
 	return k8s_zap.New(k8s_zap.UseDevMode(development), k8s_zap.Encoder(encoder), k8s_zap.Level(level))
 }
+
+// ConvertConfigEntryListToProperties function takes []samara.ConfigEntry and coverts it to Properties object
+func ConvertConfigEntryListToProperties(config []samara.ConfigEntry) (*properties.Properties, error) {
```
adamantal, line 314:
```suggestion
// ConvertConfigEntryListToProperties function takes []sarama.ConfigEntry and coverts it to Properties object
func ConvertConfigEntryListToProperties(config []sarama.ConfigEntry) (*properties.Properties, error) {
```

pkg/resources/kafka/configs.go
```diff
@@ -41,26 +44,30 @@ func (r *Reconciler) reconcilePerBrokerDynamicConfig(brokerId int32, brokerConfi
...
 	}
 
-	// query the current configs
-	brokerConfigKeys := make([]string, 0, len(fullPerBrokerConfig))
-	for key := range fullPerBrokerConfig {
-		brokerConfigKeys = append(brokerConfigKeys, key)
-	}
+	// query the current config
+	brokerConfigKeys := fullPerBrokerConfig.Keys()
+	sort.Strings(brokerConfigKeys)
```
adamantal, line 70: Why is the sort necessary?

adamantal (commented):
Thanks for the work @chrisgacsal, the properties lib is very powerful and increases the quality of the codebase by a lot!

A generic comment:
I see a lot of `config.Set(key, value, "")`. Though it's not a big effort, to put that `, ""` as a third parameter every single time, I think it may worth to have these functions as setting 
the config without comment is much frequently used than the other:
```golang
func Set(key string, value interface{})
func SetWithComment(key string, value interface{}, comment string)
```

chrisgacsal:
> Thanks for the work @chrisgacsal, the properties lib is very powerful and increases the quality of the codebase by a lot!
> 
> A generic comment:
> I see a lot of `config.Set(key, value, "")`. Though it's not a big effort, to put that `, ""` as a third parameter every single time, I think it may worth to have these functions as setting
> the config without comment is much frequently used than the other:
> 
> ```go
> func Set(key string, value interface{})
> func SetWithComment(key string, value interface{}, comment string)
> ```

Agreed. I've been thinking about this as well. Will add `SetWithComment` to the properties lib.

pkg/util/util.go
chrisgacsal: Uhh. Nice catch. Will fix.

pkg/resources/kafka/configs.go
chrisgacsal: Nope. You are right. Sorting is not needed. Will remove it.

pkg/resources/cruisecontrol/configmap.go
baluchicken: I agree we should do this in a separate PR:)

pkg/k8sutil/resource.go
```diff
@@ -133,8 +134,20 @@ func Reconcile(log logr.Logger, client runtimeClient.Client, desired runtime.Obj
 			case *corev1.ConfigMap:
 				// Only update status when configmap belongs to broker
 				if id, ok := desired.(*corev1.ConfigMap).Labels["brokerId"]; ok {
-					currentConfigs := util.ParsePropertiesFormat(current.(*corev1.ConfigMap).Data[kafka.ConfigPropertyName])
-					desiredConfigs := util.ParsePropertiesFormat(desired.(*corev1.ConfigMap).Data[kafka.ConfigPropertyName])
+					currentConfigs, err := properties.NewFromString(current.(*corev1.ConfigMap).Data[kafka.ConfigPropertyName])
+					if err != nil {
+						return errors.WrapIf(err, fmt.Sprintf("could not parse the current configuration for broker with id: %s", id))
```
stoader, line 139: Instead of `errors.WrapIf(err, fmt.Sprintf())`  `errors.WrapIff()` can be used. Thinking it more the use structured error would be even better `errors.WrapWithDetails(err, "could not parse the current configuration for broker", "borkerId", id)`

pkg/resources/kafka/configmap.go
```diff
@@ -33,74 +29,105 @@ import (
...
+		}
+	}
+
+	// Add listener configuration
+	advertisedListenerConf := generateAdvertisedListenerConfig(id, r.KafkaCluster.Spec.ListenersConfig, extListenerStatuses, intListenerStatuses, controllerIntListenerStatuses)
+	if len(advertisedListenerConf) > 0 {
+		if err := config.Set("advertised.listeners", advertisedListenerConf, ""); err != nil {
+			log.Error(err, "setting advertised.listeners in broker configuration resulted an error")
+		}
+	}
```
baluchicken, line 121: Nit: I would move this up just after line 40. In that case the listener specific configs can be together:)

pkg/k8sutil/resource.go
chrisgacsal: Agreed. Will fix. Ty.

pkg/resources/kafka/configmap.go
chrisgacsal: Agreed. Will fix. Ty.

baluchicken (approved):

chrisgacsal:
@hi-im-aren @adamantal @stoader @baluchicken Thank you guys for reviewing the PR. I have incorporated your suggestions. Please take another look on it. Many thanks!

adamantal (approved):
LGTM, thanks for making the changes.

stoader (approved):

PR #547, opened 2021-02-02
refactor: get bootstrap brokers

chrisgacsal:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Refactor the way how the list of Kafka bootstrap servers is compiled and retrieved.

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/resources/cruisecontrol/configmap.go
+++ b/pkg/resources/cruisecontrol/configmap.go
@@ -43,11 +43,10 @@ func (r *Reconciler) configMap(clientPass, capacityConfig string, log logr.Logge
 	}
 	ccConfig.Merge(conf)
 
-	// Set bootstrap servers
-	bootstrapServers := fmt.Sprintf("%s:%d",
-		generateBootstrapServer(r.KafkaCluster.Spec.HeadlessServiceEnabled, r.KafkaCluster.Name),
-		generateBootstrapServerPort(log, r.KafkaCluster.Spec.ListenersConfig.InternalListeners),
-	)
+	bootstrapServers, err := kafkautils.GetBootstrapServersService(r.KafkaCluster)
+	if err != nil {
+		log.Error(err, "getting Kafka bootstrap servers for Cruise Control failed")
+	}
 	if err = ccConfig.Set("bootstrap.servers", bootstrapServers); err != nil {
 		log.Error(err, "setting bootstrap.servers in Cruise Control configuration failed", "config", bootstrapServers)
 	}
@@ -103,24 +102,6 @@ func generateSSLConfig(l *v1beta1.ListenersConfig, clientPass string, log logr.L
 		}
 	}
 	return sslConf
-}
-
-func generateBootstrapServer(headlessEnabled bool, clusterName string) string {
-	if headlessEnabled {
-		return fmt.Sprintf(kafkautils.HeadlessServiceTemplate, clusterName)
-	}
-	return fmt.Sprintf(kafkautils.AllBrokerServiceTemplate, clusterName)
-}
-
-func generateBootstrapServerPort(log logr.Logger, listenerConfig []v1beta1.InternalListenerConfig) int32 {
-	for _, listener := range listenerConfig {
-		if listener.UsedForInnerBrokerCommunication {
-			return listener.ContainerPort
-		}
-	}
-	// This should never happen since there should be least one listener for inner broker communication
-	log.V(1).Info("listener for inner communication not found, falling back to first listener")
-	return listenerConfig[0].ContainerPort
 }
 
 const (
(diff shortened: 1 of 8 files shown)


pkg/resources/cruisecontrol/configmap.go
```diff
@@ -43,11 +44,10 @@ func (r *Reconciler) configMap(clientPass, capacityConfig string, log logr.Logge
...
 
-	// Set bootstrap servers
-	bootstrapServers := fmt.Sprintf("%s:%d",
-		generateBootstrapServer(r.KafkaCluster.Spec.HeadlessServiceEnabled, r.KafkaCluster.Name),
-		generateBootstrapServerPort(log, r.KafkaCluster.Spec.ListenersConfig.InternalListeners),
-	)
+	bootstrapServers := kafkautils.GetBootstrapServers(r.KafkaCluster)
+	if bootstrapServers == "" {
+		log.Error(errors.New("getting bootstrap servers returned an empty string"), "")
+	}
```
hi-im-aren, line 50: I would handle this in the GetBootstrapServers() func itself.

pkg/util/kafka/common.go
```diff
@@ -166,3 +171,41 @@ func newListenerSecurityProtocolMap(s string) listenerSecurityProtocolMap {
...
+		listenerServiceName = AnyBrokerListenerServiceName
+	}
+
+	var bootstrapServersList []string
+
+	if len(listenerStatusList) > 0 {
+		for _, ls := range listenerStatusList {
+			if ls.Name == listenerServiceName {
+				bootstrapServers = ls.Address
+				return
```
hi-im-aren, line 204: Not sure about using naked return in this func, its rather long for that and just creates confusion in general.

pkg/util/kafka/common.go
```diff
@@ -166,3 +171,41 @@ func newListenerSecurityProtocolMap(s string) listenerSecurityProtocolMap {
...
+	if !ok {
+		return
+	}
+
+	switch {
+	case cr.Spec.HeadlessServiceEnabled:
+		listenerServiceName = HeadlessListenerServiceName
+	default:
+		listenerServiceName = AnyBrokerListenerServiceName
+	}
```
hi-im-aren, line 196: I think an If would be more appropriate here instead of a switch case with 1 case

pkg/resources/cruisecontrol/configmap.go
chrisgacsal: Alright. Will change the `GetBootstrapServer()` to return an error in case it was not able to determine the Kafka bootstrap server from the provided CR.

pkg/util/kafka/common.go
chrisgacsal: Agreed. Will refactor.

pkg/util/kafka/common.go
chrisgacsal: Agreed. Will refactor.

hi-im-aren (approved):

pkg/resources/kafka/configmap.go
```diff
@@ -99,7 +99,10 @@ func (r *Reconciler) getConfigProperties(bConfig *v1beta1.BrokerConfig, id int32
 	if err := config.Set("metric.reporters", "com.linkedin.kafka.cruisecontrol.metricsreporter.CruiseControlMetricsReporter"); err != nil {
 		log.Error(err, "setting metric.reporters in broker configuration resulted an error")
 	}
-	bootstrapServers := getInternalListener(r.KafkaCluster.Spec.ListenersConfig.InternalListeners, id, r.KafkaCluster.Spec.GetKubernetesClusterDomain(), r.KafkaCluster.Namespace, r.KafkaCluster.Name, r.KafkaCluster.Spec.HeadlessServiceEnabled)
+	bootstrapServers, err := kafkautils.GetBootstrapServersService(r.KafkaCluster)
```
stoader, line 102: Wouldn't be more efficient if `cruise.control.metrics.reporter.bootstrap.servers` is pointed at the local broker (the broker the reporter runs in) instead of `kafka-all-brokers`?

Also, should the `cruise.control.metrics.reporter` run in each broker or only in the controller broker?

stoader (commented):
Wouldn't be more efficient if `cruise.control.metrics.reporter.bootstrap.servers` is pointed at the local broker (the broker the reporter runs in) instead of `kafka-all-brokers`?

Also, should the `cruise.control.metrics.reporter` run in each broker or only in the controller broker?

chrisgacsal:
> Wouldn't be more efficient if `cruise.control.metrics.reporter.bootstrap.servers` is pointed at the local broker (the broker the reporter runs in) instead of `kafka-all-brokers`?

I do not think it would make much difference as the `bootstrap.servers` is used by the Kafka client only for discovery purposes. I have no preference though, so either way is fine by me.

> Also, should the `cruise.control.metrics.reporter` run in each broker or only in the controller broker?

So each broker reports their own metrics only which means that we need `cruise.control.metrics.reporter` to be invoked as part of all the brokers in the cluster in order to make metrics for the whole cluster available for Cruise Control.

@stoader Thanks for the review. Please find my answers inline.

baluchicken (approved):

stoader (approved):

PR #549, opened 2021-02-05
Adding InitContainers, Volumes and VolumeMounts to the KafkaCluster's…

adamantal:
… BrokerConfig field

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

Exposing `InitContainer`, `Volumes` and `VolumeMount` fields of the broker pod. It can be configured on a brokerConfigGroup bases.

### Checklist

- [x] Implementation tested


adamantal:
When tested on a local cluster `make deploy` triggers https://github.com/kubernetes-sigs/kubebuilder/issues/1140.
Working on a workaround

adamantal:
Core changes:
- Added support for `InitContainers`, `Volumes` and `VolumeMount` fields in the `KafkaCluster` CR

Additional changes:
- Since the `KafkaCluster` CRD has grown too big, it can not be used in `kubectl apply` commands anymore. 
  - Reason: Applying the CRD would create a `last-applied-annotation` that is filled with the content of the file which is too big. When this happens, the `kubectl apply` command fails with the following error: `metadata.annotations: Too long: must have at most 262144 characters`
  - Mitigation: use `create` and `replace` in the `make` commands instead of `apply`
- The `make manifests` command now refreshes the CRDs in the helm chart `templates` folder.
  - Note: it is possible to generate the CRs directly by the `controller-gen`, but it turned out that specifying `output:crd:stdout` option and using pipe redirection of the stdout to `crds.yaml` gives nondeterministic result. 
- Removed references to the `KafkaCluster` and `KafkaUser` CRs in `kustomization.yaml` s that are processed during  `kustomize build` commands (as they are used in conjunction with `kubectl apply` in the `Makefile`). They are created separately during the `make install` command. Kept `KafkaTopic` CRs in the `kustomization.yaml` s as `config/base/webhook/webhook_in_kafkatopics.yaml` changes that definition, and that CR is not too big for the `last-applied-configutation` annotation.

stoader (approved):

baluchicken (approved):

chrisgacsal (approved):

pkg/resources/kafka/pod.go
```diff
@@ -75,9 +75,9 @@ func (r *Reconciler) pod(id int32, brokerConfig *v1beta1.BrokerConfig, pvcs []co
 
 	dataVolume, dataVolumeMount := generateDataVolumeAndVolumeMount(pvcs)
 
-	var volume []corev1.Volume
-	var volumeMount []corev1.VolumeMount
-	var initContainers []corev1.Container
+	volume := brokerConfig.Volumes
+	volumeMount := brokerConfig.VolumeMounts
+	initContainers := brokerConfig.InitContainers
```
stoader, line 80: To be on the safe side it worth sorting the content of `volume`, `volumeMount` and `initContainers` to ensure that the broker pod is not restarted unnecessary between two reconciles. E.g. user changes the ordering of volumes in `brokerConfig.Volumes` that should cause an update to the pod. If object matcher can handle this already than no need for the sorting.

stoader (approved):

baluchicken (approved):

PR #553, opened 2021-02-09
Add operator.annotations also to Deployment object

kirecek:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | n/a

### What's in this PR?
Add `.Values.operator.annotations` to Deployment object. Not just pods

### Why?
To allow other k8s  tooling manipulate with the deployment.

Simple usacase: kube-janitor cleans ups development deployments after specified ttl in annotations. By not providing this annotation users cannot have stable kafkaclusters.


kirecek:
It would be nice to allow annotate also other operator objects. I'd like to add something like: `operator.service.annotations` etc...

baluchicken (approved):

stoader (approved):

PR #562, opened 2021-02-18
Update helm chart to prepare v0.15.0 release

baluchicken:
Update helm chart to prepare new release v0.15.0

### DIFF
--- a/charts/kafka-operator/values.yaml
+++ b/charts/kafka-operator/values.yaml
@@ -12,7 +12,7 @@ operator:
   annotations: {}
   image:
     repository: ghcr.io/banzaicloud/kafka-operator
-    tag: v0.14.0
+    tag: v0.15.0
     pullPolicy: IfNotPresent
   vaultAddress: ""
   # vaultSecret containing a `ca.crt` key with the Vault CA Certificate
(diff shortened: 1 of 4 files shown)


charts/kafka-operator/values.yaml
```diff
@@ -12,7 +12,7 @@ operator:
   annotations: {}
   image:
     repository: ghcr.io/banzaicloud/kafka-operator
-    tag: v0.14.0
+    tag: v0.15.0
```
pbalogh-sa, line 15: Due to we defined the appVersion in `Charts.yaml` it can be used in templates as well.
Instead of using image: `"{{ .Values.operator.image.repository }}:{{ .Values.operator.image.tag }}"` in this [line](https://github.com/banzaicloud/kafka-operator/blob/361c8cd2a0991f0e94246acca4827414a4399889/charts/kafka-operator/templates/operator-deployment-with-webhook.yaml#L159) we would use `image: "{{ .Values.operator.image.repository }}:{{ .Values.operator.image.tag | default .Chart.AppVersion }}"` and in the values.yaml we would set tag to empty string. `tag: ""`

stoader (approved):

hi-im-aren (approved):

chrisgacsal (approved):

PR #564, opened 2021-02-24
Fix ingress controller behaviour in case of downscale

baluchicken:
| Q               | A
| --------------- | ---
| Bug fix?        |yes|
| New feature?    | no|
| API breaks?     | no|
| Deprecations?   | no|

### What's in this PR?

This PR fixes the ingress controller behaviour. In case of downscale the ingress controller gets terminated before the broker which means that broker gets unavailable despite of it is available inside the cluster. With this fix the ingress controller gets reconfigured after the broker has been deleted.

- [x] Implementation tested

### DIFF
--- a/config/samples/simplekafkacluster-with-brokerbindings.yaml
+++ b/config/samples/simplekafkacluster-with-brokerbindings.yaml
@@ -6,7 +6,6 @@ metadata:
   name: kafka
 spec:
   headlessServiceEnabled: true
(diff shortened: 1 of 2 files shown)


stoader (approved):

schultheiszn (approved):

chrisgacsal (approved):
:shipit:

PR #592, opened 2021-04-15
K8s provided nodeport ports for nodeportaccess

baluchicken:
| Q               | A
| --------------- | ---
| Bug fix?        | yes|
| New feature?    | no|
| API breaks?     | no|
| Deprecations?   | no|

### What's in this PR?

This PR fixes the external access if the externalStartingPort is set to zero in case of a NodePort setup.
Nodeport services received nodeports from kubernetes but this did not propagated back to advertised.listener config, which made these nodeport services useless.

### Checklist

- [x] Implementation tested

### DIFF
--- a/config/samples/simplekafkacluster_with_k8s_provided_nodeport.yaml
+++ b/config/samples/simplekafkacluster_with_k8s_provided_nodeport.yaml
@@ -1,0 +1,257 @@
+apiVersion: kafka.banzaicloud.io/v1beta1
+kind: KafkaCluster
+metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: kafka
+spec:
+  headlessServiceEnabled: true
+  zkAddresses:
+    - "zookeeper-client.zookeeper:2181"
+  propagateLabels: false
+  oneBrokerPerNode: false
+  clusterImage: "ghcr.io/banzaicloud/kafka:2.13-2.7.0-bzc.1"
+  readOnlyConfig: |
+    auto.create.topics.enable=false
+    cruise.control.metrics.topic.auto.create=true
+    cruise.control.metrics.topic.num.partitions=1
+    cruise.control.metrics.topic.replication.factor=2
+  brokerConfigGroups:
+    default:
+      storageConfigs:
+        - mountPath: "/kafka-logs"
+          pvcSpec:
+            accessModes:
+              - ReadWriteOnce
+            resources:
+              requests:
+                storage: 10Gi
+      brokerAnnotations:
+        prometheus.io/scrape: "true"
+        prometheus.io/port: "9020"
+  brokers:
+    - id: 0
+      brokerConfigGroup: "default"
+    - id: 1
+      brokerConfigGroup: "default"
+    - id: 2
+      brokerConfigGroup: "default"
+  rollingUpgradeConfig:
+    failureThreshold: 1
+  listenersConfig:
+    internalListeners:
+      - type: "plaintext"
+        name: "internal"
+        containerPort: 29092
+        usedForInnerBrokerCommunication: true
+      - type: "plaintext"
+        name: "controller"
+        containerPort: 29093
+        usedForInnerBrokerCommunication: false
+        usedForControllerCommunication: true
+    externalListeners:
+      - type: "plaintext"
+        name: "external"
+# When externalStartingPort is set to 0 kubernetes will choose an open nodeport for the service.
+# This behaviour is only supported for NodePort type accessMethod.
+        externalStartingPort: 0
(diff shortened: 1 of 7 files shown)


config/samples/simplekafkacluster_with_k8s_provided_nodeport.yaml
```diff
@@ -0,0 +1,257 @@
...
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: kafka
+spec:
+  headlessServiceEnabled: true
+  zkAddresses:
+    - "zookeeper-client.zookeeper:2181"
+  propagateLabels: false
+  oneBrokerPerNode: false
+  clusterImage: "ghcr.io/banzaicloud/kafka:2.13-2.7.0-bzc.1"
```
stoader, line 13:
```suggestion
  clusterImage: "ghcr.io/banzaicloud/kafka:2.13-2.7.0-bzc.2"
```

pkg/resources/kafka/kafka.go
```diff
@@ -905,6 +911,30 @@ func (r *Reconciler) createExternalListenerStatuses(log logr.Logger) (map[string
 	return extListenerStatuses, nil
 }
 
+func (r *Reconciler) getK8sAssignedNodeport(log logr.Logger, eListenerName string, brokerId int32) (int32, error) {
+	log.Info("determining automatically assigned nodeport",
+		"brokerId", brokerId, "listenerName", eListenerName)
+	nodePortSvc := &corev1.Service{}
+	err := r.Client.Get(context.Background(),
+		types.NamespacedName{Name: fmt.Sprintf("%s-%d-%s", r.KafkaCluster.GetName(), brokerId, eListenerName),
```
stoader, line 919: can we factor out the nodeport service name generation logic into a separate function and use it here and also where the service is created?

pkg/resources/kafka/kafka.go
```diff
@@ -905,6 +911,30 @@ func (r *Reconciler) createExternalListenerStatuses(log logr.Logger) (map[string
...
+			Namespace: r.KafkaCluster.GetNamespace()}, nodePortSvc)
+	if err != nil {
+		return 0, errors.WrapWithDetails(err, "could not get nodeport service",
+			"brokerId", brokerId, "listenerName", eListenerName)
+	}
+	log.Info("nodeport service successfully found",
+		"brokerId", brokerId, "listenerName", eListenerName)
+
+	var nodePort int32
+	for _, port := range nodePortSvc.Spec.Ports {
```
stoader, line 929: Is the port number assigned to `port.NodePort` instantaneously by K8s or there might be some delay thus we need to wait here until the port is set in case `port.NodePort` is 0?

stoader (changes requested):

pkg/resources/kafka/kafka.go
```diff
@@ -905,6 +911,30 @@ func (r *Reconciler) createExternalListenerStatuses(log logr.Logger) (map[string
...
+		"brokerId", brokerId, "listenerName", eListenerName)
+	nodePortSvc := &corev1.Service{}
+	err := r.Client.Get(context.Background(),
+		types.NamespacedName{Name: fmt.Sprintf("%s-%d-%s", r.KafkaCluster.GetName(), brokerId, eListenerName),
+			Namespace: r.KafkaCluster.GetNamespace()}, nodePortSvc)
+	if err != nil {
+		return 0, errors.WrapWithDetails(err, "could not get nodeport service",
+			"brokerId", brokerId, "listenerName", eListenerName)
+	}
+	log.Info("nodeport service successfully found",
```
adamantal, line 925: Do we really need this at Info level? This way we can print `[#external listeners]x[#ingress config]x[#brokers]` log entries worst case, which can pretty high even in a mid size cluster.
```suggestion
	log.V(1).Info("nodeport service successfully found",
```

adamantal (approved):
LGTM, one thing.

config/samples/simplekafkacluster_with_k8s_provided_nodeport.yaml
baluchicken: I will do a separate PR for that where I will update all our samples with the new clusterImage version

hi-im-aren (approved):

pkg/resources/kafka/kafka.go
baluchicken: Accourding to the documentation it happens instantaneously. It is like assigning the clusterIP to the clusterIP types service. I check the internet regarding this error but I found nothing.

stoader (approved):

hi-im-aren (approved):

PR #594, opened 2021-04-19
Handle when nodeport is not allocated yet

baluchicken:
| Q               | A
| --------------- | ---
| Bug fix?        | yes|
| New feature?    | no|
| API breaks?     | no|
| Deprecations?   | no|

### What's in this PR?

This PR handles the case when nodeport is not allocated by the Kubernetes reconciler. We are going to wait one more cycle if that happens.

- [X] Implementation tested
- [X] Error handling code meets the [guideline](https://github.com/banzaicloud/developer-guide/blob/master/docs/coding-style/error-handling-guide.md)

### DIFF
--- a/pkg/resources/kafka/kafka.go
+++ b/pkg/resources/kafka/kafka.go
@@ -933,6 +933,10 @@ func (r *Reconciler) getK8sAssignedNodeport(log logr.Logger, eListenerName strin
 			break
 		}
 	}
+	if nodePort == 0 {
+		return nodePort, errorfactory.New(errorfactory.ResourceNotReady{},
+			errors.New("nodeport port number is not allocated"), "nodeport number is still 0")
+	}
 	return nodePort, nil
 }
 


pkg/resources/kafka/kafka.go
```diff
@@ -928,11 +928,15 @@ func (r *Reconciler) getK8sAssignedNodeport(log logr.Logger, eListenerName strin
 
 	var nodePort int32
 	for _, port := range nodePortSvc.Spec.Ports {
-		if port.Name == fmt.Sprintf("broker-%d", brokerId) {
+		if port.Name == fmt.Sprintf("broker-%d", brokerId) && port.NodePort != 0 {
 			nodePort = port.NodePort
 			break
 		}
 	}
+	if nodePort == 0 {
```
hi-im-aren, line 936: This will never be true, because we skip nodePort value assignment when port.NodePort != 0

hi-im-aren (dismissed):

pkg/resources/kafka/kafka.go
```diff
@@ -928,11 +928,15 @@ func (r *Reconciler) getK8sAssignedNodeport(log logr.Logger, eListenerName strin
 
 	var nodePort int32
 	for _, port := range nodePortSvc.Spec.Ports {
-		if port.Name == fmt.Sprintf("broker-%d", brokerId) {
+		if port.Name == fmt.Sprintf("broker-%d", brokerId) && port.NodePort != 0 {
```
hi-im-aren, line 931: `&& port.NodePort != 0` is this needed? 
`if nodePort == 0 ` below will handle the zero case anyway

hi-im-aren (approved):

stoader (approved):

PR #602, opened 2021-05-04
Move GetBrokerConfig utility function to the API

adamantal:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -
| License         | Apache 2.0

### What's in this PR?

Migrated the `util.GetBrokerConfig()` function to the API.

### Why?

Dependent packages could use that function instead of implementing the same logic that is in the operator pacakge.

### Checklist

- [x] Implementation tested
- [x] Error handling code meets the [guideline](https://github.com/banzaicloud/developer-guide/blob/master/docs/coding-style/error-handling-guide.md)
- [x] Logging code meets the guideline

PR #604, opened 2021-05-07
Adding support for broker config and broker level overrides of environment variables

adamantal:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | *yes*
| Deprecations?   | no
| Related tickets | breaks #602

### What's in this PR?

This PR extends the existing KafkaCluster CR with the following fields:`spec.brokers[].brokerConfig.envs[]`, `spec.brokerConfigGroup{}.envs[]`. These two slices add the capability of specifying environment variables in broker config group and broker level.

### Why?

There are several use cases when different Kafka brokers needs different settings: e.g. different JVM settings, thread count, limits etc.

### Feature description
When assembling the environment variables the operator takes the following steps (in the following order):
1. Default env vars are added (`CLASSPATH`, `KAFKA_OPTS`, `ENVOY_SIDECAR_STATUS`)
1. Cluster wide env vars are added from `spec.envs`
1. \* Env vars defined in broker config group level are added from `spec.brokerConfigGroup{}.envs[]`
1. \* Env vars defined in broker level are added from `spec.brokers[].brokerConfig.envs[]`
1. Special env vars that can be configured through the KafkaCluster CR are added if not specified in the previous steps (`KAFKA_LOG4J_OPTS`, `KAFKA_HEAP_OPTS`, `KAFKA_JVM_PERFORMANCE_OPTS`).

\* new steps introduced by this feature 

It is now also possible to define environment variables in *prepend mode* and *append mode*. 
- If the user puts the `+` character to the front of the name of the environment variable, it will not overwrite the value added in a previous step, but instead its value gets appended to the existing value.
- Similarly, if the user puts the `+` character after the name of the environment variable, it will not overwrite the value added in a previous step, but instead its value gets prepended to the existing value.

Example:
```yaml
apiVersion: kafka.banzaicloud.io/v1beta1
kind: KafkaCluster
metadata:
  name: kafka
spec:
  envs:
    - name: CLASSPATH+
      value: "/opt/kafka/libs/dev/*:"
    - name: +CLASSPATH
      value: ":/opt/kafka/libs/special-libs/*"
```
results the following `CLASSPATH` env var value to be added to the Kafka broker pod: `/opt/kafka/libs/dev/*:/opt/kafka/libs/extensions/*:/opt/kafka/libs/special-libs/*` where `/opt/kafka/libs/extensions/*` is the default provided by the operator. (Note that leading and trailing `:` is ignored by the JVM).

If an environment variable hasn't been set earlier, and the value is provided in prepend/append mode, the extra `+` character from the front/back is simply removed (interpreted as appending/prepending to the empty string `""`).

### Additional context

Tested locally by specifying broker config level `env` (uncommenting the `env` config in `simplekafkacluster.yaml`) and after the rolling upgrade the pods were deployed correctly.

**IMPORTANT:**
This PR breaks the API defined in #602 and released in API package `v0.8.1`. 
The following function declaration in `kafkacluster_types.go`
```go
func (b *Broker) GetBrokerConfig(brokerConfigGroups map[string]BrokerConfig) (*BrokerConfig, error)
```
has been changed to
```go
func (b *Broker) GetBrokerConfig(kafkaClusterSpec KafkaClusterSpec) (*BrokerConfig, error)
```
Applications depending on this function will break.
Taking into account that this was released few days ago we can release a new major version (`v0.9.0`) with this change causing minimal disruption.

### Checklist

- [x] Implementation tested
- [x] Updated implementation based on feedbacks


pkg/sdk/v1beta1/kafkacluster_types.go
```diff
@@ -1061,7 +1066,9 @@ func (mConfig *MonitoringConfig) GetCCJMXExporterConfig() string {
 }
 
 // GetBrokerConfig composes the brokerConfig for a given broker using the broker's config group
-func (b *Broker) GetBrokerConfig(brokerConfigGroups map[string]BrokerConfig) (*BrokerConfig, error) {
+func (b *Broker) GetBrokerConfig(kafkaClusterSpec KafkaClusterSpec) (*BrokerConfig, error) {
```
stoader, line 1069: Instead of passing in the entire `KafkaClusterSpec` why don't we pass in only what's needed by this method? e.g. `GetBrokerConfig(brokerConfigGroups map[string]BrokerConfig, envs []corev1.EnvVar)`

pkg/sdk/v1beta1/kafkacluster_types.go
```diff
@@ -1074,37 +1081,53 @@ func (b *Broker) GetBrokerConfig(brokerConfigGroups map[string]BrokerConfig) (*B
...
+	bConfig.Envs = envs
 
 	return bConfig, nil
 }
 
+func mergeEnvs(kafkaClusterSpec KafkaClusterSpec, groupConfig, bConfig *BrokerConfig) []corev1.EnvVar {
+	var envs []corev1.EnvVar
+	envs = append(envs, kafkaClusterSpec.Envs...)
+	envs = append(envs, groupConfig.Envs...)
+	return append(envs, bConfig.Envs...)
```
stoader, line 1108: if `bConfig` can be null than check for null here.

stoader (changes requested):

pkg/sdk/v1beta1/kafkacluster_types.go
adamantal: I thought about this too, but I have the impression that we may need further fields from `KafkaCluster` in the future. Also note, that if we do this later, it will brake the API again when we change that. In my opinion it is safer to leave it like this - so we can adjust the implementation later without breaking the syntax if we need some additional behaviour.

controllers/tests/common_test.go
```diff
@@ -95,14 +95,29 @@ func createMinimalKafkaClusterCR(name, namespace string) *v1beta1.KafkaCluster {
 				{
 					Id:                0,
 					BrokerConfigGroup: defaultBrokerConfigGroup,
```
schultheiszn, line 97: maybe a null value test/check would be worthful

pkg/resources/kafka/pod.go
```diff
@@ -404,15 +405,24 @@ func generateVolumeMountForSSL() []corev1.VolumeMount {
...
 
 	for _, v := range defaultEnvVars {
 		envs[v.Name] = v
 	}
 
-	for _, v := range clusterEnvVars {
-		envs[v.Name] = v
+	// merge the env variables
+	for _, envVar := range brokerConfig.Envs {
+		if envVar.Value != "" && envVar.Value[0] == '+' {
```
schultheiszn, line 417: do we strip this values somewhere? whitespaces could be a real paint in the a...

stoader, line 417: `envVar.Value` should be left trimmed otherwise `envVar.Value[0] == '+'` will evaluate to false in case there are leading spaces.

stoader (changes requested):

pkg/sdk/v1beta1/kafkacluster_types_test.go
```diff
@@ -342,32 +349,112 @@ func TestGetBrokerConfigUniqueStorage(t *testing.T) {
...
+	}
+}
+
+func TestGetBrokerConfigEnvs(t *testing.T) {
+	expected := &BrokerConfig{
+		Envs: []corev1.EnvVar{
+			{Name: "VAR", Value: "cluster"},
+			{Name: "VAR", Value: "group"},
+			{Name: "VAR", Value: "broker"},
+			/*{Name: "a", Value: "global"},
```
baluchicken, line 394: I think these were unintentionally left commented

schultheiszn (dismissed):

adamantal:
Updated the implementation based on offline feedbacks.
Changes in latest patch:
- shifted implementation from a value-based append to a name-based append (see description)
- introduced prepend mode which can be used for prepending to the `CLASSPATH` env var
- fixed tests, documentation and PR description to reflect this change

schultheiszn (approved):

chrisgacsal:
Apparently I am late to the party since the PR got merged. I'll leave my 2 cents as I am having a hard time to wrap my head around the logic used to define how values for an environment variable are combined. So in my mind it should be the exact opposite of what was implemented here, because the `+` sign in the environment variable name should indicate the order of concatenation. If the `+` appears at the end of the variable name (`CLASSPATH+`) that suggests me that the current value will be **appended** to the original value hold by the `CLASSPATH` variable and not the other way around. For me the current implementation is quite confusing. But maybe it just me.

@adamantal @stoader @schultheiszn

stoader:
This looks to be bug that went through unnoticed:
https://github.com/banzaicloud/kafka-operator/blob/master/pkg/resources/kafka/pod.go#L425 should prepend
https://github.com/banzaicloud/kafka-operator/blob/master/pkg/resources/kafka/pod.go#L434 should append

PR #609, opened 2021-05-17
bug: fix how prepand/append to environment variables works

### LINKED ISSUE #608: bug: fix how prepand/append to environment variables works
**Describe the bug**

The https://github.com/banzaicloud/kafka-operator/pull/604 has introduced the possibility to define environment variables at both cluster and broker levels. The implementation also allows _extending_ a previously set environment variable by providing its name with `+` char as suffix/prefix (depending on the desired order of concatenation) and the value to be added to the original value.

However the current implementation seems to be counter intuitive as adding the `+` as prefix resulting the new value gets appended while adding it as suffix gets the new value prepended to the original.

See: https://github.com/banzaicloud/kafka-operator/pull/604#issuecomment-840735600

N/A

**Expected behavior**
See: https://github.com/banzaicloud/kafka-operator/pull/604#issuecomment-840746546

**Screenshots**
N/A

**Additional context**
N/a

adamantal:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | caused by #604, fixes #608 

### What's in this PR?

Switch the append and prepend cases - adjust unit tests accordingly.

### Why?

The existing implementation is counter-intuitive to use.

### Checklist

- [x] Implementation tested


chrisgacsal (approved):
LGTM! Thanks for fixing this! :shipit:

schultheiszn (approved):

baluchicken (approved):

stoader (approved):

PR #619, opened 2021-05-28
Disable envoy ingress upstream circuit breaker

### LINKED ISSUE #616: Envoy ingress active max_connections limit is hit in large clusters
**Describe the bug**

In our environment with lots of Kafka clients connecting via envoy ingress to Kafka Cluster we had an incident where the client connections were dropped at the ingress level.
Investigating the cause we found that this was caused by envoy circuit breaker mechanism limiting the `max_connections` to an upstream cluster (in kafka case, it's a single broker)
https://www.envoyproxy.io/docs/envoy/latest/faq/load_balancing/disable_circuit_breaking#faq-disable-circuit-breaking

```
Circuit breakers are enabled by default and have modest default values, e.g. 1024 connections per cluster.
To disable circuit breakers, set the thresholds to the highest allowed values. 
```

Looking at the actual inbound connections we are seeing around 10k connections to individual brokers when not proxied through ingress which is higher the envoy default capacity.

Create a `KafkaCluster` using the operator, enable `envoy` as an ingress and start lots of kafka clients 

**Expected behavior**
The clients should be able to connect to kafka within the kafka brokers capacity.

**Screenshots**

Checking the envoy `upstream_cx_overflow` metric this confirms the circuit breaker kicks in, closing connections

image

**Additional context**

- Running envoy proxy 1.16.2
- https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/circuit_breaking#arch-overview-circuit-break-cluster-maximum-connections

>Cluster maximum connections: 
The maximum number of connections that Envoy will establish to all hosts in an upstream cluster. If this circuit breaker overflows the upstream_cx_overflow counter for the cluster will increment. All connections, whether active or draining, count against this limit. Even if this circuit breaker has overflowed, Envoy will ensure that a host selected by cluster load balancing has at least one connection allocated. This has the implication that the upstream_cx_active count for a cluster may be higher than the cluster maximum connection circuit breaker, with an upper bound of cluster maximum connections + (number of endpoints in a cluster) * (connection pools for the cluster). This bound applies to the sum of connections across all workers threads. See connection pooling for details on how many connection pools a cluster may have.

amuraru:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no|
| New feature?    | yes |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets | fixes #616 |

### What's in this PR?
Each envoy pod in kafka ingress (when envoy ingress is enabled)
defaults to max_connections=1024 per upstream broker
This is limiting kafka ingress capacity in larger clusters
Disabling the circuit breaking per:
https://www.envoyproxy.io/docs/envoy/latest/faq/load_balancing/disable_circuit_breaking

### Why?

Fixing performance issues in large clusters

### Additional context

See https://github.com/banzaicloud/kafka-operator/issues/616

### Checklist

- [x] Implementation tested

### DIFF
--- a/controllers/tests/kafkacluster_controller_envoy_test.go
+++ b/controllers/tests/kafkacluster_controller_envoy_test.go
@@ -92,7 +92,18 @@ func expectEnvoyConfigMap(kafkaCluster *v1beta1.KafkaCluster, eListenerTemplate 
       portValue: 9901
 staticResources:
   clusters:
-  - connectTimeout: 1s
+  - circuitBreakers:
+      thresholds:
+      - maxConnections: 1000000000
+        maxPendingRequests: 1000000000
+        maxRequests: 1000000000
+        maxRetries: 1000000000
+      - maxConnections: 1000000000
+        maxPendingRequests: 1000000000
+        maxRequests: 1000000000
+        maxRetries: 1000000000
+        priority: HIGH
+    connectTimeout: 1s
     loadAssignment:
       clusterName: broker-0
       endpoints:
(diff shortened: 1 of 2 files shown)


baluchicken (approved):
LGTM thx @amuraru

stoader (approved):

PR #641, opened 2021-07-16
Upgrade dependencies

amuraru:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no|
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |

### What's in this PR?
- Upgrade controller-gen from 0.5.0 to 0.6.1
- upgrade k8s compile deps from 0.20.2 to 0.21.3
- upgrade controller-runtime from 0.8.3 to 0.9.3
- all other transitive dependencies

### Why?
Keeping the deps up-to-date.

### Additional context
N/A

### Checklist

- [x] Implementation tested


controllers/tests/suite_test.go
```diff
@@ -82,116 +83,122 @@ func TestAPIs(t *testing.T) {
...
-	Expect(crd.Spec.Names.Kind).To(Equal("KafkaUser"))
+		err = k8sClient.Get(context.TODO(), types.NamespacedName{Name: "kafkausers.kafka.banzaicloud.io"}, crd)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(crd.Spec.Names.Kind).To(Equal("KafkaUser"))
 
-	close(done)
-}, 60)
+		close(done)
+	}()
+	Eventually(done, 240).Should(BeClosed())
```
stoader, line 200: What's the benefit of running the body of `BeforeSuite` in a goroutine and pass in a `done` channel to verify that has finished in the given time window instead of just simply relying on the `timeout` implemented by `BeforeSuite`which would not require the creation and use of the `done` channel ?

amuraru: Following to upgrade to of ginko and gomega, there was a WARNING reported pointing to: 
https://github.com/onsi/ginkgo/blob/v2/docs/MIGRATING_TO_V2.md#removed-async-testing

So they plan to remove the support for `BeforeSuite(func(done Done) {` and instead ignore the timeout if done that in the upcoming 2.0
Thought we should be proactive and refactor that

stoader: I wonder if this can be simplified to:
```
BeforeSuite(func() {
...
....
}, 240)
```
in order to get rid of introducing a go routine and a channel.

amuraru: good call - updated

pkg/sdk/go.mod
```diff
@@ -4,11 +4,14 @@ go 1.16
...
+	github.com/imdario/mergo v0.3.12
 	github.com/jetstack/cert-manager v0.15.2
+	github.com/json-iterator/go v1.1.11 // indirect
+	golang.org/x/net v0.0.0-20210716203947-853a461950ff // indirect
 	gotest.tools v2.2.0+incompatible
-	k8s.io/api v0.20.2
-	k8s.io/apimachinery v0.20.2
+	k8s.io/api v0.21.3
+	k8s.io/apimachinery v0.21.3
 	sigs.k8s.io/controller-runtime v0.8.3
```
baluchicken, line 16: Can you please bump this to v0.9.2 as well? Thanks.

amuraru: good catch @baluchicken!
Re-run `go get -u=patch ./...` in both root and pkg/sdk go modules

controllers/tests/kafkacluster_controller_externalnodeport_test.go
```diff
@@ -237,7 +237,7 @@ var _ = Describe("KafkaClusterNodeportExternalAccess", func() {
 						ContainerPort: 9733,
 						Type:          "plaintext",
 					},
-					ExternalStartingPort: 30100,
+					ExternalStartingPort: 30300,
```
amuraru, line 240: I had to change this to 30300 as 30100 seems to be already used in the new version of envtest

baluchicken (approved):

bartam1 (approved):

PR #645, opened 2021-08-09
Csr support

baluchicken:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no |
| New feature?    |yes |
| API breaks?     | yes |
| Deprecations?   |yes |

### What's in this PR?

This PR adds the ability to use k8s CertificateSigningRequest to generate certificate instead of using Vault or CertManager.
The default generation mode still uses CertManager.
The internal cert generation is not supported yet that will be covered in a separate PR.

### Checklist

- [x] Implementation tested


controllers/kafkauser_controller.go
```diff
@@ -40,45 +47,64 @@ import (
...
-			OwnerType:    &v1alpha1.KafkaUser{},
-		})
-		if err != nil {
-			if _, ok := err.(*meta.NoKindMatchError); !ok {
-				return err
-			}
-		}
+func certificateSigningRequestMapper(obj client.Object) []ctrl.Request {
+	certSigningReqAnnotations := obj.(*certsigningreqv1.CertificateSigningRequest).Annotations
+	kafkaUserResourceNamespacedName, ok := certSigningReqAnnotations[pkicommon.KafkaUserAnnotationName]
```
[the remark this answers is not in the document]
bartam1: CertificateSigningRequest resource is global

pkg/util/cert/certutil.go
```diff
@@ -99,62 +107,84 @@ func EnsureSecretPassJKS(secret *corev1.Secret) (injected *corev1.Secret, err er
 	return
 }
 
-// GenerateJKS creates a JKS with a random password from a client cert/key combination
-func GenerateJKS(clientCert, clientKey, clientCA []byte) (out, passw []byte, err error) {
-	cert, err := DecodeCertificate(clientCert)
+func GenerateJKSFromByte(certByte []byte, privateKey []byte, caCert []byte) (out, passw []byte, err error) {
```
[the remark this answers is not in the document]
bartam1: It uses that one " https://github.com/pavel-v-chernykh/keystore-go" in the background

bartam1 (dismissed):

pkg/pki/k8scsrpki/k8scsr_user.go
```diff
@@ -0,0 +1,266 @@
...
+		},
+	}
+}
+
+func generateCSRResource(csr []byte, name, namespace, signerName string,
+	annotation map[string]string) *certsigningreqv1.CertificateSigningRequest {
+	return &certsigningreqv1.CertificateSigningRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: name + "-",
+			Namespace:    namespace,
```
stoader, line 219: No need for `namespace` as CSR is a cluster scoped resource.

pkg/pki/k8scsrpki/k8scsr_user.go
```diff
@@ -0,0 +1,266 @@
...
+	if parseErr != nil {
+		return nil, parseErr
+	}
+	c.logger.Info("Generating SigningRequest")
+	csr, err := certutil.GenerateSigningRequestInPemFormat(privKey, user.GetName(), []string{""})
+	if err != nil {
+		return nil, err
+	}
+	c.logger.Info("Generating k8s csr object")
+	signingReq := generateCSRResource(csr, user.GetName(), user.GetNamespace(),
```
stoader, line 244: Using `user.GetName()` as name for CSR doesn't guarantee uniqueness cluster wide. (e.g. if there are two KafkaUser with the same name but different namespaces will map to the same CRS)

pkg/util/pki/common.go
```diff
@@ -49,6 +55,9 @@ const (
 	BrokerControllerFQDNTemplate = "%s.%s.mgt.%s"
 	// CAFQDNTemplate is the template used for the FQDN of a CA
 	CAFQDNTemplate = "%s-ca.%s.cluster.local"
+	// KafkaUserAnnotationName used in case of PKIbackend is k8s-csr to find the appropriate kafkauser in case of
+	// signing request event
+	KafkaUserAnnotationName = "banzaicloud.io/kafkauser-id"
```
stoader, line 60: How about `banzaicloud.io/owner` or `banzaicloud.io/belongs-to`?

This is more generic at allows introducing different type of owners in the future next to KafkaUser

pkg/pki/k8scsrpki/k8scsr.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"github.com/banzaicloud/kafka-operator/api/v1beta1"
+	"github.com/banzaicloud/kafka-operator/pkg/util/pki"
+
+	"github.com/go-logr/logr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const DependingCsrAnnotation string = "banzaicloud.io/depending-csr"
```
stoader, line 25: This could be `banzaicloud.io/owner` and the value for annotation placed onto the secret `csr://<csr name>` or `csr@<csr name>` for example.

stoader (changes requested):

pkg/pki/k8scsrpki/k8scsr_user.go
```diff
@@ -0,0 +1,266 @@
...
+		signingReq, err = c.generateAndCreateCSR(ctx, clientKey, user)
+		if err != nil {
+			return nil, err
+		}
+
+		if err = c.secretUpdateAnnotation(ctx, secret, signingReq.GetName()); err != nil {
+			return nil, err
+		}
+	} else if err != nil {
+		return nil, err
```
adamantal, line 78: Maybe add a little detail to the error?
```suggestion
		return nil, errors.WrapIfWithDetails(err, "failed to get user's secret from K8s", "secretName", user.Spec.SecretName, "namespace", user.GetNamespace())
```

pkg/pki/k8scsrpki/k8scsr_user.go
```diff
@@ -0,0 +1,266 @@
...
+				if err != nil {
+					return nil, err
+				}
+				secret.TypeMeta = typeMeta
+				return nil, errorfactory.New(errorfactory.ResourceNotReady{},
+					errors.New("instance not found"), "kubernetes deleted the csr request",
+					"csrName", signingRequestGenName)
+			}
+		} else if err != nil {
+			return nil, err
```
adamantal, line 123:
```suggestion
			return nil, errors.WrapIfWithDetails(err, "failed to get signing request from K8s", "signingRequestName", signingRequestGenName, "namespace", secret.GetNamespace())
```

pkg/pki/k8scsrpki/k8scsr_user.go
```diff
@@ -0,0 +1,266 @@
...
+	if user.Spec.IncludeJKS {
+		// we don't have an existing one - make a new one
+		if value, ok := secret.Data[v1alpha1.TLSJKSKeyStore]; !ok || len(value) == 0 {
+			jks, jksPasswd, err = certutil.GenerateJKS(certBundleX509, secret.Data[corev1.TLSPrivateKeyKey])
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+	secret.Data[v1alpha1.TLSJKSKeyStore] = jks
```
adamantal, line 171: It seems to me when `user.Spec.IncludeJKS` is true and `secret.Data[v1alpha1.TLSJKSKeyStore]` already has a value set, the reconciler overwrites it with an empty jks and jksPasswd string. Is this inteded?

adamantal (commented):
Nice work, thanks for putting the effort! 
Had some comments, but mostly small ones.

pkg/sdk/v1alpha1/kafkauser_types.go
```diff
@@ -85,3 +88,16 @@ func (spec *KafkaUserSpec) GetIfCertShouldBeCreated() bool {
 	}
 	return true
 }
+
+//GetAnnotations returns Annotations to use for Envoy generated Deployment and Pods
+func (spec *KafkaUserSpec) GetAnnotations() map[string]string {
+	return cloneAnnotationMap(spec.Annotations)
+}
+
+func cloneAnnotationMap(original map[string]string) map[string]string {
```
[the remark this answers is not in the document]
baluchicken: Thanks, I cannot call util.MergeAnnotations here since it is in a different package. kafka-operator and this one is in the pkg/sdk. If I call that function I will introduce a concurrent dependency.

pkg/pki/k8scsrpki/k8scsr_user.go
baluchicken: That is right that why we are using generate name for the CSR, see https://github.com/banzaicloud/kafka-operator/blob/882ff6f80e5eee652b1e153c6ec55604b550cf65/pkg/pki/k8scsrpki/k8scsr_user.go#L219

charts/kafka-operator/templates/crds.yaml
```diff
@@ -15756,6 +15765,12 @@ spec:
           spec:
             description: KafkaUserSpec defines the desired state of KafkaUser
             properties:
+              annotations:
```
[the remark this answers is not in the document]
baluchicken: They are refreshed for sure since they are generated by the contoller-gen.

pkg/sdk/v1alpha1/kafkauser_types.go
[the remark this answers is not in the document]
stoader: but we could still have a single `cloneAnnotationMap` across the pkg/sdk package.

charts/kafka-operator/templates/crds.yaml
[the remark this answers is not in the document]
adamantal: Thanks

pkg/pki/k8scsrpki/k8scsr_user.go
baluchicken: Oops nice catch. Thanks.

adamantal (dismissed):
+1
Thanks for taking care the review items.

stoader (approved):

adamantal (approved):

PR #670, opened 2021-09-16
Fix unnecessary reconcile caused by envoy pdb struct

baluchicken:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        |yes |
| New feature?    | no|
| API breaks?     | no|
| Deprecations?   | no|

### What's in this PR?

This PR fixes unnecessary reconcile which caused by the pod distruption budget field under envoy config. Since the struct was not handled as a pointer and has an inline struct and labeled omitempty the inline struct always instantiated which caused diff inside kubernetes.

### Checklist

- [x] Implementation tested


chrisgacsal (approved):

adamantal (approved):
Thanks @baluchicken !

PR #678, opened 2021-10-03
Expose http healthcheck endpoint to envoy anycast ingress

amuraru:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no|
| New feature?    | yes |
| API breaks?     | no|
| Deprecations?   | yes |
| Related tickets | fixes #X |

### What's in this PR?

When envoy is selected as an ingress for kafka, a new endpoint `:19900/healthcheck` is exposed by each envoy pod and in turn available through the Envoy Service.

This is implemented as an active heath-check in envoy configured to probe jmx exporter /-/healthy endpoint for each broker.

This is an http port that can be probed to verify if there is at least one broker up in cluster.

### Why?

This http healthcheck endpoint can be used in fronting LoadBalancers when envoy ingress is used to expose Kafka externally.
Before that - we had to rely to tcp connect LB health-checks but these are usually unreliable and moreover creates unneeded broker connections.

### Checklist

- [x] Implementation tested


adamantal (approved):
LGTM

PR #687, opened 2021-10-21
Ensure Kafka broker pods are maintained in Service rotation until clean shutdown completes

amuraru:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no|
| New feature?    | yes |
| API breaks?     | no|
| Deprecations?   | no|

### What's in this PR?

1/ Configurable `terminationGracePeriodSeconds ` 
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#podspec-v1-core

Default 120s, can be overridden for larger Kafka brokers that may take longer to terminate

2/ Set `publishNotReadyAddresses=true` for Kafka headless service
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#servicespec-v1-core

This is set to ensure that a broker entering the Terminating phase is still advertised in the headless DNS services so clients can continue reaching to it until it migrates all the leader partitions and shuts down successfully

There is also a drawback for this though, the kafka pod will be advertised 
in the headless service as soon as the pod is started and not waiting for
the readiness probe to be successful. This was considered to be a non-issue
as:
1/ the operator doesn't set yet a readiness probe for kafka pod
2/ even if there is one, the kafka clients will retry if they hit a kafka
broker not yet fully ready

### Why?

On kafka pod clean termination (e.g. when pod is evicted during an underlying k8s node maintenance,
or rescheduling) kubelet sends a SIGTERM to the pod and kafka process initiates the clean shutdown procedure:
https://kafka.apache.org/documentation/#basic_ops_restarting

During this period and upon `terminationGracePeriodSeconds` the kafka broker will:
- flush and sync all the segment logs
- migrate any partitions the server is the leader for to other replicas

### Additional context

In our environment we noticed client being unable to connect to terminating brokers.

### Checklist

- [x] Implementation tested


adamantal:
Hey!
This one's been around for a while, got any objections/thoughts @baluchicken @stoader ?

baluchicken (approved):

stoader (approved):

PR #691, opened 2021-10-28
Add an optional field in KafkaCluster resource to specify cruise control reporter image

amuraru:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no|
| New feature?    | yes |
| API breaks?     | no |
| Deprecations?   | no |

### What's in this PR?
Previously the same CruiseControl.Image field was used for both
cruise-control server image version but also kafka pod metrics reporter image

### Why?

In the case only the cruise-control server version needs to be updated
and avoid a kafka cluster rolling restart, `ClusterMetricsReporterImage` or `BrokerConfig.MetricsReporterImage` can be kept and only the CruiseControl.Image upgraded

### Checklist

- [x] Implementation tested

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ bin/golangci-lint: bin/golangci-lint-${GOLANGCI_VERSION}
 	@ln -sf golangci-lint-${GOLANGCI_VERSION} bin/golangci-lint
 bin/golangci-lint-${GOLANGCI_VERSION}:
 	@mkdir -p bin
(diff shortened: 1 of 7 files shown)


baluchicken (approved):
LGTM thx @amuraru

adamantal (approved):
LGTM
Currently the default CC image used by the operator contains the metrics reporter jar too. I think it would be beneficial for us too to have this separated.

PR #698, opened 2021-11-16
Add possibility for user provided credentials for listeners and internal clients

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no|
| New feature?    | yes |

### What's in this PR?
Users are able to use their certificates when configuring SSL for listeners and internal clients (cruise control, koperator, cruise control metrics reporter) There are two API fields for this configuration. ServerSSLCertSecretName for the listeners and ClientSSLCertSecretName for the clients.

### Why?
Since kafka supports multiple certificate per multiple listener we should leverage this feature by enhancing the operator to support that. We should give the possibility to add custom certificates for these listeners and for the clients for internal communication.


api/v1beta1/kafkacluster_types.go
```diff
@@ -303,10 +303,12 @@ type StorageConfig struct {
...
 //ListenersConfig defines the Kafka listener types
 type ListenersConfig struct {
-	ExternalListeners  []ExternalListenerConfig `json:"externalListeners,omitempty"`
-	InternalListeners  []InternalListenerConfig `json:"internalListeners"`
-	SSLSecrets         *SSLSecrets              `json:"sslSecrets,omitempty"`
-	ServiceAnnotations map[string]string        `json:"serviceAnnotations,omitempty"`
+	ExternalListeners []ExternalListenerConfig `json:"externalListeners,omitempty"`
+	InternalListeners []InternalListenerConfig `json:"internalListeners"`
+	// +kubebuilder:validation:Pattern=^[a-z0-9\-]+
+	ClientSSLCertSecretName string            `json:"clientSSLCertSecretName,omitempty"`
```
stoader, line 309: Please add description what is this field for?

baluchicken (changes requested):
I will go through once again:) Here are my first findings.

api/v1beta1/kafkacluster_types.go
```diff
@@ -436,6 +444,12 @@ type CommonListenerSpec struct {
 	// +kubebuilder:validation:Enum=ssl;plaintext;sasl_ssl;sasl_plaintext
 	Type SecurityProtocol `json:"type"`
 	// +kubebuilder:validation:Pattern=^[a-z0-9\-]+
+	// ServerSSLCertSecretName is where custom server SSL certificate can be provided.
+	// Secret has to contain the keystore and truststore in jks and the password for them in base64 encoded format format.
+	// Data fields must be: keystore.jks, truststore.jks, password
+	// SecurityProtocol has to be set to ssl.
+	ServerSSLCertSecretName string `json:"serverSSLCertSecretName,omitempty"`
```
stoader, line 451: Since this is a reference to a Kubernetes resource in the same namespace the type of this field could be `corev1.LocalObjectReference` to be in line with how Kubernetes resources references other resources in the same namespace.

baluchicken (changes requested):

api/v1beta1/kafkacluster_types.go
```diff
@@ -303,10 +303,18 @@ type StorageConfig struct {
...
 type ListenersConfig struct {
-	ExternalListeners  []ExternalListenerConfig `json:"externalListeners,omitempty"`
-	InternalListeners  []InternalListenerConfig `json:"internalListeners"`
-	SSLSecrets         *SSLSecrets              `json:"sslSecrets,omitempty"`
-	ServiceAnnotations map[string]string        `json:"serviceAnnotations,omitempty"`
+	ExternalListeners []ExternalListenerConfig `json:"externalListeners,omitempty"`
+	InternalListeners []InternalListenerConfig `json:"internalListeners"`
+	// +kubebuilder:validation:Pattern=^[a-z0-9\-]+
+	// ClientSSLCertSecretName is where custom client SSL certificate can be provided
+	// It can be used by the koperator, cruise control, cruise control metrics reporter
```
stoader, line 310:
```suggestion
	// It can be used by the koperator, cruise control
```

pkg/util/util.go
```diff
@@ -124,7 +133,17 @@ func ConvertStringToInt32(s string) int32 {
...
-		if listener.Type.IsSSL() {
+		if listener.UsedForInnerBrokerCommunication && listener.Type.IsSSL() {
+			enabled = true
+			break
+		}
+	}
+	return enabled
+}
+
+func IsCruiseControlMetricsNeedSSL(l []v1beta1.InternalListenerConfig, brokerPort int32) (enabled bool) {
```
baluchicken, line 144: We should rename this function since this not relates to CC metrics at all as far as I can tell from the code.

pkg/resources/kafka/configmap.go
```diff
@@ -247,17 +244,50 @@ func generateListenerSpecificConfig(l *v1beta1.ListenersConfig, log logr.Logger)
 	if err := config.Set("listeners", listenerConfig); err != nil {
 		log.Error(err, "setting listeners parameter in broker configuration resulted an error")
 	}
-
 	return config
 }
 
+func generateListenerSSLConfig(config *properties.Properties, name string, ssLFormat string, passwordKeyMap map[string]string) {
```
baluchicken, line 250: Typo in sslFormat

baluchicken (changes requested):

pkg/resources/cruisecontrol/cruisecontrol.go
```diff
@@ -145,19 +145,21 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
...
+	var clientPass string
+	if r.KafkaCluster.Spec.ListenersConfig.ClientSSLCertSecretName != "" || r.KafkaCluster.Spec.ListenersConfig.SSLSecrets != nil {
+		clientNamespacedName := types.NamespacedName{Name: fmt.Sprintf(pkicommon.BrokerControllerTemplate, r.KafkaCluster.Name), Namespace: r.KafkaCluster.Namespace}
+		if r.KafkaCluster.Spec.ListenersConfig.ClientSSLCertSecretName != "" {
+			clientNamespacedName = types.NamespacedName{Name: r.KafkaCluster.Spec.ListenersConfig.ClientSSLCertSecretName, Namespace: r.KafkaCluster.Namespace}
+		}
+		clientSecret := &corev1.Secret{}
+		if err := r.Client.Get(context.TODO(), clientNamespacedName, clientSecret); err != nil {
+			if apierrors.IsNotFound(err) {
+				return "", errorfactory.New(errorfactory.ResourceNotReady{}, err, "client secret not ready")
```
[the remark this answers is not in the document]
bartam1: Nice catch!

api/v1beta1/kafkacluster_types.go
bartam1: Make sense.
Thank you!

pkg/util/util.go
bartam1: True, but its only used when we check that cruise control metrics need SSL or not so somehow relates to it. What do you think should we rename this for sure

pkg/resources/kafka/kafka.go
```diff
@@ -417,41 +426,93 @@ func arePodsAlreadyDeleted(pods []corev1.Pod, log logr.Logger) bool {
...
+			if globKeyPass == "" || iListener.ServerSSLCertSecretName != "" {
+				serverSecret, err = getListenerSSLCertSecret(r.Client, iListener.CommonListenerSpec, r.KafkaCluster.Name, r.KafkaCluster.Namespace)
+				if err != nil {
+					return nil, nil, err
+				}
+				pair[iListener.Name] = string(serverSecret.Data[v1alpha1.PasswordKey])
+			}
+			if r.KafkaCluster.Spec.ListenersConfig.SSLSecrets != nil {
+				if globKeyPass == "" {
+					globKeyPass = string(serverSecret.Data[v1alpha1.PasswordKey])
```
baluchicken, line 485: I would refactor this using else if instead of this globKeyPass.

pkg/resources/kafka/kafka.go
```diff
@@ -417,41 +426,93 @@ func arePodsAlreadyDeleted(pods []corev1.Pod, log logr.Logger) bool {
...
+		CN = cert.Subject.String()
+		clientPass = string(clientSecret.Data[v1alpha1.PasswordKey])
+	}
+	return clientPass, CN, nil
+}
 
-func (r *Reconciler) getServerAndClientDetails() (string, string, []string, error) {
-	if r.KafkaCluster.Spec.ListenersConfig.SSLSecrets == nil {
-		return "", "", []string{}, nil
+func getListenerSSLCertSecret(client client.Reader, commonSpec v1beta1.CommonListenerSpec, clusterName string, clusterNamespace string) (*corev1.Secret, error) {
```
baluchicken, line 453: Instead of adding the client and the cluster related things to the parameter list why dont we make this function to a method?

pkg/resources/kafka/pod.go
```diff
@@ -34,7 +34,7 @@ import (
 	pkicommon "github.com/banzaicloud/koperator/pkg/util/pki"
 )
 
-func (r *Reconciler) pod(id int32, brokerConfig *v1beta1.BrokerConfig, pvcs []corev1.PersistentVolumeClaim, log logr.Logger) runtime.Object {
+func (r *Reconciler) pod(id int32, brokerConfig *v1beta1.BrokerConfig, listenerConfig *v1beta1.ListenersConfig, pvcs []corev1.PersistentVolumeClaim, log logr.Logger) runtime.Object {
```
baluchicken, line 37: Why do we need to add listenerConfig through parameter, I think that information is available through r.KafkaCLuster.ListenerConfig

pkg/resources/kafka/pod.go
```diff
@@ -218,16 +217,17 @@ func getInitContainers(brokerConfig *v1beta1.BrokerConfig, kafkaClusterSpec v1be
...
 
 	volumeMounts = append(volumeMounts, dataVolumeMount...)
 
-	if hasSSLSecrets {
-		volumeMounts = append(volumeMounts, generateVolumeMountForSSL()...)
+	if listenersConfig.SSLSecrets != nil || listenersConfig.ClientSSLCertSecretName != "" {
+		volumeMounts = append(volumeMounts, generateVolumeMountForClientSSLCerts())
 	}
 
+	volumeMounts = append(volumeMounts, generateVolumeMountForListenerCerts(listenersConfig)...)
```
baluchicken, line 230: This should be guarded by an if. ListenerCerts are only needed when SSl is enabled.

baluchicken (changes requested):

pkg/util/util.go
```diff
@@ -412,3 +431,37 @@ func GenerateEnvoyResourceName(resourceNameFormat string, resourceNameWithScopeF
...
+			err = errorfactory.New(errorfactory.ResourceNotReady{}, err, "controller secret not found")
+		}
+		return config, err
+	}
+	clientCert := tlsKeys.Data[corev1.TLSCertKey]
+	clientKey := tlsKeys.Data[corev1.TLSPrivateKeyKey]
+	caCert := tlsKeys.Data[v1alpha1.CoreCACertKey]
+	x509ClientCert, err := tls.X509KeyPair(clientCert, clientKey)
+	if err != nil {
+		err = errorfactory.New(errorfactory.InternalError{}, err, "could not decode controller certificate")
```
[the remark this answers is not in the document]
bartam1: Sorry, I dont get it :) What do you mean?

pkg/resources/kafka/kafka.go
```diff
@@ -186,10 +188,17 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
...
-	serverPass, clientPass, superUsers, err := r.getServerAndClientDetails()
+	serverPasses, superUsers, err := r.getServerPasswordKeysAndUsers()
+	if err != nil {
+		return err
+	}
+	clientPass, superUser, err := r.getClientPasswordKeyAndUser()
 	if err != nil {
 		return err
 	}
+	if superUser != "" {
```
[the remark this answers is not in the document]
bartam1: The r.getClientPasswordKeyAndUser() returns an empty string if there is no SSL. and later if there is no condition
the superUsers = append(superUsers, superUser)  append an empty string at the end of the superUsers.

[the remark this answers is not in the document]
bartam1: if there is emty string array than later an empty  superuser will be in broker configmap like CN=""

pkg/resources/kafka/kafka.go
```diff
@@ -250,7 +259,7 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
 				return errors.WrapIfWithDetails(err, "failed to reconcile resource", "resource", o.GetObjectKind().GroupVersionKind())
 			}
 		}
-		o := r.pod(broker.Id, brokerConfig, pvcs, log)
+		o := r.pod(broker.Id, brokerConfig, &r.KafkaCluster.Spec.ListenersConfig, pvcs, log)
```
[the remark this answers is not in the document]
bartam1: Nica catch! I tried to avoid this kind of mistakes but here I missed it.

pkg/resources/kafka/kafka.go
bartam1: If we use method here these parameters will be passed as hidden. I wouldn't add this func into the reconcile structure because it's only a helper function, also It doesn't change any data on the reconciler structure.

pkg/resources/kafka/pod.go
bartam1: Than there will be a double check loop because we need to check SSL on every listener. 
In my solution generateVolumeMountForListenerCerts will return an empty array if there is no SSL on any listener and appending an empty array into an array equal the same array than before append.

pkg/resources/kafka/pod.go
bartam1: True, nice catch!

pkg/resources/kafka/configmap.go
```diff
@@ -63,46 +63,36 @@ func (r *Reconciler) getConfigProperties(bConfig *v1beta1.BrokerConfig, id int32
...
-				log.Error(err, "setting cruise.control.metrics.reporter.ssl.keystore.password parameter in broker configuration resulted an error")
-			}
+		if err := config.Set("cruise.control.metrics.reporter.ssl.keystore.password", clientPass); err != nil {
+			log.Error(err, "setting cruise.control.metrics.reporter.ssl.keystore.password parameter in broker configuration resulted an error")
 		}
 	}
-
+	if err := config.Set("password.encoder.secret", "secret"); err != nil {
+		log.Error(err, "setting cruise.control.metrics.reporter.ssl.keystore.password parameter in broker configuration resulted an error")
+	}
```
baluchicken, line 91: I think this should be set only if SSL is enabled.

pkg/util/util.go
```diff
@@ -124,7 +133,21 @@ func ConvertStringToInt32(s string) int32 {
...
+			enabled = true
+			break
+		}
+	}
+	return enabled
+}
+
+func IsClientSSLSecretPresent(lconfig v1beta1.ListenersConfig) bool {
+	return lconfig.SSLSecrets != nil || lconfig.ClientSSLCertSecret.Name != ""
+}
```
baluchicken, line 146: This can be a method for ListenerConfig.

pkg/resources/kafka/configmap.go
```diff
@@ -63,46 +63,36 @@ func (r *Reconciler) getConfigProperties(bConfig *v1beta1.BrokerConfig, id int32
...
 	}
 
-	// Add SSL configuration
-	if r.KafkaCluster.Spec.ListenersConfig.SSLSecrets != nil {
-		if err := config.Set("ssl.keystore.location", serverKeystorePath+"/"+v1alpha1.TLSJKSKeyStore); err != nil {
-			log.Error(err, "setting ssl.keystore.location parameter in broker configuration resulted an error")
+	// Add Cruise Control SSL configuration
+	brokerPort, err := kafkautils.GetBrokerContainerPort(r.KafkaCluster)
+
+	if util.IsListenerPortUseSSL(r.KafkaCluster.Spec.ListenersConfig.InternalListeners, brokerPort) {
```
baluchicken, line 69: Why cannot we use `IsSSLEnabledForInternalCommunication` utility function instead?

pkg/resources/kafka/configmap.go
```diff
@@ -63,46 +63,36 @@ func (r *Reconciler) getConfigProperties(bConfig *v1beta1.BrokerConfig, id int32
...
+		log.Error(err, "setting cruise.control.metrics.reporter.ssl.keystore.password parameter in broker configuration resulted an error")
+	}
 	// Add Cruise Control Metrics Reporter configuration
 	if err := config.Set("metric.reporters", "com.linkedin.kafka.cruisecontrol.metricsreporter.CruiseControlMetricsReporter"); err != nil {
 		log.Error(err, "setting metric.reporters in broker configuration resulted an error")
 	}
-	brokerPort, err := kafkautils.GetBrokerContainerPort(r.KafkaCluster)
 	if err != nil {
 		log.Error(err, "getting Kafka broker port for Cruise Control failed")
 	}
```
baluchicken, line 98: This is a leftover code. But I would add the `brokerPort, err := kafkautils.GetBrokerContainerPort(r.KafkaCluster)` before this.

pkg/resources/kafka/pod.go
```diff
@@ -218,16 +217,17 @@ func getInitContainers(brokerConfig *v1beta1.BrokerConfig, kafkaClusterSpec v1be
...
-	kafkaClusterName string, hasSSLSecrets bool) []corev1.VolumeMount {
+	listenersConfig *v1beta1.ListenersConfig, kafkaClusterName string) []corev1.VolumeMount {
 	volumeMounts := make([]corev1.VolumeMount, 0, len(brokerConfigVolumeMounts))
 	volumeMounts = append(volumeMounts, brokerConfigVolumeMounts...)
 
 	volumeMounts = append(volumeMounts, dataVolumeMount...)
 
-	if hasSSLSecrets {
-		volumeMounts = append(volumeMounts, generateVolumeMountForSSL()...)
+	if listenersConfig.SSLSecrets != nil || listenersConfig.ClientSSLCertSecret.Name != "" {
```
baluchicken, line 226: The introduced utility function can be used here.

pkg/resources/kafka/kafka.go
```diff
@@ -417,41 +426,94 @@ func arePodsAlreadyDeleted(pods []corev1.Pod, log logr.Logger) bool {
 	}
 	return true
 }
+func (r *Reconciler) getClientPasswordKeyAndUser() (string, string, error) {
+	var clientPass, CN string
+	if r.KafkaCluster.Spec.ListenersConfig.ClientSSLCertSecret.Name != "" || r.KafkaCluster.Spec.ListenersConfig.SSLSecrets != nil {
```
baluchicken, line 431: The introduced utility function can be used here.

pkg/resources/kafka/pod.go
```diff
@@ -372,39 +373,84 @@ func generateDataVolumeAndVolumeMount(pvcs []corev1.PersistentVolumeClaim) (volu
...
+		}
+		ret = append(ret, vm)
+	}
+	for _, eListener := range listenerConfig.ExternalListeners {
+		if eListener.CommonListenerSpec.Type != v1beta1.SecurityProtocolSSL {
+			continue
+		}
+		vm := corev1.VolumeMount{
+			Name:      fmt.Sprintf(iListenerSSLCertVolumeNameTemplate, eListener.Name),
+			MountPath: fmt.Sprintf(iListenerServerKeyStorePathTemplate, serverKeystorePath, eListener.CommonListenerSpec.Name),
```
baluchicken, line 443: Misleading template name, since this part is for externalListeners.

baluchicken (changes requested):

pkg/resources/kafka/configmap.go
```diff
@@ -247,17 +245,50 @@ func generateListenerSpecificConfig(l *v1beta1.ListenersConfig, log logr.Logger)
 	if err := config.Set("listeners", listenerConfig); err != nil {
 		log.Error(err, "setting listeners parameter in broker configuration resulted an error")
 	}
-
 	return config
 }
 
+func generateListenerSSLConfig(config *properties.Properties, name string, sSLFormat string, passwordKeyMap map[string]string, log logr.Logger) {
+	if sSLFormat == "JKS" {
```
baluchicken, line 252: typo in sslFormat

pkg/resources/kafka/pod.go
```diff
@@ -372,39 +373,84 @@ func generateDataVolumeAndVolumeMount(pvcs []corev1.PersistentVolumeClaim) (volu
...
-		{
-			Name: serverKeystoreVolume,
-			VolumeSource: corev1.VolumeSource{
-				Secret: &corev1.SecretVolumeSource{
-					SecretName:  fmt.Sprintf(pkicommon.BrokerServerCertTemplate, clusterName),
-					DefaultMode: util.Int32Pointer(0644),
-				},
+func generateVolumeForListenersCertsFromCommonSpec(commonSpec v1beta1.CommonListenerSpec, clusterName string) corev1.Volume {
+	// Use default one if custom has not specified
+	secretName := fmt.Sprintf(pkicommon.BrokerControllerTemplate, clusterName)
```
baluchicken, line 378: Using BrokerControllerTemplate here is wrong. SInce this is the server side we should use the BrokerServerCertTemplate.

baluchicken (changes requested):

api/v1beta1/kafkacluster_types.go
```diff
@@ -303,10 +303,17 @@ type StorageConfig struct {
...
-	ServiceAnnotations map[string]string        `json:"serviceAnnotations,omitempty"`
+	ExternalListeners []ExternalListenerConfig `json:"externalListeners,omitempty"`
+	InternalListeners []InternalListenerConfig `json:"internalListeners"`
+	// ClientSSLCertSecret is a reference to that secret where custom client SSL certificate can be provided
+	// It can be used by the koperator, cruise control to communicate on SSL with
+	// the internal listener which is used for the interbroker communication.
+	// The included certificate have to be signed by the same CA as the corresponding internal listener's server certificate.
+	// Secret has to contain the keystore and truststore in jks and the password for them in base64 encoded format.
+	// Data fields must be: keystore.jks, truststore.jks, password
+	ClientSSLCertSecret corev1.LocalObjectReference `json:"clientSSLCertSecret,omitempty"`
```
stoader, line 314: Since this is not a listener related config but rather a config for internal clients IMHO it should be moved outside of `ListenersConfig`

api/v1beta1/kafkacluster_types.go
```diff
@@ -435,6 +442,11 @@ type InternalListenerConfig struct {
 type CommonListenerSpec struct {
 	// +kubebuilder:validation:Enum=ssl;plaintext;sasl_ssl;sasl_plaintext
 	Type SecurityProtocol `json:"type"`
+	// ServerSSLCertSecret is a reference to that secret where custom server SSL certificate can be provided.
+	// Secret has to contain the keystore and truststore in jks and the password for them in base64 encoded format.
+	// Data fields must be: keystore.jks, truststore.jks, password
+	// SecurityProtocol has to be set to ssl.
+	ServerSSLCertSecret corev1.LocalObjectReference `json:"serverSSLCertSecret,omitempty"`
```
stoader, line 449: I believe this should be a pointer.

pkg/util/util.go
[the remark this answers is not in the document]
baluchicken: Like `return config, errorfactory.New(errorfactory.InternalError{}, err, "could not decode controller certificate")`

pkg/resources/kafka/configmap.go
stoader, line 252: This is rather `certificateStoreType` than `sslFormat`

pkg/resources/kafka/kafka.go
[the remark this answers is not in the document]
baluchicken: I see I took a quick look once again and yeah that one is required:) Thanks

pkg/resources/kafka/kafka.go
baluchicken: I also checked this one and globKeyPass is also needed

pkg/resources/kafka/configmap.go
```diff
@@ -247,17 +245,50 @@ func generateListenerSpecificConfig(l *v1beta1.ListenersConfig, log logr.Logger)
...
+		keyStoreLocKey := fmt.Sprintf("listener.name.%s.ssl.keystore.location", name)
```
stoader, line 253: All these keys and their values could be stored in a `map[string]string` and than loop through the items of the map and invoke `config.Set` for each item.

e.g.:
```
listenerSSLConfig := map[string]string{
  "fmt.Sprintf("listener.name.%s.ssl.truststore.location", name)": namedKeystorePath+"/"+v1alpha1.TLSJKSKeyStore,
...
...
}

for k, v := range listenerSSLConfig {
  if err := config.Set(k, v); err != nil {
 			log.Error(err, fmt.Sprintf("setting %s parameter in broker configuration resulted an error", k))
 		}
}
```

stoader (changes requested):

pkg/resources/kafka/configmap.go
bartam1: Yes, I had to set this one to test the stb 381 and accidentally pushed it. If its not set you cant change the SSL keystore password dynamicly

pkg/resources/kafka/configmap.go
bartam1: It can, nice catch

pkg/resources/kafka/pod.go
bartam1: nice catch

api/v1beta1/kafkacluster_types.go
bartam1: I think it is related to the listener because this client cert corresponds to the custom server cert which can be set by the user.  I think if we move it out somewhere else, it can mislead the user. Very likely he wouldn't set this also with the custom internal listener

stoader: if this cert will be used by `koperator` for example than it's a koperator config for me.

pkg/resources/kafka/configmap.go
bartam1: Thank you! I have changed it to that one

api/v1beta1/kafkacluster_types.go
```diff
@@ -80,6 +80,14 @@ type KafkaClusterSpec struct {
 	// Add the "+" suffix to append.
 	Envs                    []corev1.EnvVar `json:"envs,omitempty"`
 	KubernetesClusterDomain string          `json:"kubernetesClusterDomain,omitempty"`
+	// ClientSSLCertSecret is a reference to that secret where custom client SSL certificate can be provided.
```
stoader, line 83: How about something like this?

```
ClientSSLCertSecret is a reference to the Kubernetes secret that contains the client certificate for koperator, cruise control, cruise control metrics reporter when connecting brokers through SSL and interbroker communication as well. The client certificate must share the same chain of trust as the server certificate used by the corresponding internal listener. The secret must contain the keystore, truststore jks files and the password for them in base64 encoded format under the keystore.jks, truststore.jks, password data fields.

If this field is omitted koperator will auto-create a self-signed client certificate.
```

api/v1beta1/kafkacluster_types.go
```diff
@@ -435,11 +443,23 @@ type InternalListenerConfig struct {
 type CommonListenerSpec struct {
 	// +kubebuilder:validation:Enum=ssl;plaintext;sasl_ssl;sasl_plaintext
 	Type SecurityProtocol `json:"type"`
+	// ServerSSLCertSecret is a reference to that secret where custom server SSL certificate can be provided.
```
stoader, line 446:
```
ServerSSLCertSecret is a reference to the Kubernetes secret that contains the server certificate for the listener to be used for SSL communication. The secret must contain the keystore, truststore jks files and the password for them in base64 encoded format under the keystore.jks, truststore.jks, password data fields.

If this field is omitted koperator will auto-create a self-signed server certificate.
```

pkg/kafkaclient/config.go
```diff
@@ -40,8 +43,16 @@ func ClusterConfig(client client.Client, cluster *v1beta1.KafkaCluster) (*KafkaC
...
-		tlsConfig, err := pki.GetPKIManager(client, cluster, v1beta1.PKIBackendProvided, log).GetControllerTLSConfig()
+	if clientutil.UseSSL(cluster) {
+		var tlsConfig *tls.Config
+		var err error
+		if cluster.Spec.GetClientSSLCertSecretName() != "" {
+			tlsConfig, err = util.GetClientTLSConfig(client, types.NamespacedName{Name: cluster.Spec.GetClientSSLCertSecretName(), Namespace: cluster.Namespace})
+		} else if cluster.Spec.ListenersConfig.SSLSecrets != nil {
+			tlsConfig, err = pki.GetPKIManager(client, cluster, v1beta1.PKIBackendProvided, log).GetControllerTLSConfig()
+		} else {
+			err = errors.Errorf("client certificate has not specified for the controller but kafka listener requests it")
```
stoader, line 54: errors.New("either 'clientSSLCertSecret' or 'sslSecrets' must be specified as internal listener used for inner communication uses SSL")

pkg/resources/cruisecontrol/cruisecontrol.go
```diff
@@ -145,19 +145,24 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
...
-	if r.KafkaCluster.Spec.ListenersConfig.SSLSecrets == nil {
-		return "", nil
-	}
-	clientName := types.NamespacedName{Name: fmt.Sprintf(pkicommon.BrokerControllerTemplate, r.KafkaCluster.Name), Namespace: r.KafkaCluster.Namespace}
-	clientSecret := &corev1.Secret{}
-	if err := r.Client.Get(context.TODO(), clientName, clientSecret); err != nil {
-		if apierrors.IsNotFound(err) {
-			return "", errorfactory.New(errorfactory.ResourceNotReady{}, err, "client secret not ready")
+	var clientPass string
+	if r.KafkaCluster.Spec.IsClientSSLSecretPresent() {
```
stoader, line 149: Shouldn't be an error returned if `r.KafkaCluster.Spec.IsClientSSLSecretPresent()` is `false`?

pkg/resources/cruisecontrol/deployment.go
```diff
@@ -41,7 +41,7 @@ func (r *Reconciler) deployment(podAnnotations map[string]string) runtime.Object
 	volume = append(volume, r.KafkaCluster.Spec.CruiseControlConfig.Volumes...)
 	volumeMount = append(volumeMount, r.KafkaCluster.Spec.CruiseControlConfig.VolumeMounts...)
 
-	if r.KafkaCluster.Spec.ListenersConfig.SSLSecrets != nil && util.IsSSLEnabledForInternalCommunication(r.KafkaCluster.Spec.ListenersConfig.InternalListeners) {
+	if r.KafkaCluster.Spec.IsClientSSLSecretPresent() && util.IsSSLEnabledForInternalCommunication(r.KafkaCluster.Spec.ListenersConfig.InternalListeners) {
```
stoader, line 44: Shouldn't be an error returned  when `util.IsSSLEnabledForInternalCommunication` returns `true` while `r.KafkaCluster.Spec.IsClientSSLSecretPresent()` returns `false`

pkg/resources/kafka/kafka.go
```diff
@@ -974,14 +1037,14 @@ func getServiceFromExternalListener(client client.Client, cluster *v1beta1.Kafka
 func (r *Reconciler) reorderBrokers(log logr.Logger, brokers []v1beta1.Broker) []v1beta1.Broker {
 	kClient, close, err := r.kafkaClientProvider.NewFromCluster(r.Client, r.KafkaCluster)
 	if err != nil {
-		log.Info("could not create Kafka client, thus could not determine controller")
+		log.Info(fmt.Sprintf("could not create Kafka client, thus could not determine controller. %s", err))
```
stoader, line 1040:
```suggestion
		log.Error(err, "could not create Kafka client, thus could not determine controller"))
```

pkg/resources/kafka/kafka.go
```diff
@@ -974,14 +1037,14 @@ func getServiceFromExternalListener(client client.Client, cluster *v1beta1.Kafka
...
 		return brokers
 	}
 	defer close()
 
 	_, controllerID, err := kClient.DescribeCluster()
 	if err != nil {
-		log.Info("could not find controller broker")
+		log.Info(fmt.Sprintf("could not find controller broker. %s", err))
```
stoader, line 1047:
```suggestion
		log.Error(err, "could not find controller broker"))
```

pkg/resources/kafka/configmap.go
```diff
@@ -63,49 +63,36 @@ func (r *Reconciler) getConfigProperties(bConfig *v1beta1.BrokerConfig, id int32
...
+		log.Error(err, "getting Kafka broker port for Cruise Control failed")
+	}
+	// Add Cruise Control SSL configuration
+	if util.IsSSLEnabledForInternalCommunication(r.KafkaCluster.Spec.ListenersConfig.InternalListeners) {
+		if !r.KafkaCluster.Spec.IsClientSSLSecretPresent() {
+			log.Error(errors.New("cruise control metrics reporter needs ssl but client certificate hasn't specified"), "")
 		}
-		if err := config.Set("ssl.truststore.location", serverKeystorePath+"/"+v1alpha1.TLSJKSTrustStore); err != nil {
-			log.Error(err, "setting ssl.truststore.location parameter in broker configuration resulted an error")
+		if err := config.Set("cruise.control.metrics.reporter.security.protocol", "SSL"); err != nil {
```
stoader, line 75: The same approach could be used as in https://github.com/banzaicloud/koperator/pull/698/files#diff-712b068c8dcc08d66a78aa61aedf8108f247eb9d3a6ee68d3d44047d1dfc5fe4R251

baluchicken:
Regression: Cert-manager generated secret now being used immediately which causing kafka broker errors since not all related field is present at the secret.

pkg/resources/cruisecontrol/deployment.go
bartam1: The code executing stops sooner with error if this case happen.

pkg/resources/kafka/configmap.go
bartam1: Yes, It can be replaced some places in the codebase. I think this is out of the scope.

bartam1:
> 

Fixed

pkg/resources/kafka/kafka.go
```diff
@@ -140,7 +140,7 @@ func getLoadBalancerIP(foundLBService *corev1.Service) (string, error) {
 }
 
 // Reconcile implements the reconcile logic for Kafka
-func (r *Reconciler) Reconcile(log logr.Logger) error {
+func (r *Reconciler) Reconcile(log logr.Logger) error { //nolint gocyclo
```
baluchicken, line 143: Please do not apply nolint gocyclo for this mandatory method, refactor it.

api/v1beta1/kafkacluster_types.go
```diff
@@ -441,11 +449,23 @@ type InternalListenerConfig struct {
...
+	// Secret has to contain the keystore and truststore in jks and the password for them in base64 encoded format.
+	// Data fields must be: keystore.jks, truststore.jks, password
+	// SecurityProtocol has to be set to ssl.
+	ServerSSLCertSecret *corev1.LocalObjectReference `json:"serverSSLCertSecret,omitempty"`
 	// +kubebuilder:validation:Pattern=^[a-z0-9\-]+
 	Name          string `json:"name"`
 	ContainerPort int32  `json:"containerPort"`
 }
 
+func (c CommonListenerSpec) GetServerSSLCertSecretName() string {
```
baluchicken, line 462: This should be *CommonListenerSpec do not use two separate receiver for this struct.

api/v1beta1/kafkacluster_types.go
```diff
@@ -537,6 +557,19 @@ func (iIConfig *IstioIngressConfig) GetReplicas() int32 {
 	return iIConfig.Replicas
 }
 
+// GetClientSSLCertSecretName returns the ClientSSLCertSecretName. It returns empty string if It's not specified
+func (k KafkaClusterSpec) GetClientSSLCertSecretName() string {
```
baluchicken, line 561: This should be *KafkaClusterSpec do not use two separate receiver type for KafkaClusterSpec

api/v1beta1/kafkacluster_types.go
```diff
@@ -537,6 +557,19 @@ func (iIConfig *IstioIngressConfig) GetReplicas() int32 {
...
+	if k.ClientSSLCertSecret == nil {
+		return ""
+	}
+	return k.ClientSSLCertSecret.Name
+}
+
+// IsClientSSLSecretPresent returns true if ssl client certification has been setted for the operator and cruise control.
+func (k KafkaClusterSpec) IsClientSSLSecretPresent() bool {
```
baluchicken, line 569: Same here

baluchicken (changes requested):

pkg/resources/kafka/kafka.go
```diff
@@ -438,41 +440,132 @@ func arePodsAlreadyDeleted(pods []corev1.Pod, log logr.Logger) bool {
...
 
-	superUsers := make([]string, 0)
-	for _, secret := range []*corev1.Secret{serverSecret, clientSecret} {
-		cert, err := certutil.DecodeCertificate(secret.Data[corev1.TLSCertKey])
-		if err != nil {
-			return "", "", nil, errors.WrapIfWithDetails(err, "failed to decode certificate")
+	return serverSecret, nil
+}
+
+func (r *Reconciler) getServerPasswordKeysAndUsers() (map[string]string, []string, error) {
```
adamantal, line 507: The implementation of this function seems a bit complex. Could you describe it and add some note here what this method actually does?

adamantal (commented):
Thanks for the work so far! The change looks really good and covers many edge cases.
I looked the PR only, and found these nits. 
Will probably attempt to test it too with Balint.

go.mod
```diff
@@ -10,6 +10,9 @@ require (
 	github.com/banzaicloud/k8s-objectmatcher v1.6.1
 	github.com/banzaicloud/koperator/api v0.0.0
 	github.com/banzaicloud/koperator/properties v0.1.0
+	github.com/banzaicloud/operator-tools v0.21.3 // indirect
```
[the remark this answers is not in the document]
bartam1: Im using go1.17 and I have ran go mod tidy

adamantal (commented):
The tests look good to me. One nit for the comment.

pkg/resources/kafka/kafka.go
```diff
@@ -505,26 +505,35 @@ func getListenerSSLCertSecret(client client.Reader, commonSpec v1beta1.CommonLis
...
+	// globKeyPass contains the generated ssl keystore password.
 	var globKeyPass string
 	var err error
 	serverSecret := &corev1.Secret{}
 	for _, iListener := range r.KafkaCluster.Spec.ListenersConfig.InternalListeners {
 		if iListener.Type == v1beta1.SecurityProtocolSSL {
+			// This implementation logic gets the generated ssl secret only once even
+			// if multiple listener use the generated one, because they share the same.
 			if globKeyPass == "" || iListener.GetServerSSLCertSecretName() != "" {
+				// get the apropiate secret the generated as default or the custom if its specified
```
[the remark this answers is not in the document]
bartam1: Thank you for the review.

stoader (approved):

baluchicken (dismissed):

adamantal:
Can you fix the last nit? I can approve the PR after that.

adamantal (approved):

stoader (approved):

PR #702, opened 2021-11-21
Upgrade dependencies

amuraru:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no|
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |

### What's in this PR?

Added a new `update-go-deps` Makefile target to allow upgrading deps moving forward.
Upgraded the dependencies.

### Why?

Keep the dependencies up to date

### Checklist

- [x] Implementation tested


api/go.mod
```diff
@@ -6,11 +6,11 @@ require (
 	emperror.dev/errors v0.8.0
 	github.com/banzaicloud/istio-client-go v0.0.11
 	github.com/imdario/mergo v0.3.12
-	github.com/jetstack/cert-manager v1.4.4
+	github.com/jetstack/cert-manager v1.6.1
```
adamantal, line 9: > ⚠️ Following their deprecation in version 1.5, the cert-manager APIVersions v1alpha2, v1alpha3, and v1beta1 are no longer served.

According to the cert-manager v1.6.0 release description, they've removed the support for these API versions. Could you check that koperator does not use these versions?

amuraru: Thanks for the note @adamantal - compile and tests pass but will do extra checks

amuraru: hi @adamantal - I checked and only v1 version is used throughout the operator code

adamantal: Thanks for checking it!

adamantal (approved):
LGTM. Thanks!

chrisgacsal (approved):
LGTM! :shipit:

Issue #533, opened 2021-01-11
Operator dead locks during rolling upgrade if another pod is killed


matt-christiansen-exa:
**Describe the bug**

I have seen this happen a few times if we have a relatively large cluster, say 24 nodes, and a rolling update is started on that cluster if any other pod down the line is also restarted that pod never gets replaced and the rolling update deadlocks. I can make it progress forward if I start killing the pod its waiting on, its like the reconcile loop happens replaces the missing pods which allows the updated pod to come on line then it progress forward just fine. 

I believe the main problem is it waits for all topics to not be under replicated as [here](https://github.com/banzaicloud/kafka-operator/blob/e39e4de1d4a91fc6ec337e3a573af10eba43c700/pkg/resources/kafka/kafka.go#L625) which I agree with. Likely we need to keep the reconciliation loop going, maybe starting the pods with the existing config while the rolling upgrade happens. I'm not entirely sure of the best possible resolution here. 

One possible solution I have thought of is maybe a timeout waiting for it to be in sync, allowing the loop to replace the missing pods and try the upgrade again. 

**Steps to reproduce the issue:**
Create a cluster, I think it would work with any size cluster but its easier to do it with larger clusters. Trigger a rolling upgrade (change a config property or the version of kafka) and once the CR is in that state delete one of the kafka brokers that's further down the path, like if you have 5 delete broker 4. At that point it should be stuck on waiting for what ever pod its on for all replicas to be in sync. Delete what ever broker its on (a bit of guess work) and watch it replace broker 4 as well as the one its currently upgrading. 

**Expected behavior**
That rolling upgrade doesn't block downstream pod restarts and can proceed 

**Screenshots**

**Additional context**
Add any other context about the problem like release numberm version, branch, etc.

amuraru:
@matt-christiansen-exa I think `rollingUpgradeConfig.failureThreshold` can help in your case  https://github.com/banzaicloud/kafka-operator/blob/1f68f0a0c05dc53cce9d140e427ea67a937c9dfc/config/base/crds/kafka.banzaicloud.io_kafkaclusters.yaml#L5710-L5718

It controls the max number of failures the rolling restart tolerates. If you increase that the operator should be able to skip a failing broker and continuing the procedure.

baluchicken:
Closing this since a fix just got merged. Please reopen it in case of reoccurrence.

PR #706, opened 2021-11-25
Handle with priority new brokers and missing broker pods reconciliation

stoader:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |
| Related to | #533 |

### What's in this PR?

Handle with priority the reconciliation of new brokers and missing broker pods in order to ensure that new brokers are added to the cluster and missing broker pods are re-created as soon as possible.

### Why?

Missing broker pods may result in offline partitions and not all replicas being in sync which in turn may case a rolling upgrade flow to stuck. By re-creating as soon as possible the missing broker pods the offline replicas error may disappear thus allowing the rolling upgrade to escape from the stuck state. In certain scenarios it may be needed to upscale the cluster to get it out from a bad state thus reconciling newly added brokers should have priority over others.

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/resources/kafka/kafka.go
+++ b/pkg/resources/kafka/kafka.go
@@ -218,18 +218,9 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
 		return errors.WrapIf(err, "failed to list broker pods that belong to Kafka cluster")
 	}
 
-	var controllerID int32
-	kClient, close, err := r.kafkaClientProvider.NewFromCluster(r.Client, r.KafkaCluster)
+	controllerID, err := r.determineControllerId()
 	if err != nil {
-		log.Error(err, "could not create Kafka client, thus could not determine controller")
-	} else {
-		_, controllerID, err = kClient.DescribeCluster()
-		if err != nil {
-			controllerID = -1
-			log.Error(err, "could not find controller broker")
-		}
-
-		defer close()
+		log.Error(err, "could not find controller broker")
 	}
 
 	reorderedBrokers := reorderBrokers(brokerPods, r.KafkaCluster.Spec.Brokers, r.KafkaCluster.Status.BrokersState, controllerID)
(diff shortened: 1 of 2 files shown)


pkg/resources/kafka/kafka.go
```diff
@@ -211,7 +211,28 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
...
+		log.Error(err, "could not create Kafka client, thus could not determine controller")
+	} else {
+		_, controllerID, err = kClient.DescribeCluster()
+		if err != nil {
+			controllerID = -1
+			log.Error(err, "could not find controller broker")
+		}
+
+		defer close()
+	}
```
baluchicken, line 233: Nit: Can we maybe factor this code out to a method e.g.: func (r *Reconciler) determineControllerId() int32

baluchicken (commented):
THX @stoader just a small nit:)

baluchicken (approved):

pkg/resources/kafka/kafka.go
```diff
@@ -942,6 +954,22 @@ func (r *Reconciler) getK8sAssignedNodeport(log logr.Logger, eListenerName strin
...
+	kClient, close, err := r.kafkaClientProvider.NewFromCluster(r.Client, r.KafkaCluster)
+	if err != nil {
+		return -1, errors.WrapIf(err, "could not create Kafka client, thus could not determine controller")
+	}
+	_, controllerID, err := kClient.DescribeCluster()
+	if err != nil {
+		return -1, errors.WrapIf(err, "could not find controller broker")
+	}
+
+	defer close()
```
adamantal, line 968: I'd move this up before the `kClient.DescribeCluster()`. In case the controller could not be determined, the client connections could leak.

pkg/resources/kafka/kafka.go
```diff
@@ -968,32 +996,50 @@ func getServiceFromExternalListener(client client.Client, cluster *v1beta1.Kafka
...
+	// 0 : non-controller broker
+	// 1 : controller broker
+	brokersWeight := make(map[string]int, len(desiredBrokers))
+
+	for _, b := range desiredBrokers {
+		brokerID := fmt.Sprintf("%d", b.Id)
+		brokersWeight[brokerID] = 0
+
+		if _, ok := brokersState[brokerID]; !ok {
+			brokersWeight[brokerID] = -2 // new broker
```
adamantal, line 1023: What do you think about creating an int enum to be more transparent about what a certain number means?

adamantal (commented):
Th core logic looks good to me. Two nits.

pkg/resources/kafka/kafka.go
baluchicken: That is a good catch:) Or maybe this method just returns an int32 and logs error instead of returning.

stoader: Nice catch.

adamantal (dismissed):
LGTM. Thanks!

baluchicken (approved):

adamantal (approved):

PR #708, opened 2021-11-26
Fix wrong JKS password has been generated for KafkaUser

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |

### What's in this PR?
Removes password override, because it clears the password too soon and the function returns with a wrong password.

### Why?
When KafkaUser custom resource has been created, koperator can create certs in JKS format into a secret. The koperator generates wrong password for that JKS.


stoader (changes requested):
Can you update the `TestGenerateJKS` unit test to catch the bug this PR fixes.

baluchicken (approved):
THX LGTM

hi-im-aren (approved):

PR #713, opened 2021-12-02
Use the number of distinct brokers with offline partition or out of sync replicas as error count during rolling upgrade

### LINKED ISSUE #695: Rolling upgrade failureThreshold not working as intended above the default value 1
**Describe the bug**
The current form of failureThreshold would have the same effect  any value above 1 (ex. there is no difference between failureThreshold value of 2 or 3). Furthermore, the errorCount, against which the failureThreshold is compared to, is always reseted to 0 and never written back to the status.

**Expected behavior**
Failure threshold is stating how many errors can the cluster tolerate during a rolling upgrade (ex. if guarding for a broker failure during a rolling upgrade).

stoader:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | yes |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets | fixes #695 |

### What's in this PR?

The number of failures during a rolling upgrade is is computed as the sum of distinct broker replicas with either offline replicas or out of sync replicas and the number of alerts triggered by alerts with 'rollingupgrade'

### Why?

The current implementation doesn't honor failureThreshold that is greater than 1, only when there is a 'rollingupgrade' alert firing that increases the error count during rolling upgrade.

### Checklist

- [x] Implementation tested

### DIFF
--- a/api/v1beta1/kafkacluster_types.go
+++ b/api/v1beta1/kafkacluster_types.go
@@ -104,7 +104,7 @@ type RollingUpgradeStatus struct {
 
 // RollingUpgradeConfig defines the desired config of the RollingUpgrade
 type RollingUpgradeConfig struct {
-	// FailureThreshold controls how many failures can the cluster tolerate during a rolling upgrade. Once the number of
+	// FailureThreshold controls how many failures the cluster can tolerate during a rolling upgrade. Once the number of
 	// failures reaches this threshold a rolling upgrade flow stops. The number of failures is computed as the sum of
 	// distinct broker replicas with either offline replicas or out of sync replicas and the number of alerts triggered by
 	// alerts with 'rollingupgrade'
--- a/api/v1beta1/kafkacluster_types.go
+++ b/api/v1beta1/kafkacluster_types.go
@@ -97,11 +97,17 @@ type KafkaClusterStatus struct {
 // RollingUpgradeStatus defines status of rolling upgrade
 type RollingUpgradeStatus struct {
 	LastSuccess string `json:"lastSuccess"`
-	ErrorCount  int    `json:"errorCount"`
+	// ErrorCount keeps track the number of errors reported by alerts labeled with 'rollingupgrade'.
+	// It's reset once these alerts stop firing.
+	ErrorCount int `json:"errorCount"`
 }
 
 // RollingUpgradeConfig defines the desired config of the RollingUpgrade
 type RollingUpgradeConfig struct {
+	// FailureThreshold controls how many failures can the cluster tolerate during a rolling upgrade. Once the number of
+	// failures reaches this threshold a rolling upgrade flow stops. The number of failures is computed as the sum of
+	// distinct broker replicas with either offline replicas or out of sync replicas and the number of alerts triggered by
+	// alerts with 'rollingupgrade'
 	FailureThreshold int `json:"failureThreshold"`
 }
 
--- a/charts/kafka-operator/templates/crds.yaml
+++ b/charts/kafka-operator/templates/crds.yaml
@@ -18553,12 +18553,12 @@ spec:
                   RollingUpgrade
                 properties:
                   failureThreshold:
-                    description: FailureThreshold controls how many failures can the
-                      cluster tolerate during a rolling upgrade. Once the number of
-                      failures reaches this threshold a rolling upgrade flow stops.
-                      The number of failures is computed as the sum of distinct broker
-                      replicas with either offline replicas or out of sync replicas
-                      and the number of alerts triggered by alerts with 'rollingupgrade'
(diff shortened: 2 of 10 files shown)


api/v1beta1/kafkacluster_types.go
```diff
@@ -97,11 +97,17 @@ type KafkaClusterStatus struct {
...
 	LastSuccess string `json:"lastSuccess"`
-	ErrorCount  int    `json:"errorCount"`
+	// ErrorCount keeps track the number of errors reported by alerts labeled with 'rollingupgrade'.
+	// It's reset once these alerts stop firing.
+	ErrorCount int `json:"errorCount"`
 }
 
 // RollingUpgradeConfig defines the desired config of the RollingUpgrade
 type RollingUpgradeConfig struct {
+	// FailureThreshold controls how many failures can the cluster tolerate during a rolling upgrade. Once the number of
```
amberimam, line 107:
```suggestion
	// FailureThreshold controls how many failures the cluster can tolerate during a rolling upgrade. Once the number of
```

charts/kafka-operator/templates/crds.yaml
```diff
@@ -18553,6 +18553,12 @@ spec:
                   RollingUpgrade
                 properties:
                   failureThreshold:
+                    description: FailureThreshold controls how many failures can the
+                      cluster tolerate during a rolling upgrade. Once the number of
```
amberimam, line 18557:
```suggestion
                    description: FailureThreshold controls how many failures the
                      cluster can tolerate during a rolling upgrade. Once the number of
```

config/base/crds/kafka.banzaicloud.io_kafkaclusters.yaml
```diff
@@ -18552,6 +18552,12 @@ spec:
                   RollingUpgrade
                 properties:
                   failureThreshold:
+                    description: FailureThreshold controls how many failures can the
+                      cluster tolerate during a rolling upgrade. Once the number of
```
amberimam, line 18556:
```suggestion
                    description: FailureThreshold controls how many failures the
                      cluster can tolerate during a rolling upgrade. Once the number of
```

config/samples/banzaicloud_v1beta1_kafkacluster.yaml
```diff
@@ -39,7 +39,10 @@ spec:
   # brokerConfigGroups specifies multiple broker configs with unique name
   #rollingUpgradeConfig specifies the rolling upgrade config for the cluster
   #rollingUpgradeConfig:
-  #failureThreshold states that how many errors can the cluster tolerate during rolling upgrade
+  #failureThreshold controls how many failures can the cluster tolerate during a rolling upgrade. Once the number of
```
amberimam, line 42:
```suggestion
  #failureThreshold controls how many failures the cluster can tolerate during a rolling upgrade. Once the number of
```

amberimam (changes requested):
Just a couple of nits.

pkg/resources/kafka/kafka.go
```diff
@@ -663,18 +663,31 @@ func (r *Reconciler) handleRollingUpgrade(log logr.Logger, desiredPod, currentPo
...
-				errorCount++
+			impactedReplicas := make(map[int32]struct{})
+			for _, brokerID := range allOfflineReplicas {
+				impactedReplicas[brokerID] = struct{}{}
+			}
+			for _, brokerID := range outOfSyncReplicas {
+				impactedReplicas[brokerID] = struct{}{}
 			}
+
+			errorCount += len(impactedReplicas)
```
adamantal, line 689: Since we're counting the individual failures, shouldn't this value be calculated as:
```
			errorCount = len(impactedReplicas)
```

Otherwise how do you make sure you doesn't count the same replica failure twice?

adamantal (commented):
LGTM. One question on the calculation of errors.

pkg/resources/kafka/kafka.go
stoader: `errorCount` is initialised with the value from `r.KafkaCluster.Status.RollingUpgrade.ErrorCount` (the error count set by rollingupgrade alerts) that we need to add `len(impactedReplicas)` to the initial value.

baluchicken (approved):
LGTM thx @stoader

pkg/resources/kafka/kafka.go
adamantal: Discussed offline. The ErrorCount stems from Prometheus, we just add our own counter there. Thanks for the clarification.

adamantal (approved):
Thanks for resolving this issue!

hi-im-aren (approved):

amberimam (approved):
Looks good!

PR #714, opened 2021-12-05
Allow setting concurency for envoy process when envoy ingress is used

amuraru:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

envoy allows setting the number of worker threads to run:

```
(optional) The number of worker threads to run. If not specified defaults to the number of hardware threads on the machine. If set to zero, Envoy will still run one worker thread.
```
https://www.envoyproxy.io/docs/envoy/latest/operations/cli#cmdoption-concurrency

Envoy though is not container aware hence when run in a container the "hardware threads" actually means
the total number of underlying k8s host. This creates high contention when the envoy pod requests/limits 
resources are set to a much lower value than host CPUs.

This patch allows the user to override this via envoyConfig.

### Why?

Reduce CPU contention in envoy pod

### Checklist

- [x] Implementation tested


api/v1beta1/kafkacluster_types.go
```diff
@@ -273,6 +274,10 @@ type EnvoyConfig struct {
 	HealthCheckPort *int32 `json:"healthCheckPort,omitempty"`
 	// DisruptionBudget is the pod disruption budget attached to Envoy Deployment(s)
 	DisruptionBudget *DisruptionBudgetWithStrategy `json:"disruptionBudget,omitempty"`
+	// key-value map of additional envoy configuration parameters
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +optional
+	AdditionalConfig map[string]string `json:"additionalConfig,omitempty"`
```
amuraru, line 280: I opted to add a free form `additionalConfig` bag for additional configurations as we have other envoy specific configurations we may add here. let me know if this sounds ok with you

stoader: Should this be rather called `Args` or `CommandLineArgs` to separate the configuration that is passed in via the config map from config passed in via command line args?

We could still have  `AddtionalConfig` for those additional configs which would go into the configmap.

api/v1beta1/kafkacluster_types.go
```diff
@@ -736,6 +741,23 @@ func (eConfig *EnvoyConfig) GetResources() *corev1.ResourceRequirements {
...
 }
 
+// GetConcurrency returns envoy concurrency.
+// Defines the number of worker threads envoy pod should run.
+// If not specified defaults to the number of hardware threads on the underlying kubernetes node.
+func (eConfig *EnvoyConfig) GetConcurrency() int {
+	if eConfig.AdditionalConfig == nil {
+		return 0
+	}
+	if value, ok := eConfig.AdditionalConfig["concurrency"]; ok {
```
stoader, line 751: Let's make `concurrency` a const and add a link to the documentation in the comment.

api/v1beta1/kafkacluster_types.go
amuraru: that's a good suggestion.
One thing though I am not sure about is if we should simply add all k-v pairs as arguments to envoy command line:
https://github.com/banzaicloud/koperator/pull/714/files#diff-821e43efb03a091583c0af4c4906d28f7186b6c7909d8f9dbbe0120f814635d4R69

or still handle them individually like concurrency.
not all https://www.envoyproxy.io/docs/envoy/latest/operations/cli#command-line-options may be valid when running in the operator

stoader: I would use this for all cli options except `-c`, `--config-path` which is set by koperator in order to be able to pass in the path to the config file it generated.

baluchicken (approved):

api/v1beta1/kafkacluster_types.go
```diff
@@ -273,6 +273,19 @@ type EnvoyConfig struct {
 	HealthCheckPort *int32 `json:"healthCheckPort,omitempty"`
 	// DisruptionBudget is the pod disruption budget attached to Envoy Deployment(s)
 	DisruptionBudget *DisruptionBudgetWithStrategy `json:"disruptionBudget,omitempty"`
+	// Envoy command line arguments
+	// +optional
+	// +kubebuilder:pruning:PreserveUnknownFields
+	CommandLineArgs *EnvoyCommandLineArgs `json:"additionalConfig,omitempty"`
```
stoader, line 279: Should the json field name renamed as well to `envoyCommanddLineArgs`

adamantal:
Hi @amuraru,
We want to roll out a `v0.20.1` soon and this PR can be included if you resolve Seba's last comment.

amuraru:
@adamantal - I address Seba's comments. Please have another look. thanks

stoader (approved):

adamantal (approved):
Thanks! LGTM

PR #717, opened 2021-12-06
Prevent creation of multiple KafkaTopics referencing the same Kafka topic

### LINKED ISSUE #715: Prevent creation of multiple KafkaTopics referencing the same Kafka topic
Currently there are no mechanisms to prevent creation of multiple `KafkaTopic`s referencing the same Kafka topic.

An admission webhook preventing the creation of another `KafkaTopic` having the same underlying Kafka topic.

This would require the client to keep in mind all their created `KafkaTopic` resource, which is tedious. No viable workaround.

**Additional context**
none

adamantal:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets | fixes #715 |

### What's in this PR?

Additional checks in the `KafkaTopic` admission webhook that prevents from multiple `KafkaTopic` CRs from being created referencing the same `KafkaCluster`'s same Kafka topic.

### Why?

This could create infinite cycles if two CRs are created with these details but with different configuration.

### Additional context

Tested by attempting to create topics with the properties above.

### Checklist

- [x] Implementation tested

### DIFF
--- a/main.go
+++ b/main.go
@@ -106,12 +106,15 @@ func main() {
 
 	ctrl.SetLogger(util.CreateLogger(verboseLogging, developmentLogging))
 
+	// adding indexers to KafkaTopics so that the KafkaTopic admission webhooks could work
+	ctx := context.TODO()
+	managerWatchCache := k8sutil.AddKafkaTopicIndexers(ctx)
+
 	//When operator is started to watch resources in a specific set of namespaces, we use the MultiNamespacedCacheBuilder cache.
 	//In this scenario, it is also suggested to restrict the provided authorization to this namespace by replacing the default
 	//ClusterRole and ClusterRoleBinding to Role and RoleBinding respectively
 	//For further information see the kubernetes documentation about
 	//Using [RBAC Authorization](https://kubernetes.io/docs/reference/access-authn-authz/rbac/).
-	managerWatchCache := (cache.NewCacheFunc)(nil)
 	var namespaceList []string
 	if namespaces != "" {
 		namespaceList = strings.Split(namespaces, ",")
@@ -119,11 +122,6 @@ func main() {
 			namespaceList[i] = strings.TrimSpace(namespaceList[i])
 		}
 		managerWatchCache = cache.MultiNamespacedCacheBuilder(namespaceList)
-	}
-	ctx := context.TODO()
-	if !webhookDisabled {
-		// we need to add indexers to KafkaTopics so that the KafkaTopic admission webhooks could work
-		managerWatchCache = k8sutil.AddKafkaTopicIndexers(ctx, managerWatchCache)
 	}
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
--- a/main.go
+++ b/main.go
@@ -30,6 +30,7 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"flag"
 	"os"
 	"strings"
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -118,6 +120,11 @@ func main() {
 		}
 		managerWatchCache = cache.MultiNamespacedCacheBuilder(namespaceList)
 	}
+	ctx := context.TODO()
+	if !webhookDisabled {
+		// we need to add indexers to KafkaTopics so that the KafkaTopic admission webhooks could work
+		managerWatchCache = k8sutil.AddKafkaTopicIndexers(ctx, managerWatchCache)
```
stoader, line 126: Would it worth adding the indexers always regardless webhooks are enabled or disabled?

pkg/webhook/topic_validator.go
```diff
@@ -121,9 +145,65 @@ func (s *webhookServer) validateKafkaTopic(topic *banzaicloudv1alpha1.KafkaTopic
...
-	// everything looks a-okay
-	return &admissionv1beta1.AdmissionResponse{
-		Allowed: true,
+// checkExistingKafkaTopicCRs checks whether there's any other duplicate KafkaTopic CR exists
+// that refers to the same KafkaCluster's same topic
+func (s *webhookServer) checkExistingKafkaTopicCRs(ctx context.Context,
+	clusterNamespace string, topic *banzaicloudv1alpha1.KafkaTopic) *admissionv1beta1.AdmissionResponse {
+	// check KafkaTopic in the referred KafkaCluster's namespace
+	kafkaTopicList := banzaicloudv1alpha1.KafkaTopicList{}
+	err := s.client.List(ctx, &kafkaTopicList,
```
stoader, line 157: Why can't just search for duplicate topics across all namespaces?

pkg/webhook/topic_validator.go
```diff
@@ -121,9 +145,65 @@ func (s *webhookServer) validateKafkaTopic(topic *banzaicloudv1alpha1.KafkaTopic
...
+		},
+		client.MatchingFieldsSelector{
+			Selector: fields.OneTermEqualSelector("spec.clusterRef.namespace", topic.Spec.ClusterRef.Namespace),
+		},
+	)
+	if err != nil {
+		log.Info("failed to list KafkaTopics")
+		return notAllowed("API failure while retrieving KafkaTopic list, please try again", metav1.StatusReasonServiceUnavailable)
+	}
+	if len(kafkaTopicList.Items) > 0 {
```
stoader, line 201: The list would contain the current `topic` so I think the check should rather be `if len(kafkaTopicList.Items) > 1 {`

stoader (changes requested):

main.go
adamantal: I wasn't sure. Since we don't the use the fieldselectors if the webhooks are disabled, I did not see it necessary. We can enable it regardless.

pkg/webhook/topic_validator.go
adamantal: We can merge it into one list query.

pkg/webhook/topic_validator.go
```diff
@@ -121,9 +145,47 @@ func (s *webhookServer) validateKafkaTopic(topic *banzaicloudv1alpha1.KafkaTopic
...
-		Allowed: true,
+	var foundKafkaTopic *banzaicloudv1alpha1.KafkaTopic
+	for i, kafkaTopic := range kafkaTopicList.Items {
+		// filter the cr under admission
+		if kafkaTopic.Name == topic.Name && kafkaTopic.Namespace == topic.Namespace {
+			continue
+		}
+
+		referredNamespace := kafkaTopic.Spec.ClusterRef.Namespace
+		if (kafkaTopic.Namespace == topic.Namespace && referredNamespace == "") || referredNamespace == clusterNamespace {
```
stoader, line 178:
```suggestion
		if (kafkaTopic.Namespace == clusterNamespace && referredNamespace == "") || referredNamespace == clusterNamespace {
```

stoader (changes requested):

stoader (approved):

baluchicken (approved):

PR #719, opened 2021-12-07
Fix wrong creation of JKS keystore and truststore for kafkauser

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |

### What's in this PR?
JKS assumes that private keys are PKCS8 encoded.
Koperator used PKCS1 for JKS. This produced an invalid jks format.
This fix corrects that. 

### Why?
Generated jks files were unusable

### DIFF
--- a/pkg/util/cert/certutil.go
+++ b/pkg/util/cert/certutil.go
@@ -90,6 +90,19 @@ var passChars []rune = []rune("ABCDEFGHIJKLMNOPQRSTUVWXYZ" +
 	"abcdefghijklmnopqrstuvwxyz" +
 	"0123456789")
 
+// DecodePKCS1PrivateKeyBytes will decode a PEM PKCS1 encoded private key into a rsa.PrivateKey format.
+func DecodePKCS1PrivateKeyBytes(keyBytes []byte) (*rsa.PrivateKey, error) {
+	block, _ := pem.Decode(keyBytes)
+	if block == nil {
+		return nil, errors.New("failed to decode PEM data")
+	}
+	key, err := x509.ParsePKCS1PrivateKey(block.Bytes)
+	if err != nil {
+		return nil, err
+	}
+	return key, nil
+}
+
 // DecodeKey will take a PEM encoded Private Key and convert to raw der bytes
 func DecodeKey(raw []byte) (parsedKey []byte, err error) {
 	block, _ := pem.Decode(raw)
@@ -97,6 +110,7 @@ func DecodeKey(raw []byte) (parsedKey []byte, err error) {
 		err = errors.New("failed to decode PEM data")
 		return
 	}
+
 	var keytype certv1.PrivateKeyEncoding
 	var key interface{}
 	if key, err = x509.ParsePKCS1PrivateKey(block.Bytes); err != nil {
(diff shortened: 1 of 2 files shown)


baluchicken (approved):

chrisgacsal (changes requested):
This should not go in in this form. I'll block it until we have discussed this offline.

pkg/util/cert/certutil.go
```diff
@@ -166,7 +149,12 @@ func GenerateJKSFromByte(certByte []byte, privateKey []byte, caCert []byte) (out
 
 // GenerateJKS creates a JKS with a random password from a client cert/key combination
 func GenerateJKS(certs []*x509.Certificate, privateKey []byte) (out, passw []byte, err error) {
-	pKeyRaw, err := DecodeKey(privateKey)
+	pKeyRaw, err := DecodePKCS1PrivateKeyBytes(privateKey)
+	if err != nil {
+		return
+	}
+
+	pKeyPKCS8, err := x509.MarshalPKCS8PrivateKey(pKeyRaw)
```
baluchicken, line 157: Is this thing fails if the format is other than pkcs1? I think we should keep using the old Decode function with this.

pkg/util/cert/certutil.go
```diff
@@ -90,30 +89,14 @@ var passChars []rune = []rune("ABCDEFGHIJKLMNOPQRSTUVWXYZ" +
 	"abcdefghijklmnopqrstuvwxyz" +
 	"0123456789")
 
-// DecodeKey will take a PEM encoded Private Key and convert to raw der bytes
-func DecodeKey(raw []byte) (parsedKey []byte, err error) {
-	block, _ := pem.Decode(raw)
+// DecodePKCS1PrivateKeyBytes will decode a PEM PKCS1 encoded private key into a rsa.PrivateKey format.
+func DecodePKCS1PrivateKeyBytes(keyBytes []byte) (*rsa.PrivateKey, error) {
```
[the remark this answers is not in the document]
baluchicken: The old function also has an error because it does not handle cert bundles which can be an issue.

baluchicken (changes requested):

pkg/util/cert/certutil.go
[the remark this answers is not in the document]
bartam1: koperator generates RSA priv key with encodePrivateKeyInPemFormat which is PKCS1 format. I think we shouldnt need to handle different format of private key if we generates only one possible format.

pkg/util/cert/certutil.go
bartam1: The prev function generates error x509.ParsePKCS1PrivateKey()  if it get other than pkcs1

pkg/util/cert/certutil.go
```diff
@@ -34,6 +35,12 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+var (
```
stoader, line 38:
```suggestion
const (
```

stoader (changes requested):

pkg/util/cert/certutil_test.go
```diff
@@ -35,50 +34,6 @@ type PKCS8Key struct {
...
-	}
-
-	var decoded []byte
-	if decoded, err = DecodeKey(key); err != nil {
-		t.Error("Expected to decode PKCS1 key, got error:", err)
-	}
-
-	// Convert to PKCS8 and test
-
-	var pkey PKCS8Key
```
stoader, line 52: Remove the `PKCS8Key` struct if not used.

stoader (approved):

stoader (approved):

chrisgacsal (approved):
LGTM! :shipit:

baluchicken (approved):

PR #721, opened 2021-12-13
Pass DNS fields to the CSR if specified in the KafkaUserSpec

adamantal:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets | - |

### What's in this PR?

A simple PR to pass DNS fields in the CSR if specified in the `KafkaUserSpec`.

### Why?

Otherwise the PKI manager would create a certificate that does not contain the DNS names as SAN.

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/pki/k8scsrpki/k8scsr_test.go
+++ b/pkg/pki/k8scsrpki/k8scsr_test.go
@@ -29,22 +29,26 @@ type mockClient struct {
 }
 
 func newMockCluster() *v1beta1.KafkaCluster {
-	cluster := &v1beta1.KafkaCluster{}
-	cluster.Name = "test"
-	cluster.Namespace = "test-namespace"
-	cluster.Spec = v1beta1.KafkaClusterSpec{}
-	cluster.Spec.ListenersConfig = v1beta1.ListenersConfig{}
-	cluster.Spec.ListenersConfig.InternalListeners = []v1beta1.InternalListenerConfig{
-		{
-			CommonListenerSpec: v1beta1.CommonListenerSpec{
-				ContainerPort: 9092,
+	return &v1beta1.KafkaCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test",
+			Namespace: "test-namespace",
+		},
+		Spec: v1beta1.KafkaClusterSpec{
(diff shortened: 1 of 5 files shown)


stoader (changes requested):
Can we add a unit test which verifies that the CSR is created with the correct content?

adamantal:
Tested manually and verified that the generated cert contains the string in its SAN field.

pkg/pki/k8scsrpki/k8scsr_test.go
```diff
@@ -0,0 +1,55 @@
...
+		{
+			CommonListenerSpec: v1beta1.CommonListenerSpec{
+				ContainerPort: 9092,
+			},
+		},
+	}
+	cluster.Spec.ListenersConfig.SSLSecrets = &v1beta1.SSLSecrets{
+		PKIBackend: v1beta1.PKIBackendK8sCSR,
+	}
+	return cluster
```
hi-im-aren, line 47:
```suggestion
	return &v1beta1.KafkaCluster{
		ObjectMeta: metav1.ObjectMeta{
			Name:      "test",
			Namespace: "test-namespace",
		},
		Spec: v1beta1.KafkaClusterSpec{
			ListenersConfig: v1beta1.ListenersConfig{
				InternalListeners: []v1beta1.InternalListenerConfig{
					{
						CommonListenerSpec: v1beta1.CommonListenerSpec{
							ContainerPort: 9092,
						},
					},
				},
				SSLSecrets: &v1beta1.SSLSecrets{
					PKIBackend: v1beta1.PKIBackendK8sCSR,
				},
			},
		},
	}
```

hi-im-aren (commented):
I wonder if gomega based assertions would be beneficial even if we skip ginkgo. It's already used in various parts of the project, might as well decide sooner or later which route we are taking.

adamantal:
> I wonder if gomega based assertions would be beneficial even if we skip ginkgo. It's already used in various parts of the project, might as well decide sooner or later which route we are taking instead of mix matching.

The tests use gomega now.

hi-im-aren (approved):

stoader (approved):

Issue #740, opened 2022-01-20
Cruisecontrol deployment does not get recreated if deleted accidentally by user.


baluchicken:
**Describe the bug**
Cruisecontrol deployment does not get recreated if deleted accidentally by user.

**Steps to reproduce the issue:**
Create a simpleKafkacluster
Delete the cruise control deployment

**Expected behaviour**
CruiseControl deployment should be recreated immediately.

Kuvesz:
Tried deleting the deployment and it got recreated so this issue was completely fixed by #749 .

PR #749, opened 2022-01-21
refactor: Cruise Control controller

### LINKED ISSUE #685: Operator might stuck if CruiseControl crashes during downscale
**Describe the bug**
If CruiseControl crashes during a downscale operation it might happen that the koperator stuck in a reconciling loop and will never reach a stable state.

There are no known instructions to reproduce this bug. The hypothetical steps that lead to this phenomenon are supposedly the following:
- Assume the cluster has 3+ brokers.
- Remove brokers from the cluster.
- The operator starts downscaling gracefully by submitting a downscale task to `CruiseControl`.
- The task finishes and the broker gets removed, but around that time the `CruiseControl` crashes.
- As CruiseControl comes back, it reports "missing task id" to the operator (as it had crashed and did not have a history of tasks), so the operator updates the status of the brokers to `GracefulDownscaleRequired` and attempts to submit the same downscale event again.
- The CruiseControl constantly returns a non-2xx return code and the following error:
```
Error processing POST request '/remove_broker' due to: 'com.linkedin.kafka.cruisecontrol.exception.KafkaCruiseControlException: java.lang.IllegalArgumentException: Broker 6 does not exist.'
```

**Expected behavior**
We drop the error response, remove the broker from `KafkaCluster` status field, and the cluster is stabilised with `ClusterRunning` state.

**Screenshots**
no

**Additional context**
The used Koperator version is `<0.11.0`.
This can be seen in the operator logs as the response payload from `CruiseControl`:
```
{"level":"info","ts":1634008654.4861798,"logger":"cruise-control-methods","msg":"Error processing POST request '/remove_broker' due to: 'com.linkedin.kafka.cruisecontrol.exception.KafkaCruiseControlException: java.lang.IllegalArgumentException: Broker 6 does not exist.'."}
{"level":"info","ts":1634008654.4862342,"logger":"controller.CruiseControlTask","msg":"cruise control communication error during downscaling broker(s)","clusterName":"kafka","clusterNamespace":"kafka","id(s)":["8","10","6","7","9"]}
```

chrisgacsal:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets | fixes #685, partially #740 |

### What's in this PR?
The Cruise Control controller got completely refactored by enhancing the reconcile logic and adopting the [go-cruise-control](https://github.com/banzaicloud/go-cruise-control) library for interacting with Cruise Control.

### Checklist

- [x] Implementation tested


controllers/controller_common.go
```diff
@@ -37,6 +37,12 @@ var clusterRefLabel = "kafkaCluster"
 // use as var so it can be overwritten from unit tests
 var newKafkaFromCluster = kafkaclient.NewFromCluster
 
+func requeueAfter(sec int) (ctrl.Result, error) {
```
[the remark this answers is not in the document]
chrisgacsal: The return error is needed as this function is used to end the reconcile flow and the controller runtime expects two values to be return where the first is a `ctrl.Result` and the second is an error which happens to be `nil` in this case.

So you can do this:
```
return requeueAfter(5)
```
instead of this:
```
return requeueAfter(5), nil
```
or as it was used previously
```
return ctrl.Result{
    RequeueAfter: time.Duration(5)*time.Second
}, nil
```

controllers/cruisecontroltask_controller.go
```diff
@@ -62,448 +59,412 @@ func (r *CruiseControlTaskReconciler) Reconcile(ctx context.Context, request ctr
...
-		err = k8sutil.UpdateBrokerStatus(r.Client, brokersWithTimedOutCCTask, kafkaCluster, brokersVolumesStateWithTimedOutDiskCCTask, log)
-		if err != nil {
-			return errors.WrapIfWithDetails(err, "could not update status for broker(s)", "id(s)", strings.Join(brokersWithTimedOutCCTask, ","))
+		for mountPath, volumeState := range brokerStatus.GracefulActionState.VolumeStates {
+			if volumeState.CruiseControlVolumeState.IsActionRequired() {
+				t := &CruiseControlTask{
+					TaskID:      volumeState.CruiseControlTaskId,
+					BrokerID:    brokerId,
+					StartedAt:   volumeState.TaskStarted,
+					Volume:      mountPath,
```
stoader, line 433: I wonder if `Volume` should be renamed to `KafkaLogDir` for example.

api/v1beta1/common_types.go
```diff
@@ -58,6 +58,14 @@ type PKIBackend string
 // CruiseControlVolumeState holds information about the state of volume rebalance
 type CruiseControlVolumeState string
 
+func (s CruiseControlVolumeState) IsRunningState() bool {
+	return s == GracefulDiskRebalanceRunning
+}
+
+func (s CruiseControlVolumeState) IsActionRequired() bool {
```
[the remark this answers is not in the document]
chrisgacsal: Fixed.

controllers/cruisecontroltask_controller.go
chrisgacsal: I agree that `KafkaLogDir` would be more appropriate however I would stick to the `Volume` as the state for it is described as `VolumeState`. The renaming would just make it more confusing.

api/v1beta1/common_types.go
```diff
@@ -58,22 +58,46 @@ type PKIBackend string
...
+func (s CruiseControlVolumeState) IsActive() bool {
+	return s == GracefulDiskRebalanceRunning || s == GracefulDiskRebalanceRequired
+}
+
+// IsUpscale returns true if CruiseControlState in GracefulUpscale* state.
 func (r CruiseControlState) IsUpscale() bool {
 	return r == GracefulUpscaleRequired || r == GracefulUpscaleSucceeded || r == GracefulUpscaleRunning
 }
 
+// IsUpscale returns true if CruiseControlState in GracefulDownscale* state.
```
baluchicken, line 78:
```suggestion
// IsDownscale returns true if CruiseControlState in GracefulDownscale* state.
```

chrisgacsal: Fixed.

controllers/cruisecontroltask_controller.go
```diff
@@ -62,448 +59,412 @@ func (r *CruiseControlTaskReconciler) Reconcile(ctx context.Context, request ctr
...
 			}
 		}
 	}
+}
 
-	if ccTaskId == "" {
-		return errors.New("no CC task id provided to be checked")
+// FromResult takes a scale.Result instance returned by scale.CruiseControlScaler and updates its own state accordingly.
+func (t *CruiseControlTask) FromResult(result *scale.Result) {
+	if result == nil {
```
stoader, line 310:
```
if t == nil || result == nil {
  return
```
just in case this method is invoked on a nil reference

controllers/cruisecontroltask_controller.go
```diff
@@ -62,448 +59,412 @@ func (r *CruiseControlTaskReconciler) Reconcile(ctx context.Context, request ctr
...
 	}
 
-	var brokersWithDownscaleRequired []string
-	var brokersWithUpscaleRequired []string
-	brokersWithDiskRebalanceRequired := make(map[string][]string)
+	switch {
+	case tasksAndStates.NumActiveTasksByOp(OperationAddBroker) > 0:
+		addBrokerTasks := make([]*CruiseControlTask, 0)
+		brokerIDs := make([]string, 0)
+		for _, task := range tasksAndStates.GetActiveTasksByOp(OperationAddBroker) {
```
stoader, line 104: Could be there any edge case when the list of broker IDs collected from `tasksAndStates.GetActiveTasksByOp` contains duplicate which later may cause an issue when  `scaler.AddBrokers` is invoked. If that's the case `bokerIDs` should be rather a `map[string]struct{}` where the key is broker id

chrisgacsal: There should not be any duplicated entries. Having duplicated entries indicates an issue in the controller logic which we want to surface as soon as possible. Also Cruise Control API is able to handle duplicated entries for broker IDs, so it does not trigger an error on Cruise Control end.

controllers/cruisecontroltask_controller.go
chrisgacsal: Nice catch. Fixed it.

adamantal (approved):
LGTM, two nits.
Thanks for working on this! Using the new CC client lib makes the code much nicer!

pkg/scale/utils.go
```diff
@@ -0,0 +1,80 @@
...
+func cruiseControlURL(endpoint string, secure bool) string {
+	scheme := "http"
+	if secure {
+		scheme = "https"
+	}
+	return fmt.Sprintf("%s://%s/%s", scheme, endpoint, "kafkacruisecontrol")
+}
+
+func stringSliceToMap(s []string) map[string]bool {
+	brokerIDsMap := make(map[string]bool, len(s))
```
stoader, line 67:
```suggestion
	brokerIDsMap := make(map[string]struct{}, len(s))
```

pkg/scale/utils.go
```diff
@@ -0,0 +1,80 @@
...
+	for _, id := range s {
+		brokerIDsMap[id] = true
```
stoader, line 69:
```suggestion
		brokerIDsMap[id] = struct{}{}
```

pkg/scale/utils.go
```diff
@@ -0,0 +1,80 @@
...
+		instance.Name,
+	)
+}
+
+func CruiseControlURL(namespace, domain, endpoint, name string) string {
+	var url string
+	if endpoint != "" {
+		url = endpoint
+	} else {
+		url = fmt.Sprintf("%s-cruisecontrol-svc.%s.svc.%s:8090", name, namespace, domain)
```
stoader, line 53: I think the https://github.com/banzaicloud/koperator/blob/master/pkg/resources/cruisecontrol/cruisecontrol.go#L38 should be used in order to ensure that the URL matches the K8s service name in case the service name format changes

See also https://github.com/banzaicloud/koperator/blob/master/pkg/resources/cruisecontrol/service.go#L31

controllers/cruisecontroltask_controller.go
```diff
@@ -62,448 +59,412 @@ func (r *CruiseControlTaskReconciler) Reconcile(ctx context.Context, request ctr
...
-		return errorfactory.New(errorfactory.CruiseControlTaskTimeout{}, errors.New("cc task timed out"), fmt.Sprintf("cc task id: %s", ccTaskId))
-	}
+const (
+	OperationAddBroker CruiseControlOperation = iota
+	OperationRemoveBroker
+	OperationRebalanceDisks
+)
+
+// CruiseControlTask defines a task to be performed via Cruise Control.
+type CruiseControlTask struct {
```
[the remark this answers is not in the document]
chrisgacsal: Ok. I'll move these to a separate `cruisecontrol_types.go` file.

pkg/resources/kafka/kafka.go
```diff
@@ -297,44 +297,75 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
...
 	}
 
-	deletedBrokers := make([]corev1.Pod, 0)
-OUTERLOOP:
+	brokerIDsFromSpec := make(map[string]bool, len(r.KafkaCluster.Spec.Brokers))
+	for _, broker := range r.KafkaCluster.Spec.Brokers {
+		brokerIDsFromSpec[strconv.Itoa(int(broker.Id))] = true
+	}
+
+	podsDeletedFromSpec := make([]corev1.Pod, 0, len(podList.Items))
```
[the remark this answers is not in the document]
chrisgacsal: The reason for using the `len(podList.Items)` for setting the capacity is that that is the maximum number of broker pods which can be removed from the cluster and to avoid unnecessary allocations. I personally don't see issue with the current implementation.

pkg/scale/utils.go
chrisgacsal: I prefer `map[string]bool` as it is more convenient to use. See also: https://gist.github.com/davecheney/3be245c92b61e5045f75

pkg/scale/utils.go
chrisgacsal: Same as above. It is a personal preference.

pkg/scale/utils.go
chrisgacsal: The current implementation represents the original logic for assembling the Cruise Control URL. I did not want to refactor this as well to avoid exploding the number of files/code changes in this PR.
I agree with your suggestion but it should be done in a separate PR. I'll create a PR for this after this one got merged.

controllers/cruisecontroltask_controller.go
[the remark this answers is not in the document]
chrisgacsal: Done.

pkg/scale/scale.go
```diff
@@ -15,497 +15,377 @@
...
+				brokersWithEmptyDisks = append(brokersWithEmptyDisks, brokerStat.Broker)
+			}
+		}
 	}
 
-	uResp, err := cc.postCruiseControl(addBrokerAction, options)
-	if err != nil && err != errCruiseControlNotReturned200 {
-		log.Error(err, "can't upscale cluster gracefully since post to cruise-control failed")
-		return "", "", err
+	if len(brokersWithEmptyDisks) == 0 {
```
stoader, line 276: When this method is called for a broker X with a single disk attached (log dir) which is empty (e.g. some custom message partitioning logic is used by producers which lead to message replicas to be stored only on the other brokers) a disk rebalance operation will triggered in CC unnecessarily since this condition won't be satisfied.

Could this cause a loop in the reconcile logic? or not as the triggered task is being tracked and once completed this method won't be invoked for broker X any more (unless a new disk is added again)

pkg/scale/utils.go
stoader: funny thing in the past I read that map[]struct{}{} performs better than map[]bool hence that's used inside apimachinery lib

internal/alertmanager/currentalert/process.go
```diff
@@ -305,24 +305,29 @@ func downScale(log logr.Logger, labels model.LabelSet, client client.Client) err
...
-		brokerId = string(broker)
+		brokerID = string(broker)
 	} else {
-		cc := scale.NewCruiseControlScaler(string(labels["namespace"]), cr.Spec.GetKubernetesClusterDomain(), cr.Spec.CruiseControlConfig.CruiseControlEndpoint, cr.Name)
-		brokerId, err = cc.GetBrokerIDWithLeastPartition()
+		cruiseControlURL := scale.CruiseControlURLFromKafkaCluster(cr)
+		// FIXME: we should reuse the context of passed to AController.Start() here
+		cc, err := scale.NewCruiseControlScaler(context.TODO(), cruiseControlURL)
+		if err != nil {
+			log.Error(err, "failed to initialize Cruise Control Scaler", "cruise control url", cruiseControlURL)
```
stoader, line 321: Since the error is passed back to callee no need to log the error `err` otherwise it may be logged twice (by the callee as well).

return errors.WrapIfWithDetails(err, "failed to initialize Cruise Control Scaler", "cruise control url", cruiseControlURL) should do the job

pkg/scale/scale.go
chrisgacsal: > When this method is called for a broker X with a single disk attached (log dir) which is empty (e.g. some custom message partitioning logic is used by producers which lead to message replicas to be stored only on the other brokers) a disk rebalance operation will triggered in CC unnecessarily since this condition won't be satisfied.

The code checks whether topic partition replicas are assigned to the LogDir/Disk or not. This is independent from what is the partition key used by Producers pushing messages to the topic. as Kafka assignes partition replicas to brokers (and LogDirs) when the Topic is created.

Disk rebalance is only triggered if there are one or more empty LogDirs (have no partition replicas assigned) for Broker(s) identified by `brokerIDs`.

> Could this cause a loop in the reconcile logic? or not as the triggered task is being tracked and once completed this method won't be invoked for broker X any more (unless a new disk is added again)

The latter. The task is tracked by the controller and it does not trigger a new disk rebalance action unless the task has gone from Cruise Control (it has been restarted since then).

I'll update the PR to add one tiny improvement to this logic which makes sure that we don't trigger disk rebalance for dead disks.

internal/alertmanager/currentalert/process.go
chrisgacsal: Agreed. Fixed.

pkg/scale/scale.go
chrisgacsal: Added the check for dead disks here: https://github.com/banzaicloud/koperator/pull/749/commits/71d4522248e505258c965fea06e2726fa7284e41

baluchicken (approved):

pkg/resources/kafka/kafka.go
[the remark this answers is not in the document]
adamantal: understandable

adamantal (approved):
Thanks for the fix!

stoader (approved):
LGTM. Thx for work.

Issue #747, opened 2022-01-21
Unable to create multiple kafka topics


stoader:
**Describe the bug**
Unable to create multiple kafka topics using `KafkaTopic` custom resources on K8s cluster version 1.21 with latest `koperator` built from master branch.

**Steps to reproduce the issue:**
Create two `KafkaTopic` custom resources with different topic names (e.g. topic1, topic2). The second fails as the validating webhook rejects it with `"KafkaTopic CR 'topic1' in namesapce 'kafka' is already referencing Kafka topic 'topic1'`

**Expected behavior**
Ability to create multiple kafka topics with different names via `KafkaTopic` custom resources.

priyavj08:
this is happening even on k8s 1.18.3 for me with latest koperator version v0.20.1 , I am seeing such errors in koperator logs

{"level":"info","ts":"2022-01-21T05:43:06.796Z","logger":"controllers.KafkaCluster","msg":"could not create cruise control topic: admission webhook \"kafkatopics.kafka.banzaicloud.io\" denied the request: KafkaTopic CR '**mme-events' in namesapce 'fed-kafka' is already referencing Kafka topic 'mme-events'**","Request.Namespace":"fed-kafka/kafka","Request.Name":"kafka"}

{"level":"error","ts":"2022-01-21T05:41:51.264Z","logger":"controller.kafkatopic","msg":"Reconciler error","name":"smf-events","namespace":"fed-kafka","error":"admission webhook \"kafkatopics.kafka.banzaicloud.io\" denied the request: KafkaTopic CR **'pgw-events' in namesapce 'fed-kafka' is already referencing Kafka topic 'pgw-events'"**,"stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.10.3/pkg/internal/controller/controller.go:227"}

adamantal:
Probably caused by #717

priyavj08:
@adamantal in which version was this regression introduced?  will v0.20.0 also have this issue?

adamantal:
The first release that contains this issue is `0.20.1`, so `0.20.0` does not have this defect. 
Since the error message is also introduced in this version, you must be in `>=0.20.1`.

baluchicken:
This one just got fixed. We also released a new version from the operator https://github.com/banzaicloud/koperator/releases/tag/v0.20.2 which contains the fix.

PR #750, opened 2022-01-21
Fix creation of multiple KafkaTopics

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets | #747  |

### What's in this PR?
It fixes a bug that happens when trying to add multiple KafkaTopic into the kafka cluster.

### Why?
Validating webhook rejects proper KafkaTopic resource.

- [x] Implementation tested

### DIFF
--- a/pkg/webhook/topic_validator.go
+++ b/pkg/webhook/topic_validator.go
@@ -155,7 +155,7 @@ func (s *webhookServer) checkExistingKafkaTopicCRs(ctx context.Context,
 	kafkaTopicList := banzaicloudv1alpha1.KafkaTopicList{}
 	err := s.client.List(ctx, &kafkaTopicList, client.MatchingFields{"spec.name": topic.Spec.Name})
 	if err != nil {
-		log.Info(fmt.Sprintf("Failed to list KafkaTopics. Error message: %s", err))
+		log.Error(err, "couldn't list KafkaTopic custom resources")
 		return notAllowed("API failure while retrieving KafkaTopic list, please try again", metav1.StatusReasonServiceUnavailable)
 	}
 
--- a/pkg/webhook/topic_validator.go
+++ b/pkg/webhook/topic_validator.go
@@ -21,7 +21,6 @@ import (
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -154,16 +153,9 @@ func (s *webhookServer) checkExistingKafkaTopicCRs(ctx context.Context,
 	clusterNamespace string, topic *banzaicloudv1alpha1.KafkaTopic) *admissionv1beta1.AdmissionResponse {
 	// check KafkaTopic in the referred KafkaCluster's namespace
 	kafkaTopicList := banzaicloudv1alpha1.KafkaTopicList{}
-	err := s.client.List(ctx, &kafkaTopicList,
-		client.MatchingFieldsSelector{
-			Selector: fields.OneTermEqualSelector("spec.name", topic.Spec.Name),
-		},
-		client.MatchingFieldsSelector{
-			Selector: fields.OneTermEqualSelector("spec.clusterRef.name", topic.Spec.ClusterRef.Name),
-		},
-	)
+	err := s.client.List(ctx, &kafkaTopicList, client.MatchingFields{"spec.name": topic.Spec.Name})
 	if err != nil {
-		log.Info("failed to list KafkaTopics")
+		log.Info(fmt.Sprintf("Failed to list KafkaTopics. Error message: %s", err))
 		return notAllowed("API failure while retrieving KafkaTopic list, please try again", metav1.StatusReasonServiceUnavailable)
 	}
 
@@ -175,9 +167,12 @@ func (s *webhookServer) checkExistingKafkaTopicCRs(ctx context.Context,
 		}
 
 		referredNamespace := kafkaTopic.Spec.ClusterRef.Namespace
-		if (kafkaTopic.GetNamespace() == clusterNamespace && referredNamespace == "") || referredNamespace == clusterNamespace {
-			foundKafkaTopic = &kafkaTopicList.Items[i]
-			break
+		referredName := kafkaTopic.Spec.ClusterRef.Name
+		if referredName == topic.Spec.ClusterRef.Name {
+			if (kafkaTopic.GetNamespace() == clusterNamespace && referredNamespace == "") || referredNamespace == clusterNamespace {
+				foundKafkaTopic = &kafkaTopicList.Items[i]
+				break
+			}
 		}
 	}
 	if foundKafkaTopic != nil {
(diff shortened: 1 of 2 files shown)


pkg/webhook/topic_validator.go
```diff
@@ -154,16 +153,9 @@ func (s *webhookServer) checkExistingKafkaTopicCRs(ctx context.Context,
...
 	kafkaTopicList := banzaicloudv1alpha1.KafkaTopicList{}
-	err := s.client.List(ctx, &kafkaTopicList,
-		client.MatchingFieldsSelector{
-			Selector: fields.OneTermEqualSelector("spec.name", topic.Spec.Name),
-		},
-		client.MatchingFieldsSelector{
-			Selector: fields.OneTermEqualSelector("spec.clusterRef.name", topic.Spec.ClusterRef.Name),
-		},
-	)
+	err := s.client.List(ctx, &kafkaTopicList, client.MatchingFields{"spec.name": topic.Spec.Name})
```
stoader, line 156: Can't we add the `spec.clusterRef.name` to the list of fields to match by?

e.g.:
```
client.MatchingFields{"spec.name": topic.Spec.Name, "spec.clusterRef.name", topic.Spec.ClusterRef.Name}
```

pkg/webhook/topic_validator.go
```diff
@@ -154,16 +153,9 @@ func (s *webhookServer) checkExistingKafkaTopicCRs(ctx context.Context,
...
 	if err != nil {
-		log.Info("failed to list KafkaTopics")
+		log.Info(fmt.Sprintf("Failed to list KafkaTopics. Error message: %s", err))
```
stoader, line 158:
```suggestion
		log.Error(err, "couldn't list KafkaTopic custom resources")
```

stoader (changes requested):

pkg/webhook/topic_validator.go
```diff
@@ -175,9 +167,12 @@ func (s *webhookServer) checkExistingKafkaTopicCRs(ctx context.Context,
 		}
 
 		referredNamespace := kafkaTopic.Spec.ClusterRef.Namespace
-		if (kafkaTopic.GetNamespace() == clusterNamespace && referredNamespace == "") || referredNamespace == clusterNamespace {
-			foundKafkaTopic = &kafkaTopicList.Items[i]
-			break
+		referredName := kafkaTopic.Spec.ClusterRef.Name
+		if referredName == topic.Spec.ClusterRef.Name {
```
stoader, line 171: This `if` statement will always evaluate to `true`

stoader (changes requested):

pkg/webhook/topic_validator.go
bartam1: No we cant :( This error message happen when I tried this solution: Failed to list KafkaTopics. Error message: non-exact field matches are not supported by the cache"}

pkg/webhook/topic_validator.go
bartam1: Why? Its kafkaTopic.Spec.ClusterRef.Name == topic.Spec.ClusterRef.Name

stoader: I see maybe we could change the name as it's quite easy to confuse.

pkg/webhook/topic_validator.go
hi-im-aren: For that to work we have to index the fields in the client cache e.g.
```
func addIndexerForKafkaTopic(ctx context.Context, newCache cache.Cache) error {
	nameIndexFunc := func(obj client.Object) []string {
		return []string{obj.(*v1alpha1.KafkaTopic).Spec.ClusterRef.Name}
	}
	err := newCache.IndexField(ctx, &v1alpha1.KafkaTopic{}, "spec.clusterRef.name", nameIndexFunc)
	if err != nil {
		return errors.WrapIfWithDetails(err, "could not setup indexer for field", "field", "spec.clusterRef.name")
	}

	namespaceIndexFunc := func(obj client.Object) []string {
		return []string{obj.(*v1alpha1.KafkaTopic).Spec.ClusterRef.Namespace}
	}
	err = newCache.IndexField(ctx, &v1alpha1.KafkaTopic{}, "spec.clusterRef.namespace", namespaceIndexFunc)
	if err != nil {
		return errors.WrapIfWithDetails(err, "could not setup indexer for field", "field", "spec.clusterRef.namespace")
	}
	return nil
}
```

bartam1: Multiple field selection is not supported. https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cache/internal/cache_reader.go#L119

stoader: @hi-im-aren I believe that's already used. The k8s `client` used by the topic validator is the same used by the controller-runtime: https://github.com/banzaicloud/koperator/blob/master/pkg/webhook/server.go#L63

https://github.com/banzaicloud/koperator/blob/master/main.go#L111

Not sure if this is related:
https://github.com/kubernetes-sigs/controller-runtime/blob/7ba3e559790c5e3543002a0d9670b4cef3ccf743/pkg/cache/internal/cache_reader.go#L192

https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cache/internal/cache_reader.go#L119
but this looks as though only matching by a single field is supported.

pkg/webhook/topic_validator.go
bartam1: I think it can clearly see because the location of the referredName

stoader, line 171: The conditions in these two `if` statement could be combined into a single expression:
```
if referredName == topic.Spec.ClusterRef.Name &&
  (kafkaTopic.GetNamespace() == clusterNamespace && referredNamespace == "" || 
  referredNamespace == clusterNamespace) {
```

stoader (changes requested):

pkg/webhook/topic_validator.go
adamantal, line 156: Why is the fieldselector expression got removed? Is it not working?

adamantal (changes requested):
For me it seems that this PR actually removes the client side filtering and use a Go slice based filtering, which essentially does not change the implementation.
I'd like to understand why the client side filtering does not work - and fix it, if it's possible.
Before you start working on this change I'd recommend getting through the original commit: https://github.com/banzaicloud/koperator/commit/dc5c2b9d81a20228a970ed18bc2656f415d13e8f
If we decide to completely remove the client side filtering which has been introduced in that commit, then we also have to:
- delete pkg/k8sutil/cache.go as the `AddKafkaTopicIndexers` is no longer needed
- remove the `AddKafkaTopicIndexers` call from main.go.
- rewrite the implementation that uses Go slice based filtering (already done in this PR)
We might as well just simply revert #717 for sake of simplicity, if it cannot be remediated.

I am open to your thoughts, let's discuss this.

pkg/webhook/topic_validator.go
Kuvesz: It seemed to be what caused the problem. Strangely enough if only the first selector was in it worked but not with both of them I think.

pkg/webhook/topic_validator.go
bartam1: Yes it can combined but for me it looks better to separate

pkg/webhook/topic_validator.go
stoader: Apparently the `client.List` takes into account only the first term from the matching field selector under the hood:
https://github.com/kubernetes-sigs/controller-runtime/blob/7ba3e559790c5e3543002a0d9670b4cef3ccf743/pkg/cache/internal/cache_reader.go#L195
https://github.com/kubernetes-sigs/controller-runtime/blob/7ba3e559790c5e3543002a0d9670b4cef3ccf743/pkg/cache/internal/cache_reader.go#L126

stoader (approved):

adamantal (approved):
We discussed this in more detail offline.
TLDR: the indexers are working, but the Kubernetes Go client can only take into account one of this option. It's ok to keep the indexer based solution, but can only use one condition.

baluchicken (approved):

Issue #572, opened 2021-03-11
Allow to set static PVC name for a broker


AndrewBoklashko:
Hi, thanks for the great Kafka operator. We've been using it on our dev environments for quite a long time and it's working well for us.

My new use case is to setup Kafka cluster in on-premise Kubernetes cluster without dynamic persistent volumes provisioning. So I am using preexisting [local volumes](https://kubernetes.io/docs/concepts/storage/volumes/#local) which I want to pre-bind to Kafka brokers PVCs using claimRef on PV. This way volumes will always be dedicated to Kafka brokers even if cluster needs to be rebuilt.

The problem is that operator always creates PVC with generated name. Considering that, can we allow to set PVC metadata, including name, in StorageConfigs? Note, I understand that I can set volumeName on PVC, however it will require me to release the volume manually if I need to rebuild Kafka cluster.

baluchicken:
Hi @AndrewBoklashko sorry for the delayed response. We are considering your proposal to allow setting PVC metadata. In the mean time there is a workaround to your problem. The operator using the client list method to find the required PVCs. If these PVCs are not present it will create new ones, but it can be tricked to reuse already created ones.
To do that all you have to do is add specific labels and annotations to your PVC.
```yaml
  annotations:
    mountPath: "<your mount path specified in your kafkacluster cr spec>"
  labels:
    brokerId: "<brokerId>"
    app: kafka
    kafka_cr: <your kafka cr name>
```
For example in case of the following CR:
```yaml
apiVersion: kafka.banzaicloud.io/v1beta1
kind: KafkaCluster
metadata:
  name: kafka
spec:
  brokerConfigGroups:
    default:
      storageConfigs:
        - mountPath: "/kafka-logs"
          pvcSpec:
            accessModes:
              - ReadWriteOnce
            resources:
              requests:
                storage: 10Gi
      brokers:
        - id: 0
           brokerConfigGroup: "default"
```
The following PVC will be reused by the operator without placing any ownerreference:
```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: kafka-0
  namespace: kafka
  annotations:
    mountPath: "/kafka-logs"
  labels:
    brokerId: "0"
    app: kafka
    kafka_cr: kafka
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 10Gi
```

AndrewBoklashko:
[...]

PR #751, opened 2022-01-23
Pvcglobaluniqeid

### LINKED ISSUE #744: launching Kafka components using StatefulSet insted of Deployments
Hello,
My name is Ely, I am a customer success engineer at Statehub, we provide a global PersistentVolumes allowing stateful applications on Kubernetes to access their PVs across clusters (regions and clouds) to access 

I’m currently creating a blog post about launching Kafka cluster using Koperator while leveraging our technology for a stateful Kafka cluster that will allow fast failover between one Kubernetes cluster to the other.

While testing your guide and launching a fully functional example Kafka cluster I noticed that all the components launched are using deployments and not StatefulSet, thus providing the PVC a name with a suffix that is auto-generated. 

Is it possible to deploy the Kafka and Zookeeper pods using a statefulSet with PersistentVolumeClaims instead of a deployment?
Providing a global name that will be recognized across clusters will allow us to launch a new Kafka cluster using the data and volumes from the previous cluster. in order for that to happen, we need zookeeper and Kafka volumes to have a unique and global ID and not be generated.
I browsed through the repository but couldn’t find an example of a config that will use  STS instead of deployment.

I’d appreciate your opinion,
Thanks in advance,
Ely

zaidenwerg:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no |
| New feature?    | yes |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets | fixes #744 |

### What's in this PR?

Created ObjectMetaWithNameAndAnnotations and ObjectMetaWithName to allow creation of objects with a deterministic name.
Use ObjectMetaWithNameAndAnnotations instead of ObjectMetaWithGeneratedNameAndAnnotations When creating a PVC.
Removed the hyphen at the end of brokerStorageTemplate since there isn't going to be a generated suffix.

### Why?

When using storage classes that span multiple Kubernetes clusters which use the PVC name to identify the volume that is referenced by the PersistentVolumeClaim or VolumeClaimTemplates, the PVC must have a deterministic name so it could be referenced on all the Kubernetes clusters that will reference it.

### Checklist

- [ ] Implementation tested

### To Do

- [ ] Testing

### Commentary 
Will be testing this code today/tomorrow. Would love your thoughts about it in the meanwhile.
Also, I make this change optional by adding a flag to StorageConfig:
```
type StorageConfig struct {
	MountPath string                                                     `json:"mountPath"`
        IsGlobal      bool                                                       `json:"isGlobal"`
	PvcSpec    *corev1.PersistentVolumeClaimSpec `json:"pvcSpec"`
}
```

### DIFF
--- a/pkg/resources/kafka/kafka.go
+++ b/pkg/resources/kafka/kafka.go
@@ -57,7 +57,8 @@ type brokerReconcilePriority int
 const (
 	componentName         = "kafka"
 	brokerConfigTemplate  = "%s-config"
-	brokerStorageTemplate = "%s-%d-storage-%d-"
+	brokerStorageTemplate = "%s-%d-storage-%d"
+
 
 	brokerConfigMapVolumeMount = "broker-config"
 	kafkaDataVolumeMount       = "kafka-data"
--- a/pkg/resources/kafka/pvc.go
+++ b/pkg/resources/kafka/pvc.go
@@ -29,7 +29,7 @@ import (
 
 func (r *Reconciler) pvc(brokerId int32, storageIndex int, storage v1beta1.StorageConfig, _ logr.Logger) runtime.Object {
 	return &corev1.PersistentVolumeClaim{
-		ObjectMeta: templates.ObjectMetaWithGeneratedNameAndAnnotations(
+		ObjectMeta: templates.ObjectMetaWithNameAndAnnotations(
 			fmt.Sprintf(brokerStorageTemplate, r.KafkaCluster.Name, brokerId, storageIndex),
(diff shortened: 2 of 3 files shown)


baluchicken:
Hi @zaidenwerg please look my answer here how can you create manually your own PV and PVC for the operator.
https://github.com/banzaicloud/koperator/issues/572#issuecomment-805668287

stoader:
@zaidenwerg could you describe the use case global PVC names would be needed across multiple Kubernetes clusters for Kafka?

PR #758, opened 2022-01-28
Allow to set rc.x to helm chart versions

baluchicken:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no|
| New feature?    |yes |
| API breaks?     | no|
| Deprecations?   | no|

### What's in this PR?

This PR allows to set rc.x versions to helm chart.
The chart/kafka-operator cannot be avoided yet, since the circle orb needs this, to work properly.
This should not break original functionality

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,4 +73,4 @@ workflows:
                       branches:
                           ignore: /.*/
                       tags:
-                          only: /chart\/kafka-operator\/\d+.\d+.\d+/
+                          only: /chart\/kafka-operator\/\d+.\d+.\d+(-rc\.\d+)?/


pregnor (approved):

.circleci/config.yml
```diff
@@ -73,4 +73,4 @@ workflows:
                       branches:
                           ignore: /.*/
                       tags:
-                          only: /chart\/kafka-operator\/\d+.\d+.\d+/
+                          only: /chart\/kafka-operator\/\d+.\d+.\d+(-rc\.\d+)?/
```
panyuenlau, line 76: With the change, will the helm chart release versions without `-rc.x` suffix work?

baluchicken: Yep this regex has an optional field in the end `(-rc\.\d+)?` it means this field can be omitted.

panyuenlau (approved):

PR #556, opened 2021-02-10
Add default broker capacity config for CruiseControl

adamantal:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -
| License         | Apache 2.0

### What's in this PR?

Adding a default broker capacity config to the CruiseControl `capacity.json` file.

### Why?

After ungraceful broker deletion the CruiseControl might not be aware of the brokers currently being deleted. Adding a default capacity entry to the capacity json enables CC to handle these brokers (though the actual value of resources does not matter).

### Checklist

- [x] Implementation tested
- [x] Error handling code meets the [guideline](https://github.com/banzaicloud/developer-guide/blob/master/docs/coding-style/error-handling-guide.md)
- [x] Logging code meets the guideline

PR #760, opened 2022-02-01
Removed default value from capacity.json, assigning default value for…

### LINKED ISSUE #741: Operator kills CC during broker addition even if adding broker is already in progress
**Describe the bug**
Operator kills CC during broker addition even if adding broker is already in progress.
This issue is caused by the default value inside the capacity.json. Before we added the default capacity value which can be used for all brokers, CC had to wait until the operator generated the new config for the new broker, otherwise the broker add task just cannot be called. Now since we have a default for all brokers CC can proceed with broker add using the default value. Once the operator restarts the CC with the new capacity config it just simply retry the add task.

Create a SimpleKafkaCluster
Wait until CC gets operational
Add an additional broker to the cluster
Check how CC behaves:
- will start to execute broker add task
- operator generates the new capacity config
- operator kills the cc to enable the new capacity config
- cc will retry broker add task

**Expected behaviour**
CC should wait with the task until the capacity.json contains all the necessary and accurate information for the given broker and proceed with the add task.

Kuvesz:
… each broker now.

| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets | fixes #741 |

### What's in this PR?
Removed default value from configmap and through that capacity.json. Now we're setting default values for every broker instead by utilizing the broker status as well as the spec. This way if something can be found in the status and not in the spec we know that that too needs (default) capacity config values assigned. This will basically mean that there will be no default config but there will be a default value for every broker which eliminates the cause of the mentioned issue.

### DIFF
--- a/controllers/kafkacluster_controller.go
+++ b/controllers/kafkacluster_controller.go
@@ -60,7 +60,6 @@ var clusterUsersFinalizer = "users.kafkaclusters.kafka.banzaicloud.io"
 type KafkaClusterReconciler struct {
 	client.Client
 	DirectClient        client.Reader
-	Log                 logr.Logger
 	Namespaces          []string
 	KafkaClientProvider kafkaclient.Provider
 }
@@ -85,7 +84,7 @@ type KafkaClusterReconciler struct {
 // +kubebuilder:rbac:groups=networking.istio.io,resources=*,verbs=*
 
 func (r *KafkaClusterReconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
-	log := r.Log.WithValues("Request.Namespace", request.NamespacedName, "Request.Name", request.Name)
+	log := logr.FromContextOrDiscard(ctx)
 
 	log.Info("Reconciling KafkaCluster")
 
@@ -104,7 +103,7 @@ func (r *KafkaClusterReconciler) Reconcile(ctx context.Context, request ctrl.Req
 
 	// Check if marked for deletion and run finalizers
 	if k8sutil.IsMarkedForDeletion(instance.ObjectMeta) {
-		return r.checkFinalizers(ctx, log, instance)
+		return r.checkFinalizers(ctx, instance)
 	}
 
 	if instance.Status.State != v1beta1.KafkaClusterRollingUpgrading {
@@ -192,7 +191,8 @@ func (r *KafkaClusterReconciler) Reconcile(ctx context.Context, request ctrl.Req
 	return reconciled()
 }
 
-func (r *KafkaClusterReconciler) checkFinalizers(ctx context.Context, log logr.Logger, cluster *v1beta1.KafkaCluster) (ctrl.Result, error) {
+func (r *KafkaClusterReconciler) checkFinalizers(ctx context.Context, cluster *v1beta1.KafkaCluster) (ctrl.Result, error) {
+	log := logr.FromContextOrDiscard(ctx)
 	log.Info("KafkaCluster is marked for deletion, checking for children")
 
 	// If the main finalizer is gone then we've already finished up
@@ -287,7 +287,7 @@ func (r *KafkaClusterReconciler) checkFinalizers(ctx context.Context, log logr.L
 		// Do any necessary PKI cleanup - a PKI backend should make sure any
 		// user finalizations are done before it does its final cleanup
 		log.Info("Tearing down any PKI resources for the kafkacluster")
-		if err = pki.GetPKIManager(r.Client, cluster, v1beta1.PKIBackendProvided, r.Log).FinalizePKI(ctx, log); err != nil {
+		if err = pki.GetPKIManager(r.Client, cluster, v1beta1.PKIBackendProvided).FinalizePKI(ctx); err != nil {
 			switch err.(type) {
 			case errorfactory.ResourceNotReady:
 				log.Info("The PKI is not ready to be torn down")
@@ -349,7 +349,8 @@ func (r *KafkaClusterReconciler) updateAndFetchLatest(ctx context.Context, clust
 }
 
 // SetupKafkaClusterWithManager registers kafka cluster controller to the manager
-func SetupKafkaClusterWithManager(mgr ctrl.Manager, log logr.Logger) *ctrl.Builder {
+func SetupKafkaClusterWithManager(mgr ctrl.Manager) *ctrl.Builder {
+	log := mgr.GetLogger()
 	builder := ctrl.NewControllerManagedBy(mgr).
 		For(&v1beta1.KafkaCluster{}).Named("KafkaCluster")
 
--- a/controllers/kafkatopic_controller.go
+++ b/controllers/kafkatopic_controller.go
@@ -92,7 +92,6 @@ type KafkaTopicReconciler struct {
 	// that reads objects from the cache and writes to the apiserver
 	Client client.Client
 	Scheme *runtime.Scheme
-	Log    logr.Logger
 }
 
 // +kubebuilder:rbac:groups=kafka.banzaicloud.io,resources=kafkatopics,verbs=get;list;watch;create;update;patch;delete;deletecollection
@@ -100,7 +99,7 @@ type KafkaTopicReconciler struct {
 
 // Reconcile reconciles the kafka topic
 func (r *KafkaTopicReconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
-	reqLogger := r.Log.WithValues("kafkatopic", request.NamespacedName, "Request.Name", request.Name)
+	reqLogger := logr.FromContextOrDiscard(ctx)
 	reqLogger.Info("Reconciling KafkaTopic")
 	var err error
 
@@ -141,7 +140,7 @@ func (r *KafkaTopicReconciler) Reconcile(ctx context.Context, request reconcile.
 
 	// Check if marked for deletion and if so run finalizers
 	if k8sutil.IsMarkedForDeletion(instance.ObjectMeta) {
-		return r.checkFinalizers(ctx, reqLogger, broker, instance)
+		return r.checkFinalizers(ctx, broker, instance)
 	}
 
 	// Check if the topic already exists
@@ -226,7 +225,8 @@ func (r *KafkaTopicReconciler) updateAndFetchLatest(ctx context.Context, topic *
 	return topic, nil
 }
 
-func (r *KafkaTopicReconciler) checkFinalizers(ctx context.Context, reqLogger logr.Logger, broker kafkaclient.KafkaClient, topic *v1alpha1.KafkaTopic) (reconcile.Result, error) {
+func (r *KafkaTopicReconciler) checkFinalizers(ctx context.Context, broker kafkaclient.KafkaClient, topic *v1alpha1.KafkaTopic) (reconcile.Result, error) {
+	reqLogger := logr.FromContextOrDiscard(ctx)
 	reqLogger.Info("Kafka topic is marked for deletion")
 	var err error
 	if util.StringSliceContains(topic.GetFinalizers(), topicFinalizer) {
--- a/controllers/kafkauser_controller.go
+++ b/controllers/kafkauser_controller.go
@@ -56,7 +56,8 @@ import (
 var userFinalizer = "finalizer.kafkausers.kafka.banzaicloud.io"
 
 // SetupKafkaUserWithManager registers KafkaUser controller to the manager
-func SetupKafkaUserWithManager(mgr ctrl.Manager, certSigningEnabled bool, certManagerNamespace bool, log logr.Logger) *ctrl.Builder {
+func SetupKafkaUserWithManager(mgr ctrl.Manager, certSigningEnabled bool, certManagerNamespace bool) *ctrl.Builder {
+	log := mgr.GetLogger()
 	builder := ctrl.NewControllerManagedBy(mgr).
 		For(&v1alpha1.KafkaUser{}).Named("KafkaUser")
 	if certSigningEnabled {
@@ -154,7 +155,6 @@ type KafkaUserReconciler struct {
 	// that reads objects from the cache and writes to the apiserver
 	Client client.Client
 	Scheme *runtime.Scheme
-	Log    logr.Logger
 }
 
 // +kubebuilder:rbac:groups=kafka.banzaicloud.io,resources=kafkausers,verbs=get;list;watch;create;update;patch;delete;deletecollection
@@ -167,7 +167,7 @@ type KafkaUserReconciler struct {
 // Reconcile reads that state of the cluster for a KafkaUser object and makes changes based on the state read
 // and what is in the KafkaUser.Spec
 func (r *KafkaUserReconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
-	reqLogger := r.Log.WithValues("kafkauser", request.NamespacedName, "Request.Name", request.Name)
+	reqLogger := logr.FromContextOrDiscard(ctx)
 	reqLogger.Info("Reconciling KafkaUser")
 	var err error
 
@@ -212,7 +212,7 @@ func (r *KafkaUserReconciler) Reconcile(ctx context.Context, request reconcile.R
 			backend = v1beta1.PKIBackendProvided
 		}
 
-		pkiManager := pki.GetPKIManager(r.Client, cluster, backend, r.Log)
+		pkiManager := pki.GetPKIManager(r.Client, cluster, backend)
 
 		user, err := pkiManager.ReconcileUserCertificate(ctx, instance, r.Scheme, cluster.Spec.GetKubernetesClusterDomain())
 		if err != nil {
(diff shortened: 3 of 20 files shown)


pkg/resources/cruisecontrol/configmap.go
```diff
@@ -162,30 +159,50 @@ func GenerateCapacityConfig(kafkaCluster *v1beta1.KafkaCluster, log logr.Logger,
...
-		brokerCapacity := BrokerCapacity{
-			BrokerID: fmt.Sprintf("%d", broker.Id),
-			Capacity: Capacity{
-				DISK:  brokerDisks,
-				CPU:   generateBrokerCPU(broker, kafkaCluster.Spec, log),
-				NWIN:  generateBrokerNetworkIn(broker, kafkaCluster.Spec, log),
-				NWOUT: generateBrokerNetworkOut(broker, kafkaCluster.Spec, log),
-			},
-			Doc: defaultDoc,
+	var strkeys []string
```
baluchicken, line 162: This can be pre initialized using make.
```suggestion
	strkeys := make([]string, 0, len(kafkaCluster.Status.BrokersState))
```

pkg/resources/cruisecontrol/configmap.go
```diff
@@ -162,30 +159,50 @@ func GenerateCapacityConfig(kafkaCluster *v1beta1.KafkaCluster, log logr.Logger,
...
+	for key := range kafkaCluster.Status.BrokersState {
+		strkeys = append(strkeys, key)
+	}
+	sort.Strings(strkeys)
+
+	for _, brokerId := range strkeys {
+		brokerFound := false
+		for _, broker := range kafkaCluster.Spec.Brokers {
+			if brokerId == fmt.Sprintf("%d", broker.Id) {
```
baluchicken, line 171: you can use `strconv.Itoa(int(broker.Id))` here to convert int32 to string.

pkg/resources/cruisecontrol/configmap.go
baluchicken: This variable should be called something like brokerIdsFromStatus

pkg/resources/cruisecontrol/configmap.go
```diff
@@ -162,30 +159,50 @@ func GenerateCapacityConfig(kafkaCluster *v1beta1.KafkaCluster, log logr.Logger,
...
+					Capacity: Capacity{
+						DISK:  brokerDisks,
+						CPU:   generateBrokerCPU(broker, kafkaCluster.Spec, log),
+						NWIN:  generateBrokerNetworkIn(broker, kafkaCluster.Spec, log),
+						NWOUT: generateBrokerNetworkOut(broker, kafkaCluster.Spec, log),
+					},
+					Doc: defaultDoc,
+				}
+
+				log.Info("The following brokerCapacity was generated", "brokerCapacity", brokerCapacity)
```
baluchicken, line 188: This line and line in 200 can be simplified and placed after line 203

pkg/resources/cruisecontrol/configmap.go
Kuvesz: Yes, I should rename it, will do that, but if I pre initilaize it than under that when appending stuff to it I would instead have to have an index and use that in the for cycle. This seemed easier and a lot cleaner to me really.

baluchicken: Yes you are right, that is why I used the 0 as the middle parameter for make. When you are specifying that you are safe to use append, since it will append all the data to the preallocated slice. If you omitting that you will end up with a slice where you have some (in case of a string slice) empty strings in the beginning.

pkg/resources/cruisecontrol/configmap.go
```diff
@@ -145,10 +146,6 @@ func GenerateCapacityConfig(kafkaCluster *v1beta1.KafkaCluster, log logr.Logger,
 	// If there is already a config added manually, use that one
 	if kafkaCluster.Spec.CruiseControlConfig.CapacityConfig != "" {
 		userProvidedCapacityConfig := kafkaCluster.Spec.CruiseControlConfig.CapacityConfig
-		updatedProvidedCapacityConfig := ensureContainsDefaultBrokerCapacity([]byte(userProvidedCapacityConfig), log)
```
stoader, line 148: Please ensure that there is config added for those brokers that the user missed to provide a CC config for.
E.g.: if there are 3 brokers and the user provided only CC capacity config for one broker than config is added for the remaining two brokers. Also add a unit test for this scenario.

pkg/resources/cruisecontrol/configmap.go
```diff
@@ -162,30 +159,47 @@ func GenerateCapacityConfig(kafkaCluster *v1beta1.KafkaCluster, log logr.Logger,
 
 	capacityConfig := CapacityConfig{}
 
-	for _, broker := range kafkaCluster.Spec.Brokers {
-		brokerDisks, err := generateBrokerDisks(broker, kafkaCluster.Spec, log)
-		if err != nil {
-			return "", errors.WrapIfWithDetails(err, "could not generate broker disks config for broker", "brokerID", broker.Id)
+	brokerIdFromStatus := make([]string, 0, len(kafkaCluster.Status.BrokersState))
+	for key := range kafkaCluster.Status.BrokersState {
```
stoader, line 163:
```suggestion
	for brokerId := range kafkaCluster.Status.BrokersState {
```

pkg/resources/cruisecontrol/configmap.go
```diff
@@ -162,30 +159,47 @@ func GenerateCapacityConfig(kafkaCluster *v1beta1.KafkaCluster, log logr.Logger,
...
+		brokerIdFromStatus = append(brokerIdFromStatus, key)
```
stoader, line 164:
```suggestion
		brokerIdFromStatus = append(brokerIdFromStatus, brokerId)
```

pkg/resources/cruisecontrol/configmap.go
```diff
@@ -162,30 +159,47 @@ func GenerateCapacityConfig(kafkaCluster *v1beta1.KafkaCluster, log logr.Logger,
...
+	}
+	sort.Strings(brokerIdFromStatus)
+	log.Info("String keys", "brokerIdFromStatus", brokerIdFromStatus)
```
stoader, line 167: I'd make this debug log and a bit more verbose as `String keys` doesn't say to much that would help debugging.
```suggestion
	log.V(1).Info("generating broker capacity config for brokers", "brokers", brokerIdFromStatus)
```

pkg/resources/cruisecontrol/configmap.go
```diff
@@ -162,30 +159,47 @@ func GenerateCapacityConfig(kafkaCluster *v1beta1.KafkaCluster, log logr.Logger,
...
+
+	for _, brokerId := range brokerIdFromStatus {
```
stoader, line 169: Instead of collecting the broker ids into a separate array wouldn't be more straightforward something like this?
```
for _, broker := range kafkaCluster.Spec.Brokers {
  if _, ok := kafkaCluster.Status.BrokersState[strconv.Itoa(int(broker.Id))]; !ok {
    brokerCapacity = generateDefaultBrokerCapacityWithId(brokerId)
    continue
  }

....
....
}

```

stoader (changes requested):

pkg/resources/cruisecontrol/configmap.go
Kuvesz: Oops, I forgot to remove that, it was only there for me to check if everything is okay with it.

pkg/resources/cruisecontrol/configmap.go
Kuvesz: It is ensured, if it's in kafkacluster.Status.BrokerState it will receive a CC config just not a default -1 indexed one but one specifically assigned to it. I'll create a unit test as you requested.

pkg/resources/cruisecontrol/configmap.go
Kuvesz: I don't think that would work right. We had the -1 indexed default config because of PR #556, where it's stated that it was added because of ungraceful broker deletions, or rather because CruiseControl might not be aware of those brokers and thus a default value was needed to deal with those. Now this caused some problems, which I'm aiming to solve here by getting rid of the -1 default config and adding a default to every broker based on state rather than spec, where they should be available even if they're not in spec. That is what the meaning of the brokerFound bool is, so that if they're found in spec we can assign them capcityconfig values from there but if they're not in there we can still assign them some default values, but on a id by id basis instead of a universal default.
What your way would achieve (if I understand correctly) is iterating through the spec and if we can't find a broker already found in the spec also in the status then we assign default values. I don't think that can happen really.

pkg/resources/cruisecontrol/configmap.go
baluchicken: It is not ensured right now, since when the user provides the capacity config, we will simply return in line 152. But I am not sure that is the right way to do, to silently appending additional configurations to this very important config. I know earlier we always injected the default value, but that can lead to not so precise CC plans. I am leaning towards a solution where we are checking if the user provided all the info required and if not just failing with an error.

stoader: That would be a change in the functionality and needs to be tracked in the release notes for those who upgrade and prior provided a cc config. 

I think that we can provide a good enough capacity values for the missing brokers since CPU and storage can be computed automatically. Only the NWIN and NWOUT would be a default value.

Kuvesz: Oh hell, you're right, sorry it is not ensured there. I misunderstood it a bit. What should happen here? I kinda understand the idea of Bálint with the error so that the user knows that the config they gave is not a complete one but at the same time won't it be annoying if for example all they care about is setting the config for one broker and they're fine with a default for the rest?

Kuvesz: We agreed offline to implement the following:
If the user adds capacity config, but not for every broker we shall add values automatically for every missing broker.
If the user adds capacity config, but with a -1 default value we shall use that, since the user technically specifically asked for their default and we shouldn't change something like that behind the scenes.
If the user adds capacity config for every broker then that's fine as well, that will be used.

Kuvesz: Okay, so hopefully the latest commit should fix this issue and do what I described in the previous comment.

pkg/resources/cruisecontrol/configmap.go
```diff
@@ -157,6 +161,21 @@ func GenerateCapacityConfig(kafkaCluster *v1beta1.KafkaCluster, log logr.Logger,
...
 
+	capacityConfig := CapacityConfig{}
+	BrokerCapacities, err := appendGeneratedBrokerCapacities(kafkaCluster, log, nil)
+	if err != nil {
+		return "", err
+	}
+	capacityConfig.BrokerCapacities = append(capacityConfig.BrokerCapacities, BrokerCapacities...)
+	result, err := json.MarshalIndent(capacityConfig, "", "    ")
+	if err != nil {
+		log.Error(err, "Could not marshal cruise control capacity config")
```
stoader, line 172: logging the error is redundant as the returned error will be logged.
```suggestion
		return "", err
```

pkg/resources/cruisecontrol/configmap.go
```diff
@@ -146,6 +146,10 @@ func GenerateCapacityConfig(kafkaCluster *v1beta1.KafkaCluster, log logr.Logger,
 	// If there is already a config added manually, use that one
 	if kafkaCluster.Spec.CruiseControlConfig.CapacityConfig != "" {
 		userProvidedCapacityConfig := kafkaCluster.Spec.CruiseControlConfig.CapacityConfig
+		updatedProvidedCapacityConfig, err := extendUserProvidedCapacityConfig([]byte(userProvidedCapacityConfig), kafkaCluster, log)
```
stoader, line 149: I think the logic would be neater and easier to read:
1. if user provides cc config than check for broker id "-1", if present return original capacity config
2. if broker id "-1" is not present than append the missing capacity config
3. if there is no user provided capacity config just append the missing capacity config to an empty config

pkg/resources/cruisecontrol/configmap.go
```diff
@@ -165,6 +184,20 @@ func GenerateCapacityConfig(kafkaCluster *v1beta1.KafkaCluster, log logr.Logger,
 	}
 	sort.Strings(brokerIdFromStatus)
 
+RemovePresentBrokerIds:
+	for _, presentBrokerId := range presentBrokerIds {
```
stoader, line 188: I think `util.StringSliceRemove` could be used or an enhanced version of it.

pkg/resources/cruisecontrol/configmap.go
```diff
@@ -173,7 +206,7 @@ func GenerateCapacityConfig(kafkaCluster *v1beta1.KafkaCluster, log logr.Logger,
 				brokerFoundInSpec = true
 				brokerDisks, err := generateBrokerDisks(broker, kafkaCluster.Spec, log)
 				if err != nil {
-					return "", errors.WrapIfWithDetails(err, "could not generate broker disks config for broker", "brokerID", broker.Id)
+					return []BrokerCapacity{}, errors.WrapIfWithDetails(err, "could not generate broker disks config for broker", "brokerID", broker.Id)
```
stoader, line 209:
```suggestion
					return nil, errors.WrapIfWithDetails(err, "could not generate broker disks config for broker", "brokerID", broker.Id)
```

stoader (changes requested):

pkg/resources/cruisecontrol/configmap.go
Kuvesz: Just checked, it does literally the same thing but does look a lot neater, thanks :)

pkg/resources/cruisecontrol/configmap.go
```diff
@@ -198,13 +231,47 @@ func GenerateCapacityConfig(kafkaCluster *v1beta1.KafkaCluster, log logr.Logger,
 
 		capacityConfig.BrokerCapacities = append(capacityConfig.BrokerCapacities, brokerCapacity)
 	}
+	return capacityConfig.BrokerCapacities, nil
+}
 
-	result, err := json.MarshalIndent(capacityConfig, "", "    ")
+func extendUserProvidedCapacityConfig(data []byte, kafkaCluster *v1beta1.KafkaCluster, log logr.Logger) (string, error) {
```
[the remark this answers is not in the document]
Kuvesz: I don't agree really with either of these, this function is only used when extending user provided config, hence the name. Sure I can use something more generic but what's the point until it's not needed elsewhere? Also I think the only thing that would change by passing in `*JBODInvariantCapacityConfig` instead of `[]byte` is that the unmarshal would have to happen before calling this function. I might be wrong on the second one though.

pkg/resources/cruisecontrol/configmap.go
Kuvesz: This is exactly what's happening? Or I may not understand what you mean. First we check if there's a user provided config, if there is we enter `extendUserProvidedCapacityConfig` where first we check if there is a "-1" ID present and if there is, we return, if there isn't we add generated values as we agreed previously. After all this we're on the "other side" of the if checking if the user provided their own config and we go on to call `appendGeneratedBrokerCapacities`, which adds generated values to an empty config basically.

pkg/resources/cruisecontrol/configmap.go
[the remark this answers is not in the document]
Kuvesz: Had an offline discussion about this, I understand now what you meant. I'll change it up a bit.

pkg/resources/cruisecontrol/configmap.go
```diff
@@ -160,80 +183,78 @@ func GenerateCapacityConfig(kafkaCluster *v1beta1.KafkaCluster, log logr.Logger,
...
 	}
-	return result
+	return brokerCapacities, nil
 }
 
-// generates default broker capacity
-// the exact values do not matter as for every broker it is set explicitly above
-// upon broker deletion CruiseControl does not use these value, it requires to detect that a broker exists
-func generateDefaultBrokerCapacity() BrokerCapacity {
+func generateDefaultBrokerCapacityWithId(brokerId string) BrokerCapacity {
```
stoader, line 255: Add comment to the function.

pkg/resources/cruisecontrol/configmap.go
```diff
@@ -160,80 +183,78 @@ func GenerateCapacityConfig(kafkaCluster *v1beta1.KafkaCluster, log logr.Logger,
...
+			}
+		}
+		// When removing a broker it still needs to have values assigned in capacity config
+		// although it doesn't really matter what the values are, so we are setting defaults
+		// here, this way we don't have to deal with a universal default.
+		if !brokerFoundInSpec {
+			log.Info("broker spec not found, using default fallback")
+			brokerCapacity = generateDefaultBrokerCapacityWithId(brokerId)
+		}
+		log.Info("the following brokerCapacity was generated", "brokerCapacity", brokerCapacity)
```
stoader, line 248:
```suggestion
		log.V(1).Info("capacity config successfully generated for broker", "capacity config", brokerCapacity)
```

pkg/resources/cruisecontrol/configmap.go
```diff
@@ -140,16 +142,37 @@ type JBODInvariantCapacityConfig struct {
...
+				return "", errors.WrapIfWithDetails(err,
+					"could retrieve broker Id from broker capacity configuration",
+					"capacity configuration", brokerCapacityMap)
+			}
+			if !ok {
+				continue
+			}
+			// If the -1 default exists we don't have to do anything else here since all brokers will have values.
+			if brokerId == "-1" {
+				return userProvidedCapacityConfig, nil
```
stoader, line 172: Log that using user provided capacity config e.g.:

```
log.Info("using user provided capacity config", "capacity config", userProvidedCapacityConfig)
```

stoader (changes requested):
LGTM, I've found only a few nits

baluchicken (approved):

stoader (approved):

PR #762, opened 2022-02-03
Fix sequence of actions when new volume is added

### LINKED ISSUE #738: CC Task initiated too fast in case of volume addition
**Describe the bug**
CC Task initiated too fast in case of volume addition. 
In case of a volume addition, the CC task initiated before the volume is bounded to the pod. It causes a false picture of a cluster since the user can think that the rebalance happened successfully but it is not. The rebalance indeed happened successfully but on a broker which is not aware of the new volume.

Create a KafkaCluster using [SimpleKafkaCluster](https://github.com/banzaicloud/koperator/blob/master/config/samples/simplekafkacluster.yaml)
Wait until CC is operational
Add an additional volume to the default brokerConfigGroup
Check that the GracefulDiskRebalanceSucceeded status populated before the pod even restarted to bound the new volume.

**Expected behavior**
The Rebalance task in CC should happen after the pvc is bound to the given pod.

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets | fixes #738 |

### What's in this PR?
It fixes the sequence of the actions when volume addition happens.

### Why?
CC Task initiated too fast in case of volume addition.
In case of a volume addition, the CC task is initiated before the volume is bounded to the pod. The rebalance indeed happened successfully but on a broker which is not aware of the new volume.

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/resources/kafka/kafka.go
+++ b/pkg/resources/kafka/kafka.go
@@ -884,6 +884,12 @@ func (r *Reconciler) reconcileKafkaPvc(log logr.Logger, brokersDesiredPvcs map[s
 				if mountPath == pvc.Annotations["mountPath"] {
 					currentPvc = pvc.DeepCopy()
 					alreadyCreated = true
+					// Checking pvc state, if bounded, so the broker has already restarted and the CC GracefulDiskRebalance has not happened yet,
+					// then we make it happening with status update.
+					if _, ok := r.KafkaCluster.Status.BrokersState[brokerId].GracefulActionState.VolumeStates[mountPath]; !ok &&
+						currentPvc.Status.Phase == corev1.ClaimBound {
+						brokerVolumesState[currentPvc.Annotations["mountPath"]] = v1beta1.VolumeState{CruiseControlVolumeState: v1beta1.GracefulDiskRebalanceRequired}
+					}
 					break
 				}
 			}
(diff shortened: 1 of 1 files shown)


pkg/resources/kafka/kafka.go
```diff
@@ -884,6 +884,12 @@ func (r *Reconciler) reconcileKafkaPvc(log logr.Logger, brokersDesiredPvcs map[s
 				if mountPath == pvc.Annotations["mountPath"] {
 					currentPvc = pvc.DeepCopy()
 					alreadyCreated = true
+					// Checking pvc state, if bounded, so the broker has already restarted and the CC GracefulDiskRebalance has not happened yet,
+					// then we make it happening with status update.
+					if _, ok := r.KafkaCluster.Status.BrokersState[brokerId].GracefulActionState.VolumeStates[mountPath]; !ok &&
+						currentPvc.Status.Phase == corev1.ClaimBound {
+						brokerVolumesState[currentPvc.Annotations["mountPath"]] = v1beta1.VolumeState{CruiseControlVolumeState: v1beta1.GracefulDiskRebalanceRequired}
```
stoader, line 891: Is the value of `currentPvc.Annotations["mountPath"]` the same as `mountPath`? if so could we use here `brokerVolumesState[mountPath] ` ?

bartam1: I have changed it to mountPath. Thank you!

stoader (approved):

stoader (commented):
Perhaps a comment could be added here: https://github.com/banzaicloud/koperator/blob/master/controllers/kafkacluster_controller.go#L397 mentioning that in case of updates for PersistentVolumeClaims changes to status should never be ignored as when the status of the PVC changed as reconcile must run.

baluchicken (approved):

Issue #769, opened 2022-02-11
Kafka PDB (PodDisruptionBudget) is invalid during scaling operations


alungu:
**Describe the bug**
The Kafka PodDisruptionBudget current implementation considers the number of brokers from the Kafka spec as being the total number of brokers. However, this is not true during scaling operation, resulting in an invalid PDB definition

**Steps to reproduce the issue:**

1. Create a KafkaCluster with 6 Kafka brokers and a DisruptionBudget of 1
 - Kafka's PDB `MIN AVAILABLE` is set to 5 (as expected); only 1 broker can be deleted
2. Scale out the KafkaCluster with 3 Kafka brokers
 - Kafka's PDB `MIN AVAILABLE` is updated to 2 as soon as the CR is applied; up to 4 brokers can be deleted
 - The scaling operation might take hours, depending on the cluster load (and time required to drain the nodes). During this time, the PDB is invalid, allowing more (than the DisruptionBudget) Kafka PODs to be deleted.

**Expected behavior**
If the CR sets a DisruptionBudget of 1, the Kafka's PDB `MIN AVAILABLE` should be N-1, where N is the number of *available* brokers.

**Additional context**
Proposal: Instead of using the number of brokers from the Spec (or from the Status), the proposal is to use the number of Kafka PODs available in the cluster: num(ListPods(KafkaCluster))

stoader:
@alungu I'm closing this as #770 which addresses this issue is has been merged now.

PR #770, opened 2022-02-11
Update Kafka PDB to use live Pods instead of the CR Spec

alungu:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets | [TBD](https://github.com/banzaicloud/koperator/issues/769) |

### What's in this PR?
The Kafka PDB logic is update to consider the total number of healthy Kafka brokers as being the number of Running Kafka pods (instead of the number of Kafka brokers specified in the Kafka CR)

### Why?
The previous implementation lead to a faulty PDB during scaling (both in and out) since adding or removing brokers requires some time.

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/resources/kafka/poddisruptionbudget.go
+++ b/pkg/resources/kafka/poddisruptionbudget.go
@@ -15,8 +15,12 @@
 package kafka
 
 import (
+	"context"
+	"emperror.dev/errors"
 	"fmt"
+	corev1 "k8s.io/api/core/v1"
 	"math"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"strconv"
 	"strings"
 
@@ -73,8 +77,16 @@ func (r *Reconciler) computeMinAvailable(log logr.Logger) (intstr.IntOrString, e
 		Max(1, brokers-brokers*percentage) - for a percentage budget
 
 	*/
+
+	podList := &corev1.PodList{}
+	matchingLabels := client.MatchingLabels(kafka.LabelsForKafka(r.KafkaCluster.Name))
+	err := r.Client.List(context.TODO(), podList, client.ListOption(client.InNamespace(r.KafkaCluster.Namespace)), client.ListOption(matchingLabels))
+	if err != nil {
+		return intstr.FromInt(-1), errors.WrapIf(err, "failed to reconcile resource")
+	}
+
 	// number of brokers in the KafkaCluster
-	brokers := len(r.KafkaCluster.Spec.Brokers)
+	brokers := len(podList.Items)
 
 	// configured budget in the KafkaCluster
 	disruptionBudget := r.KafkaCluster.Spec.DisruptionBudget.Budget


pkg/resources/kafka/poddisruptionbudget.go
```diff
@@ -73,8 +78,16 @@ func (r *Reconciler) computeMinAvailable(log logr.Logger) (intstr.IntOrString, e
 		Max(1, brokers-brokers*percentage) - for a percentage budget
 
 	*/
+
+	podList := &corev1.PodList{}
+	matchingLabels := client.MatchingLabels(kafka.LabelsForKafka(r.KafkaCluster.Name))
```
bartam1, line 83: I think with this label filter the cruise control pod also will be selected.

bartam1 (changes requested):
Thank you for the PR!

stoader:
@alungu relying on the number of Kafka pods might not be the best approach. For example if you have a 6 broker cluster and all 6 broker pods are up and running thus the PDB:min available will be 5. When a broker pod is deleted for whatever reason outside of `koperator` that event will trigger a reconcile in`koperator` which will update the  PDB:min available to 4 as there are only 5 running broker pods found before the missing pod is recreated. After the missing pod comes up in the next reconcile flow PDB:min available will be set back to 5.

If the PDB handling https://github.com/banzaicloud/koperator/blob/master/pkg/resources/kafka/kafka.go#L162 is moved as a last step after https://github.com/banzaicloud/koperator/blob/master/pkg/resources/kafka/kafka.go#L287 than it not might be an issue.

What I'm thinking of is to determine the number of brokers for PDB handling from Status as that contains the list of running brokers and brokers pending graceful upscale/downscale. In this case the PDB handling could stay where it is now:  https://github.com/banzaicloud/koperator/blob/master/pkg/resources/kafka/kafka.go#L162

What do you think?

alungu:
> @alungu relying on the number of Kafka pods might not be the best approach. For example if you have a 6 broker cluster and all 6 broker pods are up and running thus the PDB:min available will be 5. When a broker pod is deleted for whatever reason outside of `koperator` that event will trigger a reconcile in`koperator` which will update the PDB:min available to 4 as there are only 5 running broker pods found before the missing pod is recreated. After the missing pod comes up in the next reconcile flow PDB:min available will be set back to 5.
> 
> If the PDB handling https://github.com/banzaicloud/koperator/blob/master/pkg/resources/kafka/kafka.go#L162 is moved as a last step after https://github.com/banzaicloud/koperator/blob/master/pkg/resources/kafka/kafka.go#L287 than it not might be an issue.
> 
> What I'm thinking of is to determine the number of brokers for PDB handling from Status as that contains the list of running brokers and brokers pending graceful upscale/downscale. In this case the PDB handling could stay where it is now: https://github.com/banzaicloud/koperator/blob/master/pkg/resources/kafka/kafka.go#L162
> 
> What do you think?

@stoader Thank you for mentioning this issue!
I performed a few more tests with all the options (Spec vs Status vs Pod List). All of them have blind spots, but I agree that on the issue you highlighted with the "Pod List" option.
As such, I will patch this PR to use the KafkaCluster Status instead of the "List Kafka Pods".

pkg/resources/kafka/poddisruptionbudget.go
alungu: Hi @bartam1 and thank you for your review!
If we would consider only healthy / ready pods, we might end up allowing multiple pods removal. Here is the use-case:
1. We start a Kafka rolling restart (for a cluster with a disruption = 1)
2. 1 Pod is not starting (stuck in Pending or Init due to some issues)
3. The pod from (2) is not considered by the PDB update
4. We allow a disruption of another Pod. BUt with the one from (2) we now how 2 pods down.

It's the same issue as @stoader mentioned. However, if Sebastian's case is less likely to happen, this case described here is very likely. And that's the reason for which I considered (in my previous implementation) all the Pods, regardless of their state.

alungu:
@bartam1 @stoader could you review the new solution, please? Thanks!

stoader (approved):

pregnor (approved):

Issue #734, opened 2022-01-14
TLS termination at Ingress (Envoy) level


dobrerazvan:
**Is your feature request related to a problem? Please describe.**
Yes, there is no SSL termination on Ingress (envoy).

**Describe the solution you'd like to see**
We need a way to terminate ssl at Ingress (envoy). This includes:
- change from port based multiplexing in envoy, to name based multiplexing
- make envoy terminate ssl connections by adding TLSContext to envoy config

**Describe alternatives you've considered**
We considered using the already existing support for TLS in koperator, but in our deployments we surfaced some issues:
- certificates rotation requires a rolling restart of the kafka cluster (more expensive than an envoy deployment restart)
- during the transition to SSL, current setup requires 2 ports per broker (one plaintext and one ssl) and with large deployments cloud provider limits are reached

**Additional context**
Envoy config map sample -> https://gist.github.com/dobrerazvan/aa9ec700b165817319a464368abf2cf3
Kafka config map sample -> https://gist.github.com/dobrerazvan/6c3553425bac4b71149b9ddbb8142764

Adobe will provide this feature.

stoader:
@dobrerazvan if you terminate TLS at Ingress level that client certificates (presented by the client app) won't be passed to brokers which breaks client certificate based authentication. How can you ensure the the client certificates reach the brokers?

dobrerazvan:
@stoader SSL client authentication can be done with envoy, if needed. We are interested in a security solution for kafka, where local environment for kafka deployment is already secured and just outside data comes via unsecured channels.

stoader:
@dobrerazvan kafka acls mechanism reads the `subject` field of the client certificate to identify the client application in order to determine what topics it has access to. This is why the client certificate presented by the client application has to reach the broker as the broker reads the content of the certificate.

dobrerazvan:
Currently we do have the need to use the built-in ssl based authentication. We will implement a separate OAUTH based plugin in kafka that will take care of authentication. We are only interested in a way to bring traffic into the cluster in a secure way.

stoader:
Will you make the TLS termination at Ingress level configurable (like a passthrough mode etc) such that those users who rely on client certificate based authentication and authorization can still leverage this Kafka feature?

dobrerazvan:
[...]

PR #773, opened 2022-02-23
Enable TLS on the Envoy ingress

dobrerazvan:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no |
| New feature?    | yes |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets | fixes #X, partially #Y, mentioned in #Z |

### What's in this PR?
This PR implements the issue #734 .

### Checklist

- [x] Implementation tested

### DIFF
--- a/api/v1beta1/kafkacluster_types.go
+++ b/api/v1beta1/kafkacluster_types.go
@@ -15,6 +15,7 @@
 package v1beta1
 
 import (
+	"strconv"
 	"strings"
 
 	"emperror.dev/errors"
@@ -364,6 +365,16 @@ func (c ExternalListenerConfig) GetAnyCastPort() int32 {
 	return *c.AnyCastPort
 }
 
+// When TLS is enabled AnyCastPort is enough since hostname based multiplexing
+// is used and not port based one
+func (c ExternalListenerConfig) GetBrokerPort(brokerId int32) int32 {
+	if c.TLSEnabled() {
+		return c.GetAnyCastPort()
+	} else {
+		return c.ExternalStartingPort + brokerId
+	}
+}
+
 // GetServiceAnnotations returns a copy of the ServiceAnnotations field.
 func (c IngressServiceSettings) GetServiceAnnotations() map[string]string {
 	return util.CloneMap(c.ServiceAnnotations)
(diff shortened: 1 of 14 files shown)


stoader (changes requested):

api/v1beta1/kafkacluster_types.go
```diff
@@ -377,6 +388,16 @@ func (c IngressServiceSettings) GetServiceType() corev1.ServiceType {
 	return c.ServiceType
 }
 
+// Replace %id in brokerHostnameTemplate with actual broker id
+func (c IngressServiceSettings) GetBrokerHostname(brokerId int32) string {
+	return strings.Replace(c.BrokerHostnameTemplate, "%id", strconv.Itoa(int(brokerId)), 1)
+}
+
+// We use -1 for ExternalStartingPort value to enable TLS on envoy
+func (c ExternalListenerConfig) TLSEnabled() bool {
```
stoader, line 397: Determining whether TLS is enabled or not on an external listener based on the value of the `externalStartingPort` field IMHO is rather ambiguous than robust.

From the `KafkaCluster` CR it should be clear whether:
- TLS is enabled and provided by Kafka broker (currently this can be determined from the type of the external listener config - e.g.: ssl, sasl_ssl)
- TLS is enabled at the ingress controller (in this case whether TLS is enabled or not it should be determined from the specific ingress controller config)

Since having both TLS implemented at Kafka broker and ingress controller as well at the same time, doesn't seem to be a use case - at least for now - that should be supported there needs to be either:
- validation to not allow such config in KafkaCluster CR,
- or one of the two configs having priority over the other

api/v1beta1/kafkacluster_types.go
```diff
@@ -364,6 +365,16 @@ func (c ExternalListenerConfig) GetAnyCastPort() int32 {
 	return *c.AnyCastPort
 }
 
+// When TLS is enabled AnyCastPort is enough since hostname based multiplexing
+// is used and not port based one
+func (c ExternalListenerConfig) GetBrokerPort(brokerId int32) int32 {
+	if c.TLSEnabled() {
```
stoader, line 371: We should come up with a generic solution for generating port numbers for external listeners instead regardless wether TLS is enabled or not which allows the use of a single port.

pkg/resources/envoy/configmap.go
```diff
@@ -190,6 +193,89 @@ func generateEnvoyHealthCheckListener(ingressConfig v1beta1.IngressConfig, log l
...
+	brokerTcpProxyFilter := &envoylistener.Filter{
+		Name: wellknown.TCPProxy,
+		ConfigType: &envoylistener.Filter_TypedConfig{
+			TypedConfig: pbstTcpProxy,
+		},
+	}
+
+	filterChain := &envoylistener.FilterChain{
+		FilterChainMatch: &envoylistener.FilterChainMatch{
+			ServerNames:       []string{brokerFqdn},
```
stoader, line 241: Why do we need at all the broker address based multiplexing? If the max number of ports supported on LoadBalancer in an environment is too low that can be addressed via the use of multiple ingress controllers.

stoader (changes requested):

dobrerazvan:
Thanks @stoader for the review. I will look into it in the following days.

stoader:
> Thanks @stoader for the review. I will look into it in the following days.

@dobrerazvan what's the latest on this feature?


cc @pregnor @hi-im-aren @bartam1 @amberimam @panyuenlau @Kuvesz

dobrerazvan:
Too old code.

PR #774, opened 2022-02-24
Added support for Kafka 3.1.0

Kuvesz:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |

### What's in this PR?
Updated Kafka to 3.1.0 from 2.8.1.

### Why?
This is to keep up with Kafka and if there's something useful in the new releases now we can utilize it.

### DIFF
--- a/api/v1beta1/kafkacluster_types.go
+++ b/api/v1beta1/kafkacluster_types.go
@@ -627,7 +627,7 @@ func (kSpec *KafkaClusterSpec) GetClusterImage() string {
 	if kSpec.ClusterImage != "" {
 		return kSpec.ClusterImage
 	}
-	return "ghcr.io/banzaicloud/kafka:2.13-2.8.1"
+	return "ghcr.io/banzaicloud/kafka:2.13-3.1.0"
 }
 
 // GetClusterMetricsReporterImage returns the default container image for Kafka Cluster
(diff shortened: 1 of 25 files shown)


hi-im-aren:
Can you add the content table to the description of the PR? Otherwise Mergeable will hang.

api/v1beta1/kafkacluster_types.go
```diff
@@ -627,7 +627,7 @@ func (kSpec *KafkaClusterSpec) GetClusterImage() string {
 	if kSpec.ClusterImage != "" {
 		return kSpec.ClusterImage
 	}
-	return "ghcr.io/banzaicloud/kafka:2.13-2.8.1"
+	return "ghcr.io/banzaicloud/kafka:2.13-3.1.0"
```
panyuenlau, line 630: I am wondering if we should make this default container image a variable so it can be used in the common_test.go, instead of having this as the same string in multiple places

Kuvesz: We could, but does it matter really? We would be using it in about 2 places and changing the version is just a replace all basically here.

stoader (approved):

api/v1beta1/kafkacluster_types.go
panyuenlau: "`but does it matter really?`"

Well, I think it's always better to have the same long strings refer to the common variable so we can avoid mistakes like possible typos (maybe this is just my naive opinion)

But yea, it has nothing to do with what we aim to do in the PR, just noticed it while reviewing the changes and thought we might take the chance to make our codebase a little more robust

panyuenlau (approved):

PR #779, opened 2022-03-21
Fix delete service when headlessServiceEnabled changed

### LINKED ISSUE #532: svc was not Delete
**Describe the bug**
`headlessServiceEnabled` initially is true, Then I  set `headlessServiceEnabled: false`, and found that the same number of `svc` as the broker was created, which was in line with expectations, but when I changed to `headlessServiceEnabled: true`, the corresponding `svc` was not Delete.

Please describe the steps to reproduce the issue.
- 1、set `headlessServiceEnabled：true `, apply YAML file.
- 2、set `headlessServiceEnabled：false`, apply YAML file.
- 3、set `headlessServiceEnabled：true `, apply YAML file.

image

**Expected behavior**
  `svc` is deleted.

**Screenshots**

**Additional context**
 version: 0.14.0

bincherry:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets | fixes #532 |

### What's in this PR?

Delete kafka service if not needed.

### Checklist

- [x] Implementation tested


pkg/resources/kafka/kafka.go
```diff
@@ -1223,3 +1227,53 @@ func generateServicePortForEListeners(listeners []v1beta1.ExternalListenerConfig
...
+				util.MergeLabels(
+					kafka.LabelsForKafka(r.KafkaCluster.Name),
+					map[string]string{"brokerId": fmt.Sprintf("%d", broker.Id)},
+				),
+				r.KafkaCluster)})
+			if err != nil {
+				if apierrors.IsNotFound(err) {
+					log.V(1).Info("broker service not found. Continue")
+				} else {
+					return errors.WrapIfWithDetails(err, "could not delete broker service", "id", broker.Id)
```
panyuenlau, line 1258: Instead of returning immediately on error while deleting the Kafka broker service one by one, it might be more reasonable to cache the error and move on to deleting the next Kafka broker service in the slice. And we will return the error (if any) after finishing processing all the Kafka broker services. What do you think @stoader ?

panyuenlau (commented):
Thanks for providing a fix on this @bincherry! The overall implementation LGTM but I'd love to hear about others' opinions

pkg/resources/kafka/kafka.go
```diff
@@ -1223,3 +1227,53 @@ func generateServicePortForEListeners(listeners []v1beta1.ExternalListenerConfig
 	}
 	return usedPorts
 }
+
+func (r *Reconciler) reconcileHeadlessService(log logr.Logger) error {
```
bartam1, line 1231: I would divide this part into two function like: removeServices, removeHeadlessService and use it here: https://github.com/banzaicloud/koperator/blob/815d07e9ecc68defb6c966ba1c838b08fec6d994/pkg/resources/kafka/kafka.go#L153 and here: https://github.com/banzaicloud/koperator/blob/815d07e9ecc68defb6c966ba1c838b08fec6d994/pkg/resources/kafka/kafka.go#L159
What do you think?

pkg/resources/kafka/kafka.go
bartam1: It is ok for me.

bartam1: I think fast fail is ok here. If it cannot delete a service it means we cannot reach API server properly probably the next deletion would fail too. Return error will make a requeue and it will try the deletion again after some time.

pkg/resources/cruisecontrol/deployment.go
```diff
@@ -193,6 +193,7 @@ func GeneratePodAnnotations(kafkaCluster *v1beta1.KafkaCluster, capacityConfig s
 			"cruiseControlConfig.json":        hex.EncodeToString(hashedCruiseControlConfigJson[:]),
 			"cruiseControlClusterConfig.json": hex.EncodeToString(hashedCruiseControlClusterConfigJson[:]),
 			"cruiseControlLogConfig.json":     hex.EncodeToString(hashedCruiseControlLogConfigJson[:]),
+			"kafkaHeadlessServiceEnabled":     fmt.Sprintf("%t", kafkaCluster.Spec.HeadlessServiceEnabled),
```
stoader, line 196: What's the purpose of this annotation?

pkg/resources/kafka/kafka.go
```diff
@@ -290,6 +290,10 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
 		return err
 	}
 
+	if err = r.reconcileHeadlessService(log); err != nil {
```
stoader, line 293: It's quite confusing to have  `r.headlessService()` invoked at the beginning and then`r.reconcileHeadlessService()`. The handling of the headless service (deletion, creation, update) should be done in a single place `r.headlessService()`

stoader (changes requested):

stoader (commented):
In my opinion the proper way to implement this is to modify https://github.com/banzaicloud/koperator/blob/master/pkg/resources/kafka/headlessService.go#L29 to receive a parameter whether to delete or not the service.

Than it would be invoked from `kafka.go` :

```
o := r.headlessService(r.KafkaCluster.Spec.HeadlessServiceEnabled)
		err := k8sutil.Reconcile(log, r.Client, o, r.KafkaCluster)
		if err != nil {
			return errors.WrapIfWithDetails(err, "failed to reconcile resource", "resource", o.GetObjectKind().GroupVersionKind())
		}
```

pkg/resources/kafka/kafka.go
bincherry: Sorry for the late.

But in the reconsiling the old kafka services are still in use, so we should not delete it until the end of reconciling.

pkg/resources/cruisecontrol/deployment.go
bincherry: I found that cruisecontrol should restart after the service changing, otherwise it will runs into error.

pkg/resources/kafka/kafka.go
bincherry: Sorry for the late.

But in the reconsiling the old kafka services are still in use, so we should not delete it until the end of reconciling.

pkg/resources/kafka/kafka.go
stoader: you are right. Meanwhile I worked on this as I has some spare time to be able to get this fix in, please see: https://github.com/banzaicloud/koperator/pull/799

stoader:
Closing as this issue has been covered in #799

PR #793, opened 2022-04-13
Fix kafka broker pod label selector used when reconciling broker pod deletion

### LINKED ISSUE #792: Downscale impacts other cluster's brokers
**Describe the bug**
When there are multiple Kafka clusters in the same name and one of them is downscaled koperator restarts the broker pod with same id as the downscaled one of the other cluster.

Create two separate Kafka clusters cluster1 and cluster2 with 3 brokers in the same namespace. Downscale cluster1 by deleting broker with id 2 from `cluster1`. koperator will terminate broker pod with broker id 2 that belongs to `cluster1` as expected, however it also restarts broker pod with broker id 2 that belongs to `cluster2`

**Expected behavior**
koperator should not restart the broker that belongs to the other cluster.

stoader:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets | fixes #792 |

### What's in this PR?

Remove the `client.HasLabels` from the used label selector to find kafka broker pods that belong to cluster during broker pod deletion reconciliation.
The `client.MatchingLabels(kafka.LabelsForKafka(r.KafkaCluster.Name))` label selector is enough to match broker pods that belong to a cluster in a given namespace.

### Why?

`controller-runtime` doesn't support multiple label selector option to be specified for selecting pods. When multiple label selector option is used it doesn't return an error but rather silently uses only the last label selector option from the list. This resulted to select pods in the given namespace that have `brokerId` label regardless what Kafka cluster the pods belong to as the first option `client.MatchingLabels(kafka.LabelsForKafka(r.KafkaCluster.Name))` was ignored. 

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/resources/kafka/kafka.go
+++ b/pkg/resources/kafka/kafka.go
@@ -297,11 +297,9 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
 
 func (r *Reconciler) reconcileKafkaPodDelete(log logr.Logger) error {
 	podList := &corev1.PodList{}
-
 	err := r.Client.List(context.TODO(), podList,
-		client.ListOption(client.InNamespace(r.KafkaCluster.Namespace)),
-		client.ListOption(client.MatchingLabels(kafka.LabelsForKafka(r.KafkaCluster.Name))),
-		client.ListOption(client.HasLabels{"brokerId"}),
+		client.InNamespace(r.KafkaCluster.Namespace),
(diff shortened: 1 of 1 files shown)


bartam1 (approved):
LGTM.

panyuenlau (approved):
thanks for catching && fixing this

PR #799, opened 2022-04-22
Delete old kafka service when headlessServiceEnabled changed

### LINKED ISSUE #532: svc was not Delete

stoader:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets | fixes #532 |

### What's in this PR?

Delete old kafka broker service when the `headlessServiceEnabled` is changed. 

### Why?

When `headlessServiceEnabled` is changed new Kubernetes service is created and the old one is not deleted even though is not needed any more.

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/resources/kafka/allBrokerService.go
+++ b/pkg/resources/kafka/allBrokerService.go
@@ -15,7 +15,16 @@
 package kafka
 
 import (
+	"context"
 	"fmt"
+
+	"emperror.dev/errors"
+
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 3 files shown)


Kuvesz (dismissed):
LGTM but would love if someone more experienced took a look at this. :)

pkg/resources/kafka/allBrokerService.go
```diff
@@ -36,15 +45,79 @@ func (r *Reconciler) allBrokerService() runtime.Object {
...
+	)
+
+	if err != nil {
+		return errors.WrapIfWithDetails(err, "failed to list services",
+			"namespace", r.KafkaCluster.GetNamespace(),
+			"label selector", labelSelector.String())
+	}
+
+	for i := range services.Items {
+		svc := services.Items[i]
```
panyuenlau, line 112: Since there was an `svc` created at the beginning of the function, maybe we could just overwrite that variable instead of re-creating new ones in the for loop
```suggestion
		svc = services.Items[i]
```

pkg/resources/kafka/allBrokerService.go
```diff
@@ -36,15 +45,79 @@ func (r *Reconciler) allBrokerService() runtime.Object {
...
+		if !svc.GetDeletionTimestamp().IsZero() {
+			continue
+		}
+		err := r.Client.Delete(ctx, &svc)
```
panyuenlau, line 116: Same here, we can just overwrite the `err` that was created earlier in the function to save some memory
```suggestion
		err = r.Client.Delete(ctx, &svc)
```

panyuenlau, line 116: Same here, we can just overwrite the `err` that was created earlier in the function to save some memory
```suggestion
		err = r.Client.Delete(ctx, &svc)
```

panyuenlau (dismissed):
Just a couple of nits

panyuenlau (approved):

pregnor (approved):

pregnor:
Give a thought to [this comment](https://github.com/banzaicloud/koperator/pull/779/files#r857073328) before merging.

> bincherry: I found that cruisecontrol should restart after the service changing, otherwise it will runs into error.

stoader:
`koperator` reconciler processes first the Kafka cluster than the Cruise Control deployment:
https://github.com/banzaicloud/koperator/blob/master/controllers/kafkacluster_controller.go#L120-L121

When CC is reconciled it's config will be updated with the new kafka bootstrap server address: https://github.com/banzaicloud/koperator/blob/master/pkg/resources/cruisecontrol/configmap.go#L54

CC pod will exit with an error as won't be able to connect to the old deleted service, when i's restarted by the pod will pick up the new CC config which has the new Kafka service.

If we want to make CC to pick up the new address without exiting with an error we should rather compute the hash from the config map as that contains both user provided config and generated config.

PR #802, opened 2022-04-27
Some broker pods get stuck caused by a terminated container when there is no healthy broker in the kafka cluster.

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |

### What's in this PR?
 If reconcilePerBrokerDynamicConfig gets an error then I let the loop continue to the next broker. A combined error will be returned after the broker loop so the operator will try again later. 

### Why?
When in a kafka cluster there is not any working kafka broker present, but there are multiple broker pods with terminated container, only one broker pod will be restarted. Other brokers with terminated containers will not until  at least one healthy broker will be present in the kafka cluster. The root cause of this that brokers are unreachable and reconcilePerBrokerDynamicConfig function can't set dynamic configs and It will be ended with an error.

### Checklist
- [x] Implementation tested

### DIFF
--- a/pkg/resources/kafka/kafka.go
+++ b/pkg/resources/kafka/kafka.go
@@ -140,6 +140,7 @@ func getLoadBalancerIP(foundLBService *corev1.Service) (string, error) {
 }
 
 // Reconcile implements the reconcile logic for Kafka
+//gocyclo:ignore
 func (r *Reconciler) Reconcile(log logr.Logger) error {
 	log = log.WithValues("component", componentName, "clusterName", r.KafkaCluster.Name, "clusterNamespace", r.KafkaCluster.Namespace)
 
--- a/pkg/resources/kafka/kafka.go
+++ b/pkg/resources/kafka/kafka.go
@@ -239,8 +239,8 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
 		log.Error(err, "could not find controller broker")
 	}
 
-	var perBrokerDynamicConfigCombineError error
 	reorderedBrokers := reorderBrokers(brokerPods, r.KafkaCluster.Spec.Brokers, r.KafkaCluster.Status.BrokersState, controllerID)
+	isPerBrokerDynamicConfigFailed := false
 	for _, broker := range reorderedBrokers {
 		brokerConfig, err := broker.GetBrokerConfig(r.KafkaCluster.Spec)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


Kuvesz (approved):
LGTM

pkg/resources/kafka/kafka.go
```diff
@@ -283,11 +284,18 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
 		if err = r.updateStatusWithDockerImageAndVersion(broker.Id, brokerConfig, log); err != nil {
 			return err
 		}
+		// If dynamic configuration gets an error then let the loop continue to the next broker
+		// combined error will be returned after the broker loop. This solve that case when other brokers could get healthy,
+		// but the loop exits too soon because brokers are unreachable and dynamic configs cannot be set.
 		if err = r.reconcilePerBrokerDynamicConfig(broker.Id, brokerConfig, configMap, log); err != nil {
-			return err
+			perBrokerDynamicConfigCombineError = errors.Combine(perBrokerDynamicConfigCombineError, err)
```
hi-im-aren, line 291: We could add some details for the error like `broker.Id, brokerConfig, configMap`

pregnor (dismissed):
LGTM

Kuvesz (dismissed):

pkg/resources/kafka/kafka.go
```diff
@@ -283,11 +284,21 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
...
+		err = r.reconcilePerBrokerDynamicConfig(broker.Id, brokerConfig, configMap, log)
+		if err != nil {
+			log.Error(err, "setting dynamic configs has failed", "brokerID", broker.Id)
+			isPerBrokerDynamicConfigFailed = true
 		}
 	}
 
+	if isPerBrokerDynamicConfigFailed {
+		// re-reconcile to retry setting the dynamic configs
+		return errors.New("setting dynamic configs for some brokers has failed.")
```
stoader, line 299: Add context to the error (e.g. see https://github.com/banzaicloud/koperator/blob/master/pkg/resources/kafka/kafka.go#L304-L307)

pkg/resources/kafka/kafka.go
```diff
@@ -240,6 +240,7 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
 	}
 
 	reorderedBrokers := reorderBrokers(brokerPods, r.KafkaCluster.Spec.Brokers, r.KafkaCluster.Status.BrokersState, controllerID)
+	isPerBrokerDynamicConfigFailed := false
```
stoader, line 243: How about `allBrokerDynamicConfigSucceeded := true` and than if an error occurs set it to `false`

stoader (changes requested):

pregnor (dismissed):

stoader (dismissed):

pregnor (approved):

bartam1:
I have disabled the cyclo check for the reconcile function temporarily.
Im going to make refactor for the reconcile function and put back cyclo check in another PR

stoader (approved):

PR #822, opened 2022-06-13
Fix CN length limitation issue while using cert-manager

### LINKED ISSUE #445: Controller certificate CN too long
**Describe the bug**
Kafka clusters with medium sized names in namespaces with medium sized names fail to create, with an error like the following in the logs:

```
{"level":"error","ts":1599151167.7324488,"logger":"controller-runtime.controller","msg":"Reconciler error","controller":"kafkauser","request":"pushdelivery-stage/kafka-pushdelivery-stage-controller.pushdelivery-stage.mgt.cluster.local","error":"could not create user certificate: admission webhook \"webhook.cert-manager.io\" denied the request: spec.commonName: Too long: must have at most 64 bytes","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.0/pkg/internal/controller/controller.go:258\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.0/pkg/internal/controller/controller.go:232\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.0/pkg/internal/controller/controller.go:211\nk8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1\n\t/go/pkg/mod/k8s.io/apimachinery@v0.18.2/pkg/util/wait/wait.go:155\nk8s.io/apimachinery/pkg/util/wait.BackoffUntil\n\t/go/pkg/mod/k8s.io/apimachinery@v0.18.2/pkg/util/wait/wait.go:156\nk8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/go/pkg/mod/k8s.io/apimachinery@v0.18.2/pkg/util/wait/wait.go:133\nk8s.io/apimachinery/pkg/util/wait.Until\n\t/go/pkg/mod/k8s.io/apimachinery@v0.18.2/pkg/util/wait/wait.go:90"}
```

Create a Kafka cluster in a namespace whose effective commonName is longer than 64 characters

**Additional context**
I ran into this error rebuilding my cluster after https://github.com/banzaicloud/kafka-operator/issues/444 happened. Weirdly, this was successful before, and in production I have an even longer name ("production" vs "stage") that is running just fine. Even though I have the same versions of cert-manager and kafka-operator in both environments, the stage environment has been upgraded to Kubernetes 1.16, while production is still on 1.15. 

I found a workaround for this, which was to disable the admission webhook for cert-manager. I think I created these clusters (and their certs) before the webhook was in play, which is why production is still working - the resource was already admitted before the validation was added.

panyuenlau:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets | fixes #445|

### What's in this PR?

Hash and limit the length of the generated CN for the controller if it exceeds the maximum supported length - 64 characters, reference:  https://docs.digicert.com/manage-certificates/public-certificates-data-entries-that/#64character-maximum-limit-violation

### Additional context

Tested with cert-manager (as well as csr-operator) using the same Kafka name and namespace as reported in #445 

### Checklist

- [x] Implementation tested

### TODO
- [x] Fix following error when SSL is enabled: 
`{"level":"info","ts":"2022-06-15T05:01:57.596Z","logger":"controller.KafkaCluster","msg":"A new resource was not found or may not be ready","reconciler group":"kafka.banzaicloud.io","reconciler kind":"KafkaCluster","name":"kafka-pushdelivery-stage-controller","namespace":"pushdelivery-stage","error":"checking secret data fields: SSL JKS certificate has not generated properly yet into secret: kafka-pushdelivery-stage-controller-server-certificate"}
`

### DIFF
--- a/pkg/util/pki/common.go
+++ b/pkg/util/pki/common.go
@@ -57,6 +57,8 @@ const (
 	// KafkaUserAnnotationName used in case of PKIbackend is k8s-csr to find the appropriate kafkauser in case of
 	// signing request event
 	KafkaUserAnnotationName = "banzaicloud.io/owner"
+	// MaxCertManagerCNLen specifies the maximum common name supported by cert-manager
+	MaxCertManagerCNLen = 64
 )
 
 // Manager is the main interface for objects performing PKI operations
@@ -215,7 +217,7 @@ func sortAndDedupe(hosts []string) []string {
 func ControllerUserForCluster(cluster *v1beta1.KafkaCluster) *v1alpha1.KafkaUser {
 	return &v1alpha1.KafkaUser{
 		ObjectMeta: templates.ObjectMeta(
-			TruncatedCommonName(fmt.Sprintf(BrokerControllerFQDNTemplate, fmt.Sprintf(BrokerControllerTemplate, cluster.Name), cluster.Namespace, cluster.Spec.GetKubernetesClusterDomain()), 64),
+			TruncatedCommonName(fmt.Sprintf(BrokerControllerFQDNTemplate, fmt.Sprintf(BrokerControllerTemplate, cluster.Name), cluster.Namespace, cluster.Spec.GetKubernetesClusterDomain()), MaxCertManagerCNLen),
 			LabelsForKafkaPKI(cluster.Name, cluster.Namespace),
 			cluster,
 		),
@@ -244,8 +246,8 @@ func EnsureControllerReference(ctx context.Context, user *v1alpha1.KafkaUser,
 	return nil
 }
 
-// TruncatedCommonName ensures that the passed-in CN name doesn't exceed the specified number of characters
-func TruncatedCommonName(name string, maxLen int) string{
+// TruncatedCommonName ensures that the passed-in common name doesn't exceed the specified number of characters
+func TruncatedCommonName(name string, maxLen int) string {
 	n := []rune(name)
 	if len(n) > maxLen {
 		return string(n[:maxLen])
--- a/pkg/util/pki/common.go
+++ b/pkg/util/pki/common.go
@@ -187,7 +187,7 @@ func BrokerUserForCluster(cluster *v1beta1.KafkaCluster, extListenerStatuses map
 	}
 	additionalHosts = sortAndDedupe(additionalHosts)
 	return &v1alpha1.KafkaUser{
-		ObjectMeta: templates.ObjectMeta(GetCommonName(cluster), LabelsForKafkaPKI(cluster.Name, cluster.Namespace), cluster),
+		ObjectMeta: templates.ObjectMeta(EnsureValidCommonNameLen(GetCommonName(cluster)), LabelsForKafkaPKI(cluster.Name, cluster.Namespace), cluster),
 		Spec: v1alpha1.KafkaUserSpec{
 			SecretName: fmt.Sprintf(BrokerServerCertTemplate, cluster.Name),
 			DNSNames:   append(GetInternalDNSNames(cluster), additionalHosts...),
--- a/pkg/util/pki/common.go
+++ b/pkg/util/pki/common.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/banzaicloud/koperator/pkg/errorfactory"
 	"github.com/banzaicloud/koperator/pkg/k8sutil"
+	"github.com/banzaicloud/koperator/pkg/util"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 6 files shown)


pkg/util/pki/common.go
```diff
@@ -243,3 +245,12 @@ func EnsureControllerReference(ctx context.Context, user *v1alpha1.KafkaUser,
 	}
 	return nil
 }
+
+// TruncatedCommonName ensures that the passed-in common name doesn't exceed the specified number of characters
+func TruncatedCommonName(name string, maxLen int) string {
```
hi-im-aren, line 250: We could use `TruncateCommonName` instead so it's a verb as a function name?
Maybe we could move this to a common/util package and make it a generic string truncate function.

pregnor (approved):

pkg/util/pki/common.go
```diff
@@ -57,6 +58,8 @@ const (
 	// KafkaUserAnnotationName used in case of PKIbackend is k8s-csr to find the appropriate kafkauser in case of
 	// signing request event
 	KafkaUserAnnotationName = "banzaicloud.io/owner"
+	// MaxCertManagerCNLen specifies the maximum common name supported by cert-manager
+	MaxCertManagerCNLen = 64
```
stoader, line 62: The max length limit on CN field I think is rather imposed by certificate standards and not cert-manager this comment should be updated accordingly (see https://docs.digicert.com/manage-certificates/public-certificates-data-entries-that/#64character-maximum-limit-violation)

pkg/util/pki/common.go
```diff
@@ -215,9 +218,9 @@ func sortAndDedupe(hosts []string) []string {
 func ControllerUserForCluster(cluster *v1beta1.KafkaCluster) *v1alpha1.KafkaUser {
 	return &v1alpha1.KafkaUser{
 		ObjectMeta: templates.ObjectMeta(
-			fmt.Sprintf(BrokerControllerFQDNTemplate,
-				fmt.Sprintf(BrokerControllerTemplate, cluster.Name), cluster.Namespace, cluster.Spec.GetKubernetesClusterDomain()),
-			LabelsForKafkaPKI(cluster.Name, cluster.Namespace), cluster,
+			util.TruncateStringByLen(fmt.Sprintf(BrokerControllerFQDNTemplate, fmt.Sprintf(BrokerControllerTemplate, cluster.Name), cluster.Namespace, cluster.Spec.GetKubernetesClusterDomain()), MaxCertManagerCNLen),
```
stoader, line 221: Do we need to truncate the name of the created `KafkaUser` custom resource as well? Or only just the CommonName field in the created `Certificate` resource? I'm asking as for K8s resource names different rules apply (see https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names)

pkg/util/util.go
```diff
@@ -451,3 +451,12 @@ func ObjectManagedByClusterRegistry(object metav1.Object) bool {
 	_, ok := annotations[clusterregv1alpha1.OwnershipAnnotation]
 	return ok
 }
+
+// TruncatedCommonName ensures that the passed-in common name doesn't exceed the specified number of characters
+func TruncateStringByLen(name string, l int) string {
```
stoader, line 456: Is it enough just to simply truncate the common name or we have to ensure that it stays unique after truncation? Eg two different KafkaUsers should be represented by two different certificates.

If uniqueness needs to be ensured than a has code could be be computed from the original value and appended to the truncated value.

stoader (changes requested):

pkg/util/pki/common.go
panyuenlau: We need to ensure the `name` field of the generated `KafkaUser` doesn't exceed the length limitation as well because the [`clusterCertificateForUser`](https://github.com/banzaicloud/koperator/blob/master/pkg/pki/certmanagerpki/certmanager_user.go#L146) function uses it as the `spec.commonName` while generating a `Certificate` object for the corresponding `KafkaUser`

In fact, it is the first error reported by the operator when our user reported this issue in #445

pkg/util/util.go
panyuenlau: I thought about hashing the entire string and limiting it to 64 chars as well, but I realized it could be difficult for the users to debug if things go wrong since the hashed string is un-comprehensible.

In terms of uniqueness, since the common name is constructed with `cluster.Name` and`cluster.Namespace`, I thought this is enough to ensure the uniqueness of the entire string since we cant't have the same kafka cluster name under the same namespace

pkg/util/pki/common.go
panyuenlau: will update the var and comment to reflect this

pkg/util/pki/common.go
stoader: In `clusterCertificateForUser` you could just perform the truncation just before passing it into `spec.commonName`

pregnor (dismissed):

pkg/util/pki/pki_common_test.go
```diff
@@ -175,3 +175,42 @@ func TestControllerUserForCluster(t *testing.T) {
...
+	testCases := []struct {
+		testName             string
+		commonName           string
+		expectedHashedString string
+		expectedLen          int
+	}{
+		{
+			testName:    "Common name with length exceeded limitation",
+			commonName:  "kafka-test-controller-exceeded-length.test-namespace.mgt.cluster.local",
+			expectedLen: MaxCNLen,
```
stoader, line 189: beside the lenght also the expected output could be verified as well.

panyuenlau: yea...but we since don't actually care about the content of the hashed string, I don't see the necessity of verifying it in the test

stoader (dismissed):

Kuvesz (approved):

stoader (approved):

pregnor (approved):

PR #823, opened 2022-06-17
Envoy config generated by the operator is invalid in envoy 1.22

amuraru:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | yes |
| API breaks?     | no |
| Deprecations?   | no |

### What's in this PR?

Envoy config generated by the operator was not compatible with newer version of the envoy. More specifically in v1.22 the config is invalid:
```
[2022-06-16 13:27:58.425][1][info][main] [source/server/server.cc:939] exiting
Didn't find a registered implementation for 'envoy.filters.http.router' with type URL: ''
```
This PR addresses this by:
- fixing the expected typedConfig in the router config
- fixing deprecated options used in the generated envoy.yaml file
- upgrades default envoy to v1.22
- upgrades github.com/envoyproxy/go-control-plane to latest version
- 

### Why?

Being able to upgrade to newer version of envoy when envoy ingress is used to front Kafka clusters.

### Checklist

- [x] Implementation tested


stoader (changes requested):
Since the user can override the image for Envoy via `envoyConfig.image` the config generation should take into account the version and generate config specific to the version.

The other option is we mandate the use of Envoy 1.22.2 and above from now on.

amuraru:
Good points @stoader - in fact the new generated version of envoy.yaml should be backward compatible with envoy < 1.22
I'll double check that and report back here

Kuvesz (approved):
I think this looks good if it is indeed backward compatible.

amuraru:
@stoader @Kuvesz I updated the PR to include only the minimal changes to make the generated config compatible with newer versions of envoy (e.g. 1.22.x)
Tested this out and it works with envoy 1.18 as well - so this is not breaking the clients still using older versions.

I'll follow-up with a separate PR to make envoy 1.22 as min required if you agree:
https://github.com/amuraru/koperator/commit/1e1f9bcd337964d6673c017f372c27088e3621a2
This change though is not backward compatible so we may need to think through it more.

stoader (approved):

hi-im-aren (approved):

hi-im-aren:
Thanks! 👍

PR #824, opened 2022-06-17
Re-create missing pod which has incompleted downscale operation

### LINKED ISSUE #776: Kafka broker not coming back during scale-down
During a scale-down operation if a broker that is to be removed is deleted in a non controlled fashion (while the data is being drained from the Kafka broker to other brokers) the operator doesn't bring the broker back up to finish this and then remove it again.
This in turn will slow down new replicas getting in sync (as they will have to pull from 2 or even worse, 1 in sync replica).
Even more concerning, if there is a K8s cluster rollout restart (due to VM upgrades for examples) if the Kafka pods don't come back up, this will result in offline partitions.

If a broker is killed in a non controlled fashion while it's also being drained, the broker should be brought back in the cluster and removed after all replicas are removed.

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no |
| New feature?    | yes |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets | fixes #776  |

### What's in this PR?
There is a new field in the broker status. Its name is configurationBackup. 
It stores broker configuration in gzipped base64 format.
When a broker is missing and has not been removed gracefully (incompleted downscale operation), the operator will re-create that pod based on the configurationBackup thus graceful downscale operation can be continued.

### Why?
During a scale-down operation if a broker that is to be removed is deleted in a non controlled fashion (while the data is being drained from the Kafka broker to other brokers) the operator doesn't bring the broker back up to finish this and then remove it again.
This in turn will slow down new replicas getting in sync (as they will have to pull from 2 or even worse, 1 in sync replica).
Even more concerning, if there is a K8s cluster rollout restart (due to VM upgrades for examples) if the Kafka pods don't come back up, this will result in offline partitions.

### Checklist

- [x] Implementation tested


pkg/k8sutil/status.go
```diff
@@ -43,6 +44,58 @@ func IsMarkedForDeletion(m metav1.ObjectMeta) bool {
...
+		cluster.Status.BrokersState = make(map[string]banzaicloudv1beta1.BrokerState)
+	}
+
+	for _, broker := range cluster.Spec.Brokers {
+		brokerState := cluster.Status.BrokersState[fmt.Sprint(broker.Id)]
+		configurationBackup, err := util.GzipAndBase64BrokerConfiguration(&broker)
+		if err != nil {
+			return false, errors.WrapIfWithDetails(err, "could not generate broker configuration backup", "brokerId", broker.Id)
+		}
+		if !needsUpdate && configurationBackup != brokerState.ConfigurationBackup {
```
Kuvesz, line 90: I don't think `needsUpdate` can be true here in any case as it has been created as false and hasn't even been referenced since it's creation until this part.

pkg/resources/kafka/kafka.go
```diff
@@ -1203,20 +1218,50 @@ func getServiceFromExternalListener(client client.Client, cluster *v1beta1.Kafka
...
 
+	presentPersistentVolumeClaims := make(map[string]struct{})
+	for _, pvc := range pvcList.Items {
+		brokerID := pvc.GetLabels()["brokerId"]
+		presentPersistentVolumeClaims[brokerID] = struct{}{}
+	}
+
 	brokersReconcilePriority := make(map[string]brokerReconcilePriority, len(desiredBrokers))
+	missingBrokerDownScaleRunning := make(map[string]struct{})
+	// logic for handle that case when a broker pod is removed before downscale operation completed
```
Kuvesz, line 1236:
```suggestion
	// logic for handling the case when a broker pod is removed before downscale operation completed
```

pkg/resources/kafka/kafka.go
```diff
@@ -76,8 +76,10 @@ const (
 	MetricsHealthCheck = "/-/healthy"
 	MetricsPort        = 9020
 
+	// missingBrokerDownScaleRunningPriority the priority used  for missing brokers where there is an incompleted downscale operation
```
Kuvesz, line 79:
```suggestion
	// missingBrokerDownScaleRunningPriority is the priority used for missing brokers where there is an incomplete downscale operation
```

pkg/resources/kafka/kafka.go
```diff
@@ -699,15 +710,19 @@ func (r *Reconciler) reconcileKafkaPod(log logr.Logger, desiredPod *corev1.Pod,
 			return errorfactory.New(errorfactory.StatusUpdateError{}, statusErr, "updating per broker config status for resource failed", "kind", desiredType)
 		}
 
-		if val, ok := r.KafkaCluster.Status.BrokersState[desiredPod.Labels["brokerId"]]; ok && val.GracefulActionState.CruiseControlState != v1beta1.GracefulUpscaleSucceeded {
-			gracefulActionState := v1beta1.GracefulActionState{ErrorMessage: "CruiseControl not yet ready", CruiseControlState: v1beta1.GracefulUpscaleSucceeded}
+		if val, hasBrokerState := r.KafkaCluster.Status.BrokersState[desiredPod.Labels["brokerId"]]; hasBrokerState {
+			ccState := val.GracefulActionState.CruiseControlState
+			incompletedDownscale := ccState == v1beta1.GracefulDownscaleRequired || ccState == v1beta1.GracefulDownscaleRunning
+			if ccState != v1beta1.GracefulUpscaleSucceeded && !incompletedDownscale {
```
Kuvesz, line 716:
```suggestion
			incompleteDownscale := ccState == v1beta1.GracefulDownscaleRequired || ccState == v1beta1.GracefulDownscaleRunning
			if ccState != v1beta1.GracefulUpscaleSucceeded && !incompleteDownscale {
```

pkg/resources/kafka/kafka.go
```diff
@@ -1203,20 +1218,50 @@ func getServiceFromExternalListener(client client.Client, cluster *v1beta1.Kafka
...
+
 	brokersReconcilePriority := make(map[string]brokerReconcilePriority, len(desiredBrokers))
+	missingBrokerDownScaleRunning := make(map[string]struct{})
+	// logic for handle that case when a broker pod is removed before downscale operation completed
+	for id, brokerState := range brokersState {
+		_, running := runningBrokers[id]
+		_, pvcPresent := presentPersistentVolumeClaims[id]
+		ccState := brokerState.GracefulActionState.CruiseControlState
+		if !running && ccState == v1beta1.GracefulDownscaleRequired || ccState == v1beta1.GracefulDownscaleRunning {
+			log.Info("missing broker found with incompleted downscale operation", "brokerID", id)
```
Kuvesz, line 1242:
```suggestion
			log.Info("missing broker found with incomplete downscale operation", "brokerID", id)
```

pkg/resources/kafka/kafka_test.go
```diff
@@ -438,6 +440,90 @@ func TestReorderBrokers(t *testing.T) {
 				{Id: 1}, // controller broker should be last
 			},
 		},
+		{
+			testName: "some missing broker pods, newly added brokers, and missing bokers with pvc and incompleted downscale operation",
```
Kuvesz, line 444:
```suggestion
			testName: "some missing broker pods, newly added brokers, and missing bokers with pvc and incomplete downscale operation",
```

pkg/resources/kafka/kafka_test.go
```diff
@@ -438,6 +440,90 @@ func TestReorderBrokers(t *testing.T) {
...
+				"3": {ConfigurationState: v1beta1.ConfigOutOfSync},
+				"4": {ConfigurationState: v1beta1.ConfigOutOfSync},
+				"5": {
+					ConfigurationState:  v1beta1.ConfigInSync,
+					GracefulActionState: v1beta1.GracefulActionState{CruiseControlState: v1beta1.GracefulDownscaleRequired},
+					ConfigurationBackup: "H4sIAAxcrGIAA6tWykxRsjKt5QIAiMWU3gkAAAA=",
+				}},
+			controllerBrokerID: 1,
+			expectedReorderedBrokers: []v1beta1.Broker{
+				{Id: 5}, // broker pod 5 missing and there is incompleted downscale operation thus should have highest prio
```
Kuvesz, line 480:
```suggestion
				{Id: 5}, // broker pod 5 missing and there is an incomplete downscale operation thus should have highest prio
```

pkg/resources/kafka/kafka_test.go
```diff
@@ -438,6 +440,90 @@ func TestReorderBrokers(t *testing.T) {
...
+				{Id: 6}, // broker 6 is newly added thus should have higher prio
+				{Id: 3}, // broker pod 3 missing thus should have higher prio
+				{Id: 4}, // broker pod 4 missing thus should have higher prio
+				{Id: 0},
+				{Id: 2},
+				{Id: 1}, // controller broker should be last
+			},
+		},
+		{
+			testName: "some missing broker pods, newly added brokers, and missing bokers without pvc and incompleted downscale operation",
```
Kuvesz, line 490:
```suggestion
			testName: "some missing broker pods, newly added brokers, and missing bokers without pvc and incomplete downscale operation",
```

Kuvesz (changes requested):
I just requested some small changes, otherwise seems fine.

pkg/k8sutil/status.go
Kuvesz: I messed this up, this is fine like this :D

pkg/k8sutil/status.go
```diff
@@ -43,6 +44,58 @@ func IsMarkedForDeletion(m metav1.ObjectMeta) bool {
...
+
+func generateBrokerConfigurationBackups(cluster *banzaicloudv1beta1.KafkaCluster) (bool, error) {
+	needsUpdate := false
+	if cluster.Status.BrokersState == nil {
+		cluster.Status.BrokersState = make(map[string]banzaicloudv1beta1.BrokerState)
+	}
+
+	for _, broker := range cluster.Spec.Brokers {
+		brokerState := cluster.Status.BrokersState[fmt.Sprint(broker.Id)]
+		configurationBackup, err := util.GzipAndBase64BrokerConfiguration(&broker)
```
amuraru, line 86: This would increase the size of KafkaCluster custom resource. Do you see a risk of hitting the 1MB limit overall for large kafka clusters (say hundreds of brokers)?

bartam1: It takes ~500bytes/broker 500x100 = 50Kbytes.
Configuration can be stored also in configmap but I think it is not so safe.

bartam1: When different configurations are used with lot of configuration entry for many brokers there is a chance to reach the 1MB. I think it is a very edge case.

amuraru: sounds good then

amuraru: I realised that you store only the broker node and not the whole broker expanded config.

In our case - we drive the broekr config through brokerConfigGroup(s) and thus the broker entry is quite terse, hence the size here very small:
```
  brokers:
  - brokerConfigGroup: az1
    id: 1101
    readOnlyConfig: |
      broker.rack=eastus2-1
```

pkg/k8sutil/status.go
stoader, line 90: I wonder if this comparison can be further enhanced by expanding a compressed configuration backup and perform the comparison regardless the ordering of the fields.

pkg/resources/kafka/kafka.go
```diff
@@ -1203,20 +1218,50 @@ func getServiceFromExternalListener(client client.Client, cluster *v1beta1.Kafka
...
 	runningBrokers := make(map[string]struct{})
 	for _, b := range brokerPods.Items {
 		brokerID := b.GetLabels()["brokerId"]
 		runningBrokers[brokerID] = struct{}{}
 	}
 
+	presentPersistentVolumeClaims := make(map[string]struct{})
+	for _, pvc := range pvcList.Items {
+		brokerID := pvc.GetLabels()["brokerId"]
+		presentPersistentVolumeClaims[brokerID] = struct{}{}
```
stoader, line 1231: Instead of initializing `presentPersistentVolumeClaims` here do it outside of `reorderBrokers` and pass in as parameter from outside.

stoader (changes requested):

pkg/k8sutil/status.go
bartam1: I think the change of the order rarely happens and the decompress and compare operation for every broker configurationBackup for every kafkacluster reconcile would take some memory and computation from the operator.
Do you think It worth to do those computation to avoid few status update cases?

pkg/k8sutil/status.go
```diff
@@ -43,6 +44,58 @@ func IsMarkedForDeletion(m metav1.ObjectMeta) bool {
...
+		if err := c.Get(ctx, types.NamespacedName{Namespace: cluster.Namespace, Name: cluster.Name}, cluster); err != nil {
+			return errors.WrapIf(err, "could not get config for updating status")
+		}
+		needsUpdate, err := generateBrokerConfigurationBackups(cluster)
+		if err != nil {
+			return err
+		}
+		if !needsUpdate {
+			return nil
+		}
```
ecojan, line 70: Is there a way we can avoid code duplication here? (I see we have a section above doing the same thing)

pkg/resources/kafka/kafka.go
stoader: You're right the `runnigBrokers` should be passed in from outside as well. (e.g. `runnigBrokers sets.String` and `brokersWithPVCs sets.String`

pkg/resources/kafka/kafka.go
```diff
@@ -1203,20 +1218,50 @@ func getServiceFromExternalListener(client client.Client, cluster *v1beta1.Kafka
...
-func reorderBrokers(brokerPods corev1.PodList, desiredBrokers []v1beta1.Broker, brokersState map[string]v1beta1.BrokerState, controllerBrokerID int32) []v1beta1.Broker {
+func reorderBrokers(brokerPods corev1.PodList, desiredBrokers []v1beta1.Broker, brokersState map[string]v1beta1.BrokerState, pvcList corev1.PersistentVolumeClaimList, controllerBrokerID int32, log logr.Logger) []v1beta1.Broker {
 	runningBrokers := make(map[string]struct{})
 	for _, b := range brokerPods.Items {
 		brokerID := b.GetLabels()["brokerId"]
 		runningBrokers[brokerID] = struct{}{}
 	}
 
+	presentPersistentVolumeClaims := make(map[string]struct{})
+	for _, pvc := range pvcList.Items {
```
stoader, line 1229: Checking if the broker's PVC exists alone is not enough, we need to check if the underlying PV that stores kafka log dirs exists. The status of the PVC needs to be verified to be in `Bound` state. (see `Phase` field of https://pkg.go.dev/k8s.io/api@v0.23.8/core/v1#PersistentVolumeClaimStatus)

pkg/k8sutil/status.go
stoader: I think it can be if the retry is done with the use of the `backoff` package.

pkg/resources/kafka/kafka.go
```diff
@@ -699,15 +710,19 @@ func (r *Reconciler) reconcileKafkaPod(log logr.Logger, desiredPod *corev1.Pod,
 			return errorfactory.New(errorfactory.StatusUpdateError{}, statusErr, "updating per broker config status for resource failed", "kind", desiredType)
 		}
 
-		if val, ok := r.KafkaCluster.Status.BrokersState[desiredPod.Labels["brokerId"]]; ok && val.GracefulActionState.CruiseControlState != v1beta1.GracefulUpscaleSucceeded {
-			gracefulActionState := v1beta1.GracefulActionState{ErrorMessage: "CruiseControl not yet ready", CruiseControlState: v1beta1.GracefulUpscaleSucceeded}
+		if val, hasBrokerState := r.KafkaCluster.Status.BrokersState[desiredPod.Labels["brokerId"]]; hasBrokerState {
+			ccState := val.GracefulActionState.CruiseControlState
+			incompleteDownscale := ccState == v1beta1.GracefulDownscaleRequired || ccState == v1beta1.GracefulDownscaleRunning
```
stoader, line 715: Couldn't `incompleteDownscale` be replaced with `ccState.IsDownscale()` ?

stoader (changes requested):

pkg/resources/kafka/kafka.go
bartam1: I think it can be used. It is not the same, but it should works. 
func (r CruiseControlState) IsDownscale() bool {
	return r == GracefulDownscaleRequired || r == GracefulDownscaleSucceeded || r == GracefulDownscaleRunning
}
First the kafafkaPodDelete() run and it will remove the broker from status with GracefulDownscaleSucceeded. So to check the  r == GracefulDownscaleSucceeded part is unnecessary. It is safe to use the isDownscale() I'm going to change the implementation.

stoader: Checking `GracefulDownscaleSucceeded` is still needed to avoid unnecessary update.

pkg/resources/kafka/kafka.go
amuraru: one remark here: in our environemnt we don't have access to the PersistentVolume resources - we can access only the PVC CR

stoader: I don't think it's necessary to check PVs it's enough verify the status of the PVC

ecojan:
@bartam1 managed to test this locally today, working as expected now 👍 thank you for the effort put in here!

pkg/resources/kafka/kafka.go
```diff
@@ -1203,20 +1232,38 @@ func getServiceFromExternalListener(client client.Client, cluster *v1beta1.Kafka
...
 	brokersReconcilePriority := make(map[string]brokerReconcilePriority, len(desiredBrokers))
+	missingBrokerDownScaleRunning := make(map[string]struct{})
+	// logic for handling that case when a broker pod is removed before downscale operation completed
+	for id, brokerState := range brokersState {
+		_, running := runningBrokers[id]
+		_, pvcPresent := boundPersistentVolumeClaims[id]
+		ccState := brokerState.GracefulActionState.CruiseControlState
+		if !running && ccState == v1beta1.GracefulDownscaleRequired || ccState == v1beta1.GracefulDownscaleRunning {
+			log.Info("missing broker found with incomplete downscale operation", "brokerID", id)
+			if pvcPresent {
```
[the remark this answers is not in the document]
stoader: @amuraru according to https://kubernetes.io/docs/concepts/storage/persistent-volumes/#binding the `Bound` state indicates that the PVC is bound to a PV so it's not related to the state of the Pod the underlying volume is mounted to.

pkg/resources/kafka/kafka.go
```diff
@@ -1203,20 +1232,38 @@ func getServiceFromExternalListener(client client.Client, cluster *v1beta1.Kafka
...
-
+func reorderBrokers(runningBrokers, boundPersistentVolumeClaims map[string]struct{}, desiredBrokers []v1beta1.Broker, brokersState map[string]v1beta1.BrokerState, controllerBrokerID int32, log logr.Logger) []v1beta1.Broker {
 	brokersReconcilePriority := make(map[string]brokerReconcilePriority, len(desiredBrokers))
+	missingBrokerDownScaleRunning := make(map[string]struct{})
+	// logic for handling that case when a broker pod is removed before downscale operation completed
+	for id, brokerState := range brokersState {
+		_, running := runningBrokers[id]
+		_, pvcPresent := boundPersistentVolumeClaims[id]
+		ccState := brokerState.GracefulActionState.CruiseControlState
+		if !running && ccState == v1beta1.GracefulDownscaleRequired || ccState == v1beta1.GracefulDownscaleRunning {
```
pregnor, line 1243: **Optional**: we **MAY** improve readability by adding parenthesis to emphasize the intended evaluation order which is now implicitly determined from operator precedence.

```suggestion
		if (!running && ccState == v1beta1.GracefulDownscaleRequired) || ccState == v1beta1.GracefulDownscaleRunning {
```

pkg/util/util_test.go
```diff
@@ -607,3 +607,82 @@ func TestIsIngressConfigInUse(t *testing.T) {
...
+		}
+
+		broker, err := GetBrokerFromBrokerConfigurationBackup(config)
+		if err != nil {
+			t.Errorf("error should be nil, got: %v", err)
+		}
+
+		if !reflect.DeepEqual(test.broker, broker) {
+			t.Errorf("Expected: %v  Got: %v", test.broker, broker)
+		}
```
pregnor, line 686: **Note**: we might want to separate these two unit test cases in the future and test the functions for their functionality, because right now we are only ensuring that the `GetBrokerFromBrokerConfigurationBackup()` function is the inverse of the `GzipAndBase64BrokerConfiguration()` function, but not that they do what we expect of them (gzip and base64 encoding the JSON broker configuration and doing the reverse for decoding).
For example you pass a pointer to the `test.broker` to the `GzipAndBase64BrokerConfiguration()` function which could in theory by accident overwrite the values of it to make it an empty `v1beta1.Broker{}` and `GetBrokerFromBrokerConfigurationBackup()` could just be returning empty `v1beta1.Broker{}` objects all the time and the `reflect.DeepEqual()` call would still be true, but the functions would not be doing what they are expected to do.

I do admit the functions seem to work correctly based on their current code, but the test itself does not guarantee to uphold that state in case the functions are modified in the future for any reason.

pregnor (approved):

pkg/k8sutil/status.go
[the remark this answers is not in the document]
bartam1: 1. needsUpdate is false, but if u see in the condition there is negation, so the first time it will true and later it will true so the negation will be false. This condition is only satisfied only once.

pkg/util/util_test.go
bartam1: Thank you for the explanation. This would be an edge case but you are right. I missed this case.

pkg/k8sutil/status.go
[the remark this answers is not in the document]
pregnor: Sorry, misread this part, you are right.

pregnor (changes requested):

pregnor (dismissed):
LGTM

pregnor (dismissed):

pregnor (dismissed):

pkg/k8sutil/status.go
```diff
@@ -43,6 +45,61 @@ func IsMarkedForDeletion(m metav1.ObjectMeta) bool {
...
+		errAPI = c.Status().Update(ctx, cluster)
+		if errAPI != nil {
+			if apierrors.IsConflict(errAPI) {
+				if err := c.Get(ctx, types.NamespacedName{Namespace: cluster.Namespace, Name: cluster.Name}, cluster); err != nil {
+					return errors.WrapIf(err, "could not get config for updating status")
+				}
+			}
+		} else {
```
stoader, line 74: this `else` is not needed.

pkg/util/util.go
```diff
@@ -451,3 +456,78 @@ func ObjectManagedByClusterRegistry(object metav1.Object) bool {
...
+
+// start with 20 ms, multiply by 2.5 each step, 6 steps = ~3 seconds
+var DefaultBackOffForConflict wait.Backoff = wait.Backoff{
+	Duration: 20 * time.Millisecond,
+	Jitter:   0,
+	Factor:   2.5,
+	Steps:    6,
+}
+
+func RetryOnError(backoff wait.Backoff, updateFn func() error, isRetryableError func(error) bool) error {
```
stoader, line 511:
```suggestion
func RetryOnError(backoff wait.Backoff, fn func() error, isRetryableError func(error) bool) error {
```

pkg/util/util.go
```diff
@@ -451,3 +456,78 @@ func ObjectManagedByClusterRegistry(object metav1.Object) bool {
...
+	var lastConflictErr error
```
stoader, line 512:
```suggestion
	var lastErr error
```

pkg/util/util.go
```diff
@@ -451,3 +456,78 @@ func ObjectManagedByClusterRegistry(object metav1.Object) bool {
...
+	err := wait.ExponentialBackoff(backoff, func() (bool, error) {
+		err := updateFn()
```
stoader, line 514:
```suggestion
		err := fn()
```

pkg/util/util.go
```diff
@@ -451,3 +456,78 @@ func ObjectManagedByClusterRegistry(object metav1.Object) bool {
...
+		switch {
+		case err == nil:
+			return true, nil
+		case isRetryableError(err):
+			lastConflictErr = err
```
stoader, line 519:
```suggestion
			lastErr = err
```

pkg/util/util.go
```diff
@@ -451,3 +456,78 @@ func ObjectManagedByClusterRegistry(object metav1.Object) bool {
...
+			return false, nil
+		default:
+			return false, err
+		}
+	})
+	if err == wait.ErrWaitTimeout {
+		err = lastConflictErr
```
stoader, line 526:
```suggestion
		err = lastErr
```

stoader (changes requested):

pkg/resources/kafka/kafka.go
```diff
@@ -232,17 +240,37 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
...
+	var pvcList corev1.PersistentVolumeClaimList
+	err = r.Client.List(ctx, &pvcList, client.ListOption(client.InNamespace(r.KafkaCluster.Namespace)), client.ListOption(matchingLabels))
+	if err != nil {
+		return errors.WrapIf(err, "failed to list broker pvcs that belong to Kafka cluster")
+	}
+
+	boundPersistentVolumeClaims := make(map[string]struct{})
+	for _, pvc := range pvcList.Items {
+		brokerID := pvc.GetLabels()["brokerId"]
+		if pvc.Status.Phase == corev1.ClaimBound {
```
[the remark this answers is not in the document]
stoader: Bound PVC means that there is a PV that matches the PVC and is bound to the PVC: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#binding

A PV can be created without being attached to a VM (Kubernetes node). e.g. if you create a PVC which uses a Storage class that has `volumeBindingMode: Immediate` a PV will be provisioned and bound to the PVC without the disk being mounted to any VM. The spec of the PV provisioned dynamically through the PVC will indicate which zone it was provisioned in:
```
apiVersion: v1
kind: PersistentVolume
...
spec:
  ...
  nodeAffinity:
    required:
      nodeSelectorTerms:
      - matchExpressions:
        - key: topology.kubernetes.io/zone
          operator: In
          values:
          - us-east-2a
        - key: topology.kubernetes.io/region
          operator: In
          values:
          - us-east-2
  persistentVolumeReclaimPolicy: Delete
  storageClassName: gp2immediate
  volumeMode: Filesystem
status:
  phase: Bound
```
(see https://kubernetes.io/docs/concepts/storage/persistent-volumes/#node-affinity)

If PV will be actually attached to a VM when it its mounted to a Pod in which case it will be attached to the VM that hosts the Pod.
If a Storage class with `volumeBindingMode: WaitForFirstConsumer` is used the PV won't be provisioned until there is no Pod to use it. When the scheduler decided which node to schedule the Pod onto the PV will be provisioned in the zone of the Pod.

When the VM dies and the PV is not lost, the Pod on the VM vanishes. When the missing pods a recreated if those can be recreated on another node (VM) in the same zone Kubernetes will reattach the disk to the other node(VM) . In case the scheduler cannot schedule the pods to another in the same zone but in other zone for various reasons the pods will stuck in pending state waiting for volume mounting which will never happen as disks can be attached only to VMs in same zone.

When the VM dies and the PV is lost as well the PVC will get into `phase: Lost` which indicates the underlying PV does not exist any longer and all data on it was lost.

There might be some differences from one cloud provider to another thus always worth checking their docs regarding the volume provisioning plugins/drivers they provide for Kubernetes but I think we can rely on the `phase: Bound` field of the PVC regardless of the env as that indicates if there is an existing PV linked to the PVC. The reattaching of the disk to another VM is a different story and is environment specific how exactly works and impacts the Pod scheduling.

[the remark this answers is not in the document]
amuraru: Thanks for the detailed explanation, makes sense

stoader (approved):

pregnor (approved):
LGTM

Kuvesz (approved):

PR #832, opened 2022-07-12
Refactor of the KafkaCluster and KafkaTopic CR validation webhooks

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no |
| New feature?    | no |
| Refactor?    |yes |
| API breaks?     | no |
| Deprecations?   | no |

### What's in this PR?
Refactor for KafkaCluster and KafkaTopic validation webhooks
Validation gives better feedback from the problems
e.g:
```shell
kubectl apply -f example-topic.yaml
The KafkaTopic "example-topic" is invalid:
* spec.partitions: Invalid value: 0: number of partitions must be larger than 0 (or set it to be -1 to use the broker's default)
* spec.replicationFactor: Invalid value: 0: replication factor must be larger than 0 (or set it to be -1 to use the broker's default) 
```

### Why?
The current implementation of webhooks is obsolete.

### Checklist

- [x] Implementation tested


stoader (changes requested):

pkg/webhooks/kafkacluster_validator.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+var (
+	log = logf.Log.WithName("webhooks")
+)
+
+type KafkaClusterValidator struct {
+}
+
+func (s KafkaClusterValidator) ValidateUpdate(ctx context.Context, oldObj, newObj runtime.Object) error {
```
stoader, line 36: Why this can not be: `ValidateUpdate(ctx context.Context, oldObj, newObj *banzaicloudv1beta1.KafkaCluster)`? This method is only invoked for `KafkaCluster` type. (see https://github.com/banzaicloud/koperator/blob/master/charts/kafka-operator/templates/operator-deployment-with-webhook.yaml#L72).

pkg/webhooks/topic_validator.go
```diff
@@ -0,0 +1,232 @@
...
+	"github.com/go-logr/logr"
+
+	banzaicloudv1alpha1 "github.com/banzaicloud/koperator/api/v1alpha1"
+	banzaicloudv1beta1 "github.com/banzaicloud/koperator/api/v1beta1"
+	"github.com/banzaicloud/koperator/pkg/k8sutil"
+	"github.com/banzaicloud/koperator/pkg/kafkaclient"
+	"github.com/banzaicloud/koperator/pkg/util"
+)
+
+type KafkaTopicValidator struct {
```
stoader, line 38: Inject logger

pkg/webhooks/kafkacluster_validator.go
```diff
@@ -0,0 +1,99 @@
...
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	banzaicloudv1beta1 "github.com/banzaicloud/koperator/api/v1beta1"
+)
+
+var (
+	log = logf.Log.WithName("webhooks")
+)
+
+type KafkaClusterValidator struct {
```
stoader, line 33: Inject logger

stoader (changes requested):

pkg/webhooks/kafkacluster_validator.go
bartam1: The admission.CustomValidator needs to be satisfied by the ValidateUpdate
https://github.com/banzaicloud/koperator/blob/webhookrefact/main.go#L203

stoader: I see. Since https://github.com/banzaicloud/koperator/blob/master/charts/kafka-operator/templates/operator-deployment-with-webhook.yaml#L72 ensures the right type the webhook is invoked for we can safely assume that `oldObj.(*banzaicloudv1beta1.KafkaCluster)` succeeds thus no need for 
```
if !ok {
 		log.Info(unableToRecognizeMsg)
 		return apiErrors.NewBadRequest(unableToRecognizeMsg)
 	}
```

what do you think?

stoader (changes requested):

charts/kafka-operator/templates/operator-deployment-with-webhook.yaml
```diff
@@ -37,7 +37,7 @@ webhooks:
     service:
       name: "{{ include "kafka-operator.fullname" . }}-operator"
       namespace: {{ .Release.Namespace }}
-      path: /validate
+      path: /validate-kafka.banzaicloud-io-v1alpha1-kafkatopic
```
stoader, line 40: This change in webhook path needs to be reflected in https://github.com/banzaicloud/koperator/tree/master/config/base/webhook as well

pkg/webhooks/kafkacluster_validator.go
```diff
@@ -0,0 +1,99 @@
...
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	banzaicloudv1beta1 "github.com/banzaicloud/koperator/api/v1beta1"
+)
+
+var (
+	log = logf.Log.WithName("webhooks")
+)
+
```
stoader, line 32: Add `+kubebuilder:webhook` tag onto this type: https://book.kubebuilder.io/reference/markers/webhook.html

pkg/webhooks/topic_validator.go
```diff
@@ -0,0 +1,232 @@
...
+
+	"github.com/go-logr/logr"
+
+	banzaicloudv1alpha1 "github.com/banzaicloud/koperator/api/v1alpha1"
+	banzaicloudv1beta1 "github.com/banzaicloud/koperator/api/v1beta1"
+	"github.com/banzaicloud/koperator/pkg/k8sutil"
+	"github.com/banzaicloud/koperator/pkg/kafkaclient"
+	"github.com/banzaicloud/koperator/pkg/util"
+)
+
```
stoader, line 37: Add `+kubebuilder:webhook` tag onto this type: https://book.kubebuilder.io/reference/markers/webhook.html

stoader (changes requested):

charts/kafka-operator/templates/operator-deployment-with-webhook.yaml
bartam1: According to this https://github.com/banzaicloud/koperator/pull/832#discussion_r920871898
I think problem can only happen if the operator is using another version of the KafkaCluster or KafkaTopic then what is specified in the validation admission webhook resource.

pkg/webhooks/kafkatopic_validator.go
```diff
@@ -0,0 +1,232 @@
...
+
+// checkKafka creates a Kafka admin client and connects to the Kafka brokers to check
+// whether the referred topic exists, and what are its properties
+func (s *KafkaTopicValidator) checkKafka(ctx context.Context, topic *banzaicloudv1alpha1.KafkaTopic,
+	cluster *banzaicloudv1beta1.KafkaCluster) (*field.Error, error) {
+	// retrieve an admin client for the cluster
+	broker, closeClient, err := s.NewKafkaFromCluster(s.Client, cluster)
+	if err != nil {
+		// Log as info to not cause stack traces when making CC topic
+		return nil, errors.WrapIf(err, fmt.Sprintf("%s: %s", cantConnectErrorMsg, topic.Spec.ClusterRef.Name))
```
stoader, line 151: You could use `errors.WrapIff` instead.

main.go
```diff
@@ -105,10 +103,6 @@ func main() {
 	flag.BoolVar(&certSigningDisabled, "disable-cert-signing-support", false, "Disable native certificate signing integration")
 	flag.IntVar(&maxKafkaTopicConcurrentReconciles, "max-kafka-topic-concurrent-reconciles", 10, "Define max amount of concurrent KafkaTopic reconciles")
 	flag.Parse()
-
-	podNamespace = os.Getenv("POD_NAMESPACE")
```
stoader, line 109: Since `POD_NAMESPACE` and `SERVICE_ACCOUNT` env vars are not used anymore anywhere I think this part can be removed as well: https://github.com/banzaicloud/koperator/blob/master/charts/kafka-operator/templates/operator-deployment-with-webhook.yaml#L201-L210

api/v1alpha1/kafkatopic_types.go
```diff
@@ -46,11 +46,10 @@ type KafkaTopicStatus struct {
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
 // +k8s:openapi-gen=true
-// +kubebuilder:webhook:failurePolicy="fail",sideEffects="None",name="kafkatopics.kafka.banzaicloud.io",path="/validate",mutating=false,resources={"kafkatopics"},verbs={"create","update"},groups={"kafka.banzaicloud.io"},versions={"v1alpha1"},admissionReviewVersions={"v1"}
-
 // KafkaTopic is the Schema for the kafkatopics API
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+
```
panyuenlau, line 52: Please delete this empty line so the kubebuilder would add the description "KafkaTopic is the Schema for the kafkatopics API" back to the CRD

pkg/webhooks/kafkacluster_validator.go
```diff
@@ -60,14 +65,24 @@ func (s KafkaClusterValidator) ValidateDelete(ctx context.Context, obj runtime.O
...
 			brokerNew := &kafkaClusterSpecNew.Brokers[k]
 			if brokerOld.Id == brokerNew.Id {
-				brokerConfigsOld, _ := brokerOld.GetBrokerConfig(*kafkaClusterSpecOld)
-				brokerConfigsNew, _ := brokerNew.GetBrokerConfig(*kafkaClusterSpecNew)
+				brokerConfigsOld, err := brokerOld.GetBrokerConfig(*kafkaClusterSpecOld)
+				if err != nil {
+					return nil, err
+				}
+				// checking broukerConfigGroup existence
+				if _, exists := kafkaClusterSpecNew.BrokerConfigGroups[brokerNew.BrokerConfigGroup]; !exists {
```
stoader, line 80: check for `brokerNew.BrokerConfigGroup != ""` as well as it's not mandatory to use broker config group

stoader (changes requested):

pkg/webhooks/errors.go
```diff
@@ -0,0 +1,45 @@
+// Copyright © 2019 Banzai Cloud
```
stoader, line 1: 2019 -> 2022

pkg/webhooks/kafkatopic_validator.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	defer closeClient()
+
+	existing, err := broker.GetTopic(topic.Spec.Name)
+	if err != nil {
+		return nil, errors.WrapIf(err, fmt.Sprintf("failed to list topics for kafka cluster: %s", topic.Spec.ClusterRef.Name))
```
stoader, line 156: use `errors.WrapIff`

stoader (changes requested):

main.go
bartam1: Yes, It can be removed. There is a chance for using it in the future. Do you think it should be removed?

pkg/webhooks/kafkacluster_validator.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func (s KafkaClusterValidator) ValidateUpdate(ctx context.Context, oldObj, newObj runtime.Object) error {
+	var allErrs field.ErrorList
+	kafkaClusterOld := oldObj.(*banzaicloudv1beta1.KafkaCluster)
+	kafkaClusterNew := newObj.(*banzaicloudv1beta1.KafkaCluster)
+	log := s.Log.WithValues("name", kafkaClusterNew.GetName(), "namespace", kafkaClusterNew.GetNamespace())
+	fieldErr, err := checkBrokerStorageRemoval(&kafkaClusterOld.Spec, &kafkaClusterNew.Spec)
+	if err != nil {
+		log.Error(err, errorDuringValidationMsg)
```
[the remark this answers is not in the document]
bartam1: Error will reported back to the user who want to add the kafkacluster CR into the kubernetes cluster.
I think its better to record it into the koperator logs also. Administrator can see there were problems. Also when later MCOM is used and the koperator log messages will be processed somehow, better to log these kind of errors also in the koperator logs.

api/v1alpha1/kafkatopic_types.go
bartam1: Thank you for pointing on this. I really didn't know why the CRD changed.

pkg/webhooks/kafkacluster_validator.go
bartam1: Good catch :)

config/base/webhook/manifests.yaml
```diff
@@ -11,7 +11,7 @@ webhooks:
     service:
       name: webhook-service
       namespace: system
-      path: /validate
+      path: /validate-kafka.banzaicloud-io-v1alpha1-kafkatopic
```
pregnor, line 14: **Question**: wan't this intended to be `kafka-banzaicloud` like it is for the `kafkacluster` validation endpoint?

```suggestion
      path: /validate-kafka-banzaicloud-io-v1alpha1-kafkatopic
```

pkg/webhooks/topic_validator.geso
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package webhook
+package webhooks
```
pregnor, line 15: **Question**: Why was this file's extension changed from `.go` to `.geso`? Was that intentional? I can see that some of the errors have been moved, had this meant to be a working copy which was supposed to be removed?

config/base/webhook/manifests.yaml
pregnor: You should regenerate the manifests to fix this based on the fixed path in the code.

pregnor: This is still an issue in the latest state.

stoader (approved):

pregnor (approved):
LGTM

PR #835, opened 2022-07-15
Support client IP filtering through loadBalancerSourceRanges in case istioingress

stoader:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no |
| New feature?    | yes |
| API breaks?     | no |
| Deprecations?   | no |

### What's in this PR?

Support client IP filtering through LoadBalancer by allowing users to specify the `loadBalancerSourceRanges` field for the corresponding Kubernetes service when Kafka external listeners use `istioingress` as ingress controller.

### Why?

`koperator` doesn't support client IP filtering for Kafka external listeners when those are exposed through `istioingress` ingress controller. Currently this is only supported when `envoy` is used and ingress controller.

### Checklist

- [x] Implementation tested


api/v1beta1/kafkacluster_types.go
```diff
@@ -280,8 +280,12 @@ type EnvoyConfig struct {
 	NodeSelector map[string]string   `json:"nodeSelector,omitempty"`
 	Tolerations  []corev1.Toleration `json:"tolerations,omitempty"`
 	// Annotations defines the annotations placed on the envoy ingress controller deployment
-	Annotations              map[string]string `json:"annotations,omitempty"`
-	LoadBalancerSourceRanges []string          `json:"loadBalancerSourceRanges,omitempty"`
+	Annotations map[string]string `json:"annotations,omitempty"`
+	// If specified and supported by the platform, this will restrict traffic through the cloud-provider
+	// load-balancer will be restricted to the specified client IPs. This field will be ignored if the
```
pregnor, line 285: **Recommended**: we **SHOULD** improve phrasing for clarity. We **SHOULD** also regenerate the CRDs, samples, etc. if changed.

```suggestion
	// If specified and supported by the platform, traffic through the 
	// cloud-provider load-balancer will be restricted to the specified client
	// IPs. This field will be ignored if the
```

api/v1beta1/kafkacluster_types.go
```diff
@@ -318,6 +322,12 @@ type IstioIngressConfig struct {
 	VirtualServiceAnnotations map[string]string   `json:"virtualServiceAnnotations,omitempty"`
 	// Envs allows to add additional env vars to the istio meshgateway resource
 	Envs []corev1.EnvVar `json:"envs,omitempty"`
+	// If specified and supported by the platform, this will restrict traffic through the cloud-provider
+	// load-balancer will be restricted to the specified client IPs. This field will be ignored if the
```
pregnor, line 326: **Recommended**: we **SHOULD** improve phrasing for clarity. We **SHOULD** also regenerate the CRDs, samples, etc. if changed.

```suggestion
	// If specified and supported by the platform, traffic through the 
	// cloud-provider load-balancer will be restricted to the specified client
	// IPs. This field will be ignored if the
```

pregnor (approved):
LGTM, left a suggestion on phrasing

pregnor (commented):
We should sync the CRDs and samples with the new type description through manifest generation.

pregnor (dismissed):

bartam1 (approved):

panyuenlau (approved):
thanks for adding this!

PR #850, opened 2022-08-11
Changes the necessary data fields for the clientSSLCertSecret and serverSSLCertSecret

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no |
| New feature?    | yes |
| API breaks?     | no |
| Deprecations?   | no |

### What's in this PR?
This PR makes it possible to add only keystore.jks, truststore.jks and the password  data fields in clientSSLCertSecret and serverSSLCertSecret for SSL listeners. Adding tls.crt, ca.crt, tls.key is not needed anymore. 

### Why?
To add certificates in different formats could be confusing for the users and error-prone.

### Checklist
- [x] Implementation tested


panyuenlau (approved):
Implementation looks good, just wondering if there is any doc or example config that we need to update to reflect this change

pkg/util/cert/certutil.go
```diff
@@ -315,3 +316,115 @@ func GenerateSigningRequestInPemFormat(priv *rsa.PrivateKey, commonName string,
 	signingReq := buf.Bytes()
 	return signingReq, err
 }
+
+func isSSLCertInJKS(data map[string][]byte) error {
```
pregnor, line 320: **Recommended**: We **SHOULD** only use `is...()` functions for predicates (functions with boolean return values), we can use check.../validate/etc. other function names for optional error returning functions for clarity purposes.

pkg/resources/cruisecontrol/cruisecontrol.go
```diff
@@ -148,25 +154,39 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
...
+
+func (r *Reconciler) getClientSecret() (*corev1.Secret, error) {
+	clientSecret := &corev1.Secret{}
+	// Use that secret as default which has autogenerated for clients by us
+	clientNamespacedName := types.NamespacedName{Name: fmt.Sprintf(pkicommon.BrokerControllerTemplate, r.KafkaCluster.Name), Namespace: r.KafkaCluster.Namespace}
+	if r.KafkaCluster.Spec.GetClientSSLCertSecretName() != "" {
+		clientNamespacedName = types.NamespacedName{Name: r.KafkaCluster.Spec.GetClientSSLCertSecretName(), Namespace: r.KafkaCluster.Namespace}
+	}
+
+	if err := r.Client.Get(context.TODO(), clientNamespacedName, clientSecret); err != nil {
```
pregnor, line 172: **Recommended**: We **SHOULD** prefer using [`context.Background()`](https://pkg.go.dev/context#Background) to [`context.TODO()`](https://pkg.go.dev/context#TODO) unless we are planning to refactor the function to pass down a context here, because otherwise this is a top level request context in this case.

pkg/resources/kafka/configmap.go
```diff
@@ -70,25 +70,26 @@ func (r *Reconciler) getConfigProperties(bConfig *v1beta1.BrokerConfig, id int32
...
-		if err := config.Set("cruise.control.metrics.reporter.ssl.keystore.location", clientKeystorePath+"/"+v1alpha1.TLSJKSKeyStore); err != nil {
-			log.Error(err, "setting cruise.control.metrics.reporter.ssl.keystore.location parameter in broker configuration resulted an error")
-		}
-		if err := config.Set("cruise.control.metrics.reporter.ssl.keystore.password", clientPass); err != nil {
-			log.Error(err, "setting cruise.control.metrics.reporter.ssl.keystore.password parameter in broker configuration resulted an error")
+
+		for k, v := range sslConfig {
+			if err := config.Set(fmt.Sprintf("cruise.control.metrics.reporter.%s", k), v); err != nil {
+				log.Error(err, fmt.Sprintf("setting cruise.control.metrics.reporter.%s parameter in broker configuration resulted an error", k))
+			}
```
pregnor, line 92: **Question**: This block looks similar to the one in `pkg/resources/cruisecontrol/configmap.go`, could we possibly extract this to a function to have a single place of definition for the logic or does it not make sense in this case?

Also `generateListenerSSLConfig()` looks somewhat of a generic variation of this, could that be used in some form or shape?

pregnor (commented):
LGTM, only reviewed the cert stuff on a glance level, don't have much semantic domain knowledge from the top of my head, be sure to manually test it for the original cases before merging.

pkg/resources/cruisecontrol/configmap.go
```diff
@@ -42,7 +42,7 @@ import (
 
 const MinLogDirSizeInMB = int64(1)
 
-func (r *Reconciler) configMap(clientPass, capacityConfig string, log logr.Logger) runtime.Object {
+func (r *Reconciler) configMap(clientPass string, capacityConfig string, log logr.Logger) runtime.Object {
```
Kuvesz, line 45: I'm not sure why this needed to change. It should be fine as `varName, varName2 type` as it was originally shouldn't it?

pkg/resources/cruisecontrol/cruisecontrol.go
```diff
@@ -148,25 +154,39 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
...
+		// We only return with ResourceNotReady (which is goin to retry after period time)
```
Kuvesz, line 173:
```suggestion
		// We only return with ResourceNotReady (which is going to retry after a period of time)
```

pkg/resources/kafka/configmap.go
```diff
@@ -70,25 +70,26 @@ func (r *Reconciler) getConfigProperties(bConfig *v1beta1.BrokerConfig, id int32
...
-		}
-		if err := config.Set("cruise.control.metrics.reporter.ssl.keystore.location", clientKeystorePath+"/"+v1alpha1.TLSJKSKeyStore); err != nil {
-			log.Error(err, "setting cruise.control.metrics.reporter.ssl.keystore.location parameter in broker configuration resulted an error")
-		}
-		if err := config.Set("cruise.control.metrics.reporter.ssl.keystore.password", clientPass); err != nil {
-			log.Error(err, "setting cruise.control.metrics.reporter.ssl.keystore.password parameter in broker configuration resulted an error")
+
+		for k, v := range sslConfig {
+			if err := config.Set(fmt.Sprintf("cruise.control.metrics.reporter.%s", k), v); err != nil {
+				log.Error(err, fmt.Sprintf("setting cruise.control.metrics.reporter.%s parameter in broker configuration resulted an error", k))
```
Kuvesz, line 91:
```suggestion
				log.Error(err, fmt.Sprintf("setting cruise.control.metrics.reporter.%s parameter in broker configuration resulted in an error", k))
```

Kuvesz (approved):
Looks good aside from some minor stuff, like spelling.

pkg/resources/cruisecontrol/configmap.go
pregnor: Yeah, it is a convention to shorten the function signature by omitting the continuously recurring type, but it does have a clarity reason to have the types stated explicitly, I would say it depends on personal taste.

pkg/resources/kafka/configmap.go
bartam1: This is not completely similar here we talking about cruise.control.metrics.reporter.%s. The sslConfig map variable can be used multiple times or with a variable in the  config.Set(fmt.Sprintf("cruise.control.metrics.reporter.%s", k), v) -->  config.Set(fmt.Sprintf(cruiseControlConfigString, k), v).

pkg/resources/cruisecontrol/configmap.go
bartam1: It has been changed because I put another variable between them but later I removed it.

bartam1:
> Implementation looks good, just wondering if there is any doc or example config that we need to update to reflect this change

I have changed the documentations everywhere what I found. Here also: https://github.com/banzaicloud/koperator-docs/pull/69

pregnor (approved):
LGTM, make sure we manual test the last state before merging.

panyuenlau (approved):

PR #851, opened 2022-08-12
Add CruiseControlOperation type and custom resource definition

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no |
| New feature?    | yes |
| API breaks?     | no |
| Deprecations?   | no |

### What's in this PR?
Introduce a new kind of custom resource whose name is CruiseControlOperation

### Why?
With the help of this new custom resource, cruise control tasks can be managed in a better way

### Checklist

- [x] Implementation tested


api/v1alpha1/common_types.go
```diff
@@ -32,8 +32,13 @@ type ClusterReference struct {
 	Name      string `json:"name"`
 	Namespace string `json:"namespace,omitempty"`
 }
+type CruiseControlTaskOperation string
```
panyuenlau, line 35: Please add a comment about what this exported string is about

api/v1alpha1/common_types.go
```diff
@@ -32,8 +32,13 @@ type ClusterReference struct {
...
 
 const (
+	// OperationAddBroker means a Cruise Control add_broker operation
+	OperationAddBroker CruiseControlTaskOperation = "addbroker"
+	// OperationRemoveBroker means a Cruise Control remove_broker operation
+	OperationRemoveBroker CruiseControlTaskOperation = "removebroker"
```
panyuenlau, line 41: Shouldn't the const variables be `"addBroker"` and `"removeBroker"`?

api/v1alpha1/cruisecontroloperation_types.go
```diff
@@ -0,0 +1,80 @@
+// Copyright © 2019 Banzai Cloud
```
panyuenlau, line 1:
```suggestion
// Copyright © 2022 Banzai Cloud
```

api/v1alpha1/cruisecontroloperation_types.go
```diff
@@ -0,0 +1,80 @@
...
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []CruiseControlOperation `json:"items"`
+}
+
+// CruiseControlOperationSpec defines the desired state of CruiseControlOperation.
+type CruiseControlOperationSpec struct {
+	// FailurePolicy defines how failed downscale operations should be handled.
+	// When it is "retry" the koperator re-executes the failed task in every 30 sec.
+	// When it is "ignore" the koperator handles the failed task as completed.
```
panyuenlau, line 47: I think it is better to specify where this default "30 seconds" comes from in this comment

```suggestion
	// When it is "retry", the koperator re-executes the failed task in every 30 seconds (explanation about where this default value comes from).
	// When it is "ignore", the koperator handles the failed task as completed.
```

api/v1alpha1/cruisecontroloperation_types.go
```diff
@@ -0,0 +1,80 @@
...
+	// +kubebuilder:validation:Enum=ignore;retry
+	// +optional
+	ErrorPolicy ErrorPolicyType `json:"errorPolicy,omitempty"`
+}
+
+type ErrorPolicyType string
```
panyuenlau, line 53: A short comment about this exported string

panyuenlau (changes requested):

api/v1alpha1/cruisecontroloperation_types.go
```diff
@@ -0,0 +1,80 @@
...
+// CruiseControlOperationList contains a list of CruiseControlOperation.
+type CruiseControlOperationList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []CruiseControlOperation `json:"items"`
+}
+
+// CruiseControlOperationSpec defines the desired state of CruiseControlOperation.
+type CruiseControlOperationSpec struct {
+	// FailurePolicy defines how failed downscale operations should be handled.
```
pregnor, line 45: **Required**: The comment calls the field `FailurePolicy`, but the actual field is `ErrorPolicy`, this taints the manifest generation as well causing discrepancy between the field name and it's description/documentation which could confuse the user.

```suggestion
	// ErrorPolicy defines how failed downscale operations should be handled.
```

api/v1alpha1/cruisecontroloperation_types.go
pregnor: Such as the following:

```suggestion
// ErrorPolicyType defines methods of handling CruiseControlOperation errors.
type ErrorPolicyType string
```

pregnor (changes requested):

api/v1alpha1/common_types.go
bartam1: I think it would be better to change this to add_broker, remove_broker, ... like here: https://github.com/linkedin/cruise-control/wiki/REST-APIs#post-requests. Later it can be changed. Im not sure yet.

api/v1alpha1/cruisecontroloperation_types.go
bartam1: This is documentation, we should not write here where it comes in the codebase.
At first let's try with 30s  I think it is a good default value.

Makefile
```diff
@@ -123,6 +123,7 @@ manifests: bin/controller-gen
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role paths="./controllers/..." output:rbac:artifacts:config=./config/base/rbac
 	## Regenerate CRDs for the helm chart
 	echo "{{- if .Values.crd.enabled }}" > $(HELM_CRD_PATH)
+	cat config/base/crds/kafka.banzaicloud.io_cruisecontroloperations.yaml >> $(HELM_CRD_PATH)
```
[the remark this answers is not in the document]
bartam1: New CRD is very rarely added. I don't think it's worth the effort

config/base/webhook/manifests.yaml
```diff
@@ -11,36 +11,36 @@ webhooks:
...
     - kafka.banzaicloud.io
     apiVersions:
-    - v1beta1
+    - v1alpha1
     operations:
+    - CREATE
     - UPDATE
     resources:
-    - kafkaclusters
+    - kafkatopics
```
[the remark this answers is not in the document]
bartam1: Yes, I also noticed this. I cant do much about it :)

api/v1alpha1/common_types.go
bartam1: I have changed to add_broker, remove_broker , and  I have added a new one "rebalance"

api/v1alpha1/cruisecontroloperation_types.go
panyuenlau: Well, this is where the CRD is generated from, right? I don't see why we should not add a short note about where this value comes from. If not here, then where?

panyuenlau: I wasnt asking for a detailed explanation, I was just thinking about something like `"every 30 seconds (by default)"`

api/v1alpha1/common_types.go
pregnor: I agree with using the CC terminology here to avoid unnecessary conversion requirements between enums/strings on component boundaries.

config/samples/example-cruisecontroloperation.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kafka.banzaicloud.io/v1alpha1
+kind: CruiseControlOperation
+metadata:
+  name: example-cruisecontroloperation
+  namespace: kafka
+spec:
+  errorPolicy: retry
+
```
pregnor, line 8: **Note**: this looks weird that GitHub states no new line at the end of the file, while line 8 is supposed to be that.
I think your editor added spaces to the new line when you hit return and that is the reason.
I can let this go, because the main reason is to have a blank line if you are reading from a terminal emulator which is fulfilled.

pregnor (approved):

api/v1alpha1/common_types.go
panyuenlau: thanks for the info @bartam1

api/v1alpha1/cruisecontroloperation_types.go
bartam1: Ohh OK. I thought a reference to a place somewhere in the codebase

pregnor (approved):

Kuvesz (approved):

panyuenlau (approved):

pregnor (approved):
LGTM

Issue #829, opened 2022-07-08
Cruise Control's remove_broker task completedWithError handling in Koperator


bartam1:
### Is your feature request related to a problem? Please describe.
When a Cruise Control task is ended up with completedWithError result then Koperator takes it as a succeeded operation.  https://github.com/banzaicloud/koperator/blob/97bdc71b10ef196e51d982c2ace9c4d85a77f133/controllers/cruisecontroltask_types.go#L106  When a downscale is succeded the Koperator removes the broker pod, PVC and broker configmap from the kubernetes cluster. This can lead to different problems such as data loss or missing replicas. Cruise Control  task can be ended up with completedWithError result in two cases.  
- Case (1) When CC task cannot be executed at all. (HTTP response 500 code)
_e.g:_
`< HTTP/1.1 500 Internal Server Error
{"errorMessage":"Error processing GET request \u0027/user_tasks\u0027 due to: \u0027There are already 1 active user tasks, which has reached the servlet capacity.\u0027.","version":1,"stackTrace":"java.lang.RuntimeException: There are already 1 active user tasks, which has reached the servlet capacity.\n\tat com.linkedin.kafka.cruisecontrol.servlet.UserTaskManager.ins`
OR
`< HTTP/1.1 500 Internal Server Error {"version":1,"stackTrace":"java.util.concurrent.ExecutionException: com.linkedin.kafka.cruisecontrol.exception.OptimizationFailureException: [RackAwareDistributionGoal] Cannot remove Replica[isLeader\u003dfalse,rack\u003dkafka-0.kafka.svc.cluster.local,broker\u003d0,TopicPartition\u003dexample-topic-2,origBroker\u003d0,isOriginalOffline\u003dtrue,isCurrentOffline\u003dtrue] from DEAD broker 0 (has 6 replicas). Add at least 1 broker. Add at least 1 broker.\n\tat java.base/java.util.concurrent.CompletableFuture.reportGet(Unknown Source)\n\tat java.base/java.util.concurrent.CompletableFuture`
OR
`< HTTP/1.1 500 Internal Server Error "errorMessage":"Error processing POST request \u0027/remove_broker\u0027 due to: \u0027com.linkedin.kafka.cruisecontrol.exception.KafkaCruiseControlException: java.lang.IllegalArgumentException: Broker 1 does not exist`

- Case (2) When the task has been executed but later will be interrupted unexpectedly (HTTP response 200 code). In this case, there will not be any error message.

[...]

PR #854, opened 2022-08-25
CruiseControlOperation controller

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no |
| New feature?    | yes |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets |  mentioned in #829  |

### What's in this PR?
New controller with finalizer for CruiseControlOperation custom resources.
The reconciler handles Cruise Control tasks based on the CruiseControlOperation properties.

- When a new CruiseControlOperation CR is added and there is no running task at the CC side the reconciler executes the currenTask in the status and updates it based on the API HTTP response
- When the CruiseControlOperation errorPolicy is retry and there is no pause="true" annotation on the resource and CC user task is finished with completedWithError the reconciler retires the task every 30sec and updates  the  CruiseControlOperation status (Failed task should be restarted when currentTime >= currentTask.Started + 30s)
-   Finalizer stops the execution of the running task by calling the POST /kafkacruisecontrol/stop_proposal_execution endpoint

### Why?
With this controller, the failed Cruise Control user tasks can be handled better.
In the future users can execute Cruise Control tasks declaratively.   

### Checklist

- [x] Implementation tested


controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,445 @@
...
+	if err != nil {
+		return requeueWithError(log, "couldn't get kafka cluster reference", err)
+	}
+
+	kafkaCluster := &banzaiv1beta1.KafkaCluster{}
+	err = r.Get(ctx, kafkaClusterRef, kafkaCluster)
+	if err != nil {
+		if apiErrors.IsNotFound(err) {
+			if !currentCCoperation.ObjectMeta.DeletionTimestamp.IsZero() {
+				log.Info("Cluster is gone already, remove finalizer")
```
panyuenlau, line 111:
```suggestion
				log.Info("kafka cluster has been removed, now removing finalizer from CruiseControlOperation")
```

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,445 @@
...
+				controllerutil.RemoveFinalizer(currentCCoperation, cruiseControlOperationFinalizer)
+				if err := r.Update(ctx, currentCCoperation); err != nil {
+					return requeueWithError(log, "failed to remove finalizer from CruiseControlOperation", err)
+				}
+			}
+			return reconciled()
+		}
+		return requeueWithError(log, "failed to lookup referenced cluster", err)
```
panyuenlau, line 119:
```suggestion
		return requeueWithError(log, "failed to lookup referenced kafka cluster", err)
```

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,445 @@
...
+	}
+
+	//Adding finalizer
+	if err := r.addFinalizer(ctx, currentCCoperation); err != nil {
+		return requeueWithError(log, "finalizer could not be added", err)
```
panyuenlau, line 124:
```suggestion
		return requeueWithError(log, "failed to add finalizer to CruiseControlOperation", err)
```

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,445 @@
...
+	if !r.Scaler.IsUp() {
+		log.Info("requeue event as Cruise Control is not available (yet)")
+		return requeueAfter(defaultRequeueIntervall)
+	}
+
+	// Filtering out CruiseControlOperation by kafka cluster ref
+	var ccOperationsKafkaClusterFiltered []*banzaiv1alpha1.CruiseControlOperation
+	for i := range ccOperationListClusterWide.Items {
+		operation := &ccOperationListClusterWide.Items[i]
+		ret, err := getKafkaClusterReference(operation)
```
panyuenlau, line 147: I guess this variable is meant to be named as `ref` rather than "ret"?

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,445 @@
...
+		if !reflect.DeepEqual(ccOperations[i].Status, ccOperationsCopy[i].Status) {
+			if err := r.Status().Update(ctx, ccOperations[i]); err != nil {
+				return errors.WrapIfWithDetails(err, "could not update CruiseControlOperation status", "name", ccOperations[i].GetName(), "namespace", ccOperations[i].GetNamespace())
+			}
+		}
+	}
+	return nil
+}
+
+func isWaitingforFinalize(ccOperation *banzaiv1alpha1.CruiseControlOperation) bool {
```
panyuenlau, line 425:
```suggestion
func isWaitingForFinalize(ccOperation *banzaiv1alpha1.CruiseControlOperation) bool {
```

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,445 @@
...
+		ccOperationsCopy = append(ccOperationsCopy, copy)
+	}
+
+	taskResultsByID := make(map[string]*scale.Result, len(tasks))
+	for _, task := range tasks {
+		taskResultsByID[task.TaskID] = task
+	}
+	for i := range ccOperations {
+		ccOperation := ccOperations[i]
+		log.Info("taskres current", "res", taskResultsByID[ccOperation.GetCurrentTaskID()])
```
panyuenlau, line 409: Is this meant to be `"current tasks"`? Also, I don't think the key "res" is clear enough when it is in the log message where we don't have context of the source code

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,445 @@
...
+		}
+		task.Started = &v1.Time{Time: startTime}
+	}
+	// We only have summary after execution
+	if isAfterExecution {
+		task.Summary = parseSummary(res.Result)
+		task.ErrorMessage = res.Err
+	}
+	// This is necessary because of a bug in CC (when a user task resulted in error later when get usertasks status from CC the corresponding task state is Active and not CompletedWithError...)
+	//if task.State != banzaiv1beta1.CruiseControlTaskCompletedWithError && task.State != banzaiv1beta1.CruiseControlTaskCompleted {
```
panyuenlau, line 382: According to the comment here, the if check is needed, but it is currently commented out

panyuenlau (changes requested):

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,445 @@
...
+	cruiseControlOperationFinalizer    = "finalizer.cruisecontroloperations.kafka.banzaicloud.io"
+	ccOperationForFinalize             = "ccOperationStopExecution"
+	ccOperationFirstExecution          = "ccOperationFirstExecution"
+	ccOperationRetryExecution          = "ccOperationRetryExecution"
+	ccOperationInProgress              = "ccOperationInProgress"
+)
+
+var (
+	defaultRequeueIntervall                                                   = 10
+	executionPriorityMap    map[banzaiv1alpha1.CruiseControlTaskOperation]int = map[banzaiv1alpha1.CruiseControlTaskOperation]int{
```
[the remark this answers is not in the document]
Kuvesz: This should get caught by the linter. Wonder why it's not working.

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,446 @@
...
+		}
+		return requeueWithError(log, "failed to lookup referenced cluster", err)
+	}
+
+	//Adding finalizer
+	if err := r.addFinalizer(ctx, currentCCoperation); err != nil {
+		return requeueWithError(log, "finalizer could not be added", err)
+	}
+
+	// We only get scaler when we have not had mocked one for test
```
Kuvesz, line 127:
```suggestion
	// We only get a scaler when we have not had mocked one for testing
```

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,446 @@
...
+		return requeueAfter(defaultRequeueIntervall)
+	}
+
+	// Filtering out CruiseControlOperation by kafka cluster ref
+	var ccOperationsKafkaClusterFiltered []*banzaiv1alpha1.CruiseControlOperation
+	for i := range ccOperationListClusterWide.Items {
+		operation := &ccOperationListClusterWide.Items[i]
+		ret, err := getKafkaClusterReference(operation)
+		if err != nil {
+			log.Info(err.Error())
```
Kuvesz, line 149: Why are we only logging the error here? Is it a valid config that a CC operation doesn't have a KafkaClusterReference? If so I guess this is fine, if not we should return an error here I think.

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,446 @@
...
+		// Requeue becuse we can't do more
+		return requeueAfter(defaultRequeueIntervall)
+	}
+
+	log.Info("executing Cruise Control task", "operation", ccOperationExecution.GetCurrentTaskOp(), "parameters", ccOperationExecution.GetCurrentTaskParameters())
+	// Executing operation
+	cruseControlTaskResult, err := r.executeOperation(ccOperationExecution)
+
+	if err != nil {
+		log.Error(err, "Cruise Control task execution got error", "operation", ccOperationExecution.GetCurrentTaskOp(), "parameters", ccOperationExecution.GetCurrentTaskParameters())
```
Kuvesz, line 197:
```suggestion
		log.Error(err, "Cruise Control task execution got an error", "operation", ccOperationExecution.GetCurrentTaskOp(), "parameters", ccOperationExecution.GetCurrentTaskParameters())
```

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,446 @@
...
+}
+
+func selectOperationForExecution(ccOperationQueueMap map[string][]*banzaiv1alpha1.CruiseControlOperation) *banzaiv1alpha1.CruiseControlOperation {
+	// SELECTING OPERATION FOR EXECUTION
+	var ccOperationExecution *banzaiv1alpha1.CruiseControlOperation
+	// First prio: execute the finalize task
+	switch {
+	case len(ccOperationQueueMap[ccOperationForFinalize]) > 0:
+		ccOperationExecution = ccOperationQueueMap[ccOperationForFinalize][0]
+		ccOperationExecution.GetCurrentTask().Operation = banzaiv1alpha1.OperationStopExecution
```
Kuvesz, line 287: Not quite sure but can it happen that a Finalize task is already running and we stop that to start another?

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,446 @@
...
+	// Second prio: execute add_broker operation
+	case len(ccOperationQueueMap[ccOperationFirstExecution]) > 0 && ccOperationQueueMap[ccOperationFirstExecution][0].GetCurrentTaskOp() == banzaiv1alpha1.OperationAddBroker:
+		ccOperationExecution = ccOperationQueueMap[ccOperationFirstExecution][0]
+	// Third prio: execute failed task
+	case len(ccOperationQueueMap[ccOperationRetryExecution]) > 0:
+		// When default backoff duration elapsed we retry
```
Kuvesz, line 293:
```suggestion
		// When the default backoff duration elapsed we retry
```

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,446 @@
...
+	if res == nil || operation.GetCurrentTask() == nil {
+		return nil
+	}
+	task := operation.GetCurrentTask()
+
+	if (task.State == banzaiv1beta1.CruiseControlTaskCompleted || task.State == banzaiv1beta1.CruiseControlTaskCompletedWithError) && task.Finished == nil {
+		task.Finished = &v1.Time{Time: time.Now()}
+	}
+
+	// Add the failed task into the status.failedTasks slice only when the update is happened after executing task
```
Kuvesz, line 354:
```suggestion
	// Add the failed task into the status.failedTasks slice only when the update happened after executing the task
```

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,446 @@
...
+		operation.Status.NumberOfRetries += 1
+		operation.Status.ErrorPolicy = banzaiv1alpha1.ErrorPolicyRetry
+		task.Finished = nil
+		task.State = ""
+		task.Started = nil
+	}
+
+	task.ID = res.TaskID
+	if task.Started == nil {
+		//Sat, 27 Aug 2022 12:22:21 GMT
```
Kuvesz, line 370: Is this an example? If it is could we write `Example:` before it? Also a space should be added after the //.

Kuvesz (changes requested):
There are some things I don't quite understand and some small corrections that should happen but overall looks like it's very well thought through.

api/v1alpha1/cruisecontroloperation_types.go
```diff
@@ -93,3 +95,89 @@ type CruiseControlTask struct {
 func init() {
 	SchemeBuilder.Register(&CruiseControlOperation{}, &CruiseControlOperationList{})
 }
+
+func (o *CruiseControlOperation) GetCurrentTask() *CruiseControlTask {
+	if o != nil {
+		return o.Status.CurrentTask
```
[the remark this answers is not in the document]
bartam1: Status is not pointer type. https://github.com/banzaicloud/koperator/blob/790f39311e37735ebb8b21a5a98da0d7f45afb8e/api/v1alpha1/cruisecontroloperation_types.go#L43. No need to check it.

controllers/cruisecontroloperation_controller.go
bartam1: Ref is better, Thanks.

pkg/scale/scale.go
```diff
@@ -115,7 +142,6 @@ func (cc *cruiseControlScaler) IsUp() bool {
 func (cc *cruiseControlScaler) GetUserTasks(taskIDs ...string) ([]*Result, error) {
 	req := &api.UserTasksRequest{
 		UserTaskIDs: taskIDs,
-		Entries:     100,
```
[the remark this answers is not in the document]
bartam1: When getting user tasks from Cruise Control the tasks aren't ordered by creation time. When there are more than 100 user tasks sometimes you don't get the latest task state... but when you get all of the tasks (history length is limited by a configuration setting), you get it. You can also get user task status by specifying ID. Maybe in the future, if needed I will modify this part to get user tasks by id not all at one.

pkg/scale/mock_scale.go
```diff
@@ -0,0 +1,303 @@
+// Copyright © 2022 Banzai Cloud
```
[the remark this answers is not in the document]
bartam1: Good question. In the koperator repo everywhere is banzai cloud. I think we should change all of them.

api/v1alpha1/cruisecontroloperation_types.go
```diff
@@ -93,3 +95,89 @@ type CruiseControlTask struct {
...
+func (o *CruiseControlOperation) IsWaitingForRetryExecution() bool {
+	if (!o.IsPaused() && !o.IsErrorPolicyIgnore()) &&
+		o.GetCurrentTaskState() == v1beta1.CruiseControlTaskCompletedWithError && o.GetCurrentTaskID() != "" {
+		return true
+	}
+	return false
+}
+
+func (o *CruiseControlOperation) IsReadyForRetryExecution() bool {
+	return o.IsWaitingForRetryExecution() && o.GetCurrentTask().Finished != nil && o.GetCurrentTask().Finished.Add(time.Second*DefaultRetryBackOffDurationSec).Before(time.Now())
```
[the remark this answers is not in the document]
bartam1: It is tricky but this is the right. finishtime + 30s before nowtime so in time line we passed the finishtime+30

pkg/scale/mock_scale.go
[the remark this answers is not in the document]
Kuvesz: Yeah, maybe in a different PR later on I guess.

controllers/cruisecontroloperation_controller.go
bartam1: This is only for debugging for myself. I will remove it.

controllers/cruisecontroloperation_controller.go
bartam1: After I found that the problem caused by the get user task 100 limit what I have removed. 
https://github.com/banzaicloud/koperator/blob/b5df80b9dec3efc8ee1b6085af947ee074c56c87/pkg/scale/scale.go#L118
I will remove this part. If the problem does not come back.

api/v1alpha1/cruisecontroloperation_types.go
```diff
@@ -93,3 +95,89 @@ type CruiseControlTask struct {
...
+func (o *CruiseControlOperation) IsPaused() bool {
+	return o.GetLabels()["pause"] == "true"
+}
+
+func (o *CruiseControlOperation) IsErrorPolicyIgnore() bool {
+	return o.Spec.ErrorPolicy == ErrorPolicyIgnore
+}
+
+func (o *CruiseControlOperation) IsWaitingForRetryExecution() bool {
+	if (!o.IsPaused() && !o.IsErrorPolicyIgnore()) &&
```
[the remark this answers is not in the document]
bartam1: This is true when an operation needs to be retried. When the errorPolicy is ignore than it means success and doesn't need to retry.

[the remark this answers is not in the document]
bartam1: Sure. OK. I'm going to change.

controllers/cruisecontroloperation_controller.go
bartam1: In the future when the user does not provide the kafkaclusterRef label for an CCoperation then this would break the code further execution in return case. In the log the user can see  what is the problem his own CCoperation.

controllers/cruisecontroloperation_controller.go
bartam1: Good question. I did not think about this case. It should be tested.

controllers/cruisecontroloperation_controller.go
bartam1: This can be removed.

controllers/cruisecontroloperation_controller.go
bartam1: I have tested. Stop proposal execution runs instantly and the interrupted task will get completedWithError state.
Not possible to interrupt the finalizing task.

controllers/tests/cruisecontroloperation_controller_test.go
```diff
@@ -0,0 +1,491 @@
...
+					Name:      opName1,
+				}, &operation)
+				if err != nil {
+					return ""
+				}
+				return operation.GetCurrentTaskState()
+			}, 10*time.Second, 500*time.Millisecond).Should(Equal(v1beta1.CruiseControlTaskCompleted))
+		})
+	})
+	When("add_broker operation is finished with completedWithError and 30s has not epalsed", func() {
```
panyuenlau, line 99:
```suggestion
	When("add_broker operation is finished with completedWithError and 30s has not elapsed", func() {
```

controllers/tests/cruisecontroloperation_controller_test.go
```diff
@@ -0,0 +1,491 @@
...
+				}, &operation2)
+				if err != nil {
+					return false
+				}
+
+				return operation1.Status.NumberOfRetries == 0 && operation2.GetCurrentTaskState() == v1beta1.CruiseControlTaskCompleted
+			}, 40*time.Second, 500*time.Millisecond).Should(BeTrue())
+		})
+	})
+	When("there is a new remove_roker and an errored remove_broker operation", func() {
```
panyuenlau, line 254:
```suggestion
	When("there is a new remove_broker and an errored remove_broker operation", func() {
```

controllers/tests/cruisecontroloperation_controller_test.go
```diff
@@ -0,0 +1,491 @@
...
+				}, &operation2)
+				if err != nil {
+					return false
+				}
+
+				return operation1.Status.NumberOfRetries > 1 && (operation2.GetCurrentTaskState() == "" && operation2.GetCurrentTask().ID != "1")
+			}, 10*time.Second, 500*time.Millisecond).Should(BeTrue())
+		})
+	})
+	When("there is a new remove_roker and an errored remove_broker operation with ignore errorpolicy", func() {
```
panyuenlau, line 303:
```suggestion
	When("there is a new remove_broker and an errored remove_broker operation with ignore error policy", func() {
```

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,473 @@
...
+		}
+	}
+	return nil
+}
+
+func isWaitingForFinalize(ccOperation *banzaiv1alpha1.CruiseControlOperation) bool {
+	if ccOperation.IsCurrentTaskRunning() && !ccOperation.ObjectMeta.DeletionTimestamp.IsZero() && controllerutil.ContainsFinalizer(ccOperation, cruiseControlOperationFinalizer) {
+		return true
+	}
+	return false
```
panyuenlau, line 457: same logic but shorter
```suggestion
	return ccOperation.IsCurrentTaskRunning() && !ccOperation.ObjectMeta.DeletionTimestamp.IsZero() && controllerutil.ContainsFinalizer(ccOperation, cruiseControlOperationFinalizer)
```

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,473 @@
...
+		task.HTTPRequest = res.RequestURL
+		task.HTTPResponseCode = &res.ResponseStatusCode
+	}
+
+	task.State = res.State
+
+	return nil
+}
+
+// updateActiveTasks updates the state of the CruiseControlOperation from the CruiseControlTasksAndStates instance by getting their
```
panyuenlau, line 410: Looks like `updateActiveTasks` is the old function name and it does't match the latest function name
```suggestion
// updateCurrentTasks updates the state of the CruiseControlOperation from the CruiseControlTasksAndStates instance by getting their
```

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,473 @@
...
+					return true
+				}
+				return false
+			},
+		})
+
+	return builder
+}
+
+func isFinalizeNeeded(operation *banzaiv1alpha1.CruiseControlOperation) bool {
```
panyuenlau, line 347:
```suggestion
func isFinalizerNeeded(operation *banzaiv1alpha1.CruiseControlOperation) bool {
```

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,473 @@
...
+		return requeueAfter(defaultRequeueIntervalInSeconds)
+	}
+
+	log.Info("executing Cruise Control task", "operation", ccOperationExecution.GetCurrentTaskOp(), "parameters", ccOperationExecution.GetCurrentTaskParameters())
+	// Executing operation
+	cruseControlTaskResult, err := r.executeOperation(ccOperationExecution)
+
+	if err != nil {
+		log.Error(err, "Cruise Control task execution got an error", "operation", ccOperationExecution.GetCurrentTaskOp(), "parameters", ccOperationExecution.GetCurrentTaskParameters())
+	}
```
panyuenlau, line 211: Question: why is it enough to just log the error? Shouldn't we perform other actions when it fails to execute the cruise control operation?

panyuenlau (commented):
Some more comments/questions 😶‍🌫️

controllers/cruisecontroloperation_controller.go
```diff
@@ -340,75 +358,85 @@ func getKafkaClusterReference(operation *banzaiv1alpha1.CruiseControlOperation)
 	}, nil
 }
 
-func updateResult(res *scale.Result, operation *banzaiv1alpha1.CruiseControlOperation, isAfterExecution bool) error {
-	if res == nil || operation.GetCurrentTask() == nil {
-		return nil
+func updateResult(log logr.Logger, res *scale.Result, operation *banzaiv1alpha1.CruiseControlOperation, isAfterExecution bool) error {
+	// This can happen rarely when the max cached completed user tasks reached
```
Kuvesz, line 362:
```suggestion
	// This can happen rarely when the max cached completed user tasks is reached
```

Kuvesz (commented):
This looks good to me once the stuff asked for by Darren is done, but I would appreciate someone with a lot more knowledge on Cruise Control taking a look at the logic.

controllers/cruisecontroloperation_controller.go
panyuenlau: thanks for the explanation, can you add this as a comment along with the code? Just to make things more clear about why logging the error is enough here

pkg/scale/utils.go
```diff
@@ -63,7 +63,7 @@ func cruiseControlURL(endpoint string, secure bool) string {
 	return fmt.Sprintf("%s://%s/%s", scheme, endpoint, "kafkacruisecontrol")
 }
 
-func stringSliceToMap(s []string) map[string]bool {
+func StringSliceToMap(s []string) map[string]bool {
```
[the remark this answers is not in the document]
bartam1: Im using it in the scale.go

controllers/cruisecontroloperation_controller.go
```diff
@@ -92,11 +92,20 @@ func (r *CruiseControlOperationReconciler) Reconcile(ctx context.Context, reques
...
+	// When the resource has been removed before reconcile.
 	if currentCCOperation == nil {
 		return reconciled()
 	}
 
+	//When the task is done we can remove the finalizer instantly thus we can return fast here.
+	if isFinalizerNeeded(currentCCOperation) && currentCCOperation.IsDone() {
+		controllerutil.RemoveFinalizer(currentCCOperation, cruiseControlOperationFinalizer)
+		if err := r.Update(ctx, currentCCOperation); err != nil {
+			return requeueWithError(log, "error is happened when removing finalizer", err)
```
Kuvesz, line 104:
```suggestion
			return requeueWithError(log, "error happened when removing finalizer", err)
```

Kuvesz (approved):
This looks good to me but I would be grateful for a review from someone more experienced with this.

controllers/tests/cruisecontroloperation_controller_test.go
```diff
@@ -0,0 +1,454 @@
...
+				Name: namespace,
+			},
+		}
+		kafkaClusterCRName = fmt.Sprintf("kafkacluster-%v", count)
+		kafkaCluster = createMinimalKafkaClusterCR(kafkaClusterCRName, namespace)
+	})
+
+	JustBeforeEach(func() {
+		By("creating namespace " + namespace)
+		err := k8sClient.Create(context.TODO(), namespaceObj)
```
pregnor, line 62: **Recommended**: We **SHOULD** use [`context.Background()`](https://pkg.go.dev/context#Background) in tests according to the package's suggestion.

Throughout the test files.

controllers/tests/cruisecontroloperation_controller_test.go
```diff
@@ -0,0 +1,454 @@
...
+				}, &operation2)
+				if err != nil {
+					return false
+				}
+
+				return operation1.Status.NumberOfRetries == 0 && operation2.GetCurrentTaskState() == v1beta1.CruiseControlTaskCompleted
+			}, 40*time.Second, 500*time.Millisecond).Should(BeTrue())
+		})
+	})
+})
```
pregnor, line 301: **Required**: We **MUST** add a test case for 3 different type new operations, one of each in a non-priority ordering and expect to execute them in priority order to test the sorting and selection logic.

That would convince me of the current logic working as expected faster than words.

(Also we should write unit tests for those sort and select functions if we can.)

pregnor (changes requested):
Make sure to retest before merging.

api/v1alpha1/cruisecontroloperation_types.go
```diff
@@ -93,3 +95,112 @@ type CruiseControlTask struct {
...
+	task.State = ""
+	task.Started = nil
+	task.ErrorMessage = ""
+	task.HTTPRequest = ""
+	task.HTTPResponseCode = nil
+	task.ID = ""
+	task.Summary = nil
+}
+
+func (o *CruiseControlOperation) GetCurrentTask() *CruiseControlTask {
```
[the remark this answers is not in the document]
bartam1: Thank you! Good to know. In kubernetes GetName() ... are used. Im not sure which would be the best. Follow the go pattern or follow the kubernetes :)

api/v1alpha1/cruisecontroloperation_types.go
```diff
@@ -93,3 +95,112 @@ type CruiseControlTask struct {
...
+func (o *CruiseControlOperation) IsWaitingForRetryExecution() bool {
+	if (!o.IsPaused() && o.IsErrorPolicyRetry()) &&
+		o.GetCurrentTaskState() == v1beta1.CruiseControlTaskCompletedWithError && o.GetCurrentTaskID() != "" {
+		return true
+	}
+	return false
+}
+
+func (o *CruiseControlOperation) IsReadyForRetryExecution() bool {
+	return o.IsWaitingForRetryExecution() && o.GetCurrentTask().Finished != nil && o.GetCurrentTask().Finished.Add(time.Second*DefaultRetryBackOffDurationSec).Before(time.Now())
```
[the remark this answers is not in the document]
bartam1: Yes, It is error-prone but Im using it bravely because of this check sooner: https://github.com/banzaicloud/koperator/blob/8ef9a29d6958b74fe64843563ced661b2713c49d/controllers/cruisecontroloperation_controller.go#L172
It cant be nil at this part.

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,491 @@
...
+	log := logr.FromContextOrDiscard(ctx)
+
+	userTaskIDs := make([]string, 0, len(ccOperations))
+	var ccOperationsCopy []*banzaiv1alpha1.CruiseControlOperation
+	for _, ccOperation := range ccOperations {
+		if ccOperation.GetCurrentTaskID() != "" {
+			userTaskIDs = append(userTaskIDs, ccOperation.GetCurrentTaskID())
+		}
+		copy := ccOperation.DeepCopy()
+		ccOperationsCopy = append(ccOperationsCopy, copy)
```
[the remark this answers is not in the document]
bartam1: There is no reason. I will fix that. Thanks!

api/v1alpha1/cruisecontroloperation_types.go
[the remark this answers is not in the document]
pregnor: TBH I'm heavily in favor of the Go by default.

Isn't K8s follow the Get... pattern where it does resource CRUD operations due to the different action verbs you can do from RBAC standpoint on resources? In that case Get makes sense, because it signals an HTTP GET which is fine to use Get prefix even in Go, it is just regular field accessors are ugly in Go with Get prefix and are considered unidiomatic.

api/v1alpha1/cruisecontroloperation_types.go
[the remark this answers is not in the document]
bartam1: I have fixed this. Thank you for pointing out this.

pkg/scale/types.go
```diff
@@ -14,27 +14,42 @@
...
-	Status() CruiseControlStatus
+	Status() (CruiseControlStatus, error)
 	GetUserTasks(taskIDs ...string) ([]*Result, error)
 	IsUp() bool
 	AddBrokers(brokerIDs ...string) (*Result, error)
+	AddBrokersWithParams(params map[string]string) (*Result, error)
+	RemoveBrokersWithParams(params map[string]string) (*Result, error)
+	RebalanceWithParams(params map[string]string) (*Result, error)
+	StopExecution() (*Result, error)
+	GetNumMonitoredWin() (float32, types.MonitorState, error)
```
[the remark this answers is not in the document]
bartam1: In the final implementation it is not needed at all. Im going to remove this function.

pkg/scale/types.go
```diff
@@ -14,27 +14,42 @@
...
-	StartedAt string
-	State     v1beta1.CruiseControlUserTaskState
-	Err       string
+	TaskID             string
+	StartedAt          string
+	ResponseStatusCode int
+	RequestURL         string
+	Result             *types.OptimizationResult
+	State              v1beta1.CruiseControlUserTaskState
+	Err                string
```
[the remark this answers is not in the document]
bartam1: This error message is going to the CruiseControlOperation.CurrentTask.ErrorMessage (string)
I think it would better to use the error interface here and later stringify.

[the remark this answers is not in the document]
pregnor: I agree.
I would always advise postponing destructurization (converting highly structured data into lower level of structured data or unstructured data) as far as we can put it off to avoid loosing information which might come in handy later.

pregnor (commented):
The implemented changes look good so far.

pkg/scale/mock_scale.go
```diff
@@ -0,0 +1,334 @@
...
```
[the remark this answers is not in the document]
bartam1: Yes it is generated by mockgen.

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,491 @@
...
+
+	banzaiv1alpha1 "github.com/banzaicloud/koperator/api/v1alpha1"
+	banzaiv1beta1 "github.com/banzaicloud/koperator/api/v1beta1"
+	"github.com/banzaicloud/koperator/pkg/scale"
+	"github.com/banzaicloud/koperator/pkg/util"
+)
+
+const (
+	defaultFailedTasksHistoryMaxLength = 50
+	cruiseControlOperationFinalizer    = "finalizer.cruisecontroloperations.kafka.banzaicloud.io"
```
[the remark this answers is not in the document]
bartam1: I think it would be nicer to use the shortening also for that one.

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,491 @@
...
+	// Sorting by operation type and by the k8s object creation time
+	for key := range ccOperationQueueMap {
+		sort.SliceStable(ccOperationQueueMap[key], func(i, j int) bool {
+			ccOperationQueue := ccOperationQueueMap[key]
+			if executionPriorityMap[ccOperationQueue[i].GetCurrentTaskOp()] < executionPriorityMap[ccOperationQueue[j].GetCurrentTaskOp()] {
+				return true
+			} else if executionPriorityMap[ccOperationQueue[i].GetCurrentTaskOp()] == executionPriorityMap[ccOperationQueue[j].GetCurrentTaskOp()] {
+				return ccOperationQueue[i].CreationTimestamp.Unix() < ccOperationQueue[j].CreationTimestamp.Unix()
+			}
+			return true
```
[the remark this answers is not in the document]
bartam1: Yep, good catch! It should be false at the end. I will make a unit test for this.

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,491 @@
...
+		case ccOperation.IsWaitingForRetryExecution():
+			ccOperationQueueMap[ccOperationRetryExecution] = append(ccOperationQueueMap[ccOperationRetryExecution], ccOperation)
+		case ccOperation.IsInProgress():
+			ccOperationQueueMap[ccOperationInProgress] = append(ccOperationQueueMap[ccOperationInProgress], ccOperation)
+		}
+	}
+
+	// Sorting by operation type and by the k8s object creation time
+	for key := range ccOperationQueueMap {
+		sort.SliceStable(ccOperationQueueMap[key], func(i, j int) bool {
```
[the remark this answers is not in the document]
bartam1: Thank you for this detailed explanation.

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,491 @@
...
+	}
+
+	//When the task is not in execution we can remove the finalizer
+	if isFinalizerNeeded(currentCCOperation) && !currentCCOperation.IsCurrentTaskRunning() {
+		controllerutil.RemoveFinalizer(currentCCOperation, cruiseControlOperationFinalizer)
+		if err := r.Update(ctx, currentCCOperation); err != nil {
+			return requeueWithError(log, err.Error(), err)
+		}
+		return reconciled()
+	}
```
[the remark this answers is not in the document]
bartam1: The logic here: we can remove finalizer from CCOperation when It is not running && finalization is needed (deleted)
when it is running then we stop the running task execution with a CC API call and after we can remove the finalizer in the next reconciliation.

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,491 @@
...
+		task.Finished = &v1.Time{Time: time.Now()}
+	}
+
+	// Add the failed task into the status.failedTasks slice only when the update is happened after executing the task
+	if isAfterExecution && task.Finished != nil && task.State == banzaiv1beta1.CruiseControlTaskCompletedWithError {
+		if len(operation.Status.FailedTasks) >= defaultFailedTasksHistoryMaxLength {
+			operation.Status.FailedTasks = append(operation.Status.FailedTasks[1:], *task)
+		} else {
+			operation.Status.FailedTasks = append(operation.Status.FailedTasks, *task)
+		}
```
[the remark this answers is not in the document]
bartam1: Thanks for the suggestion! I would leave it as it is.

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,491 @@
...
+		operation.Status.NumberOfRetries += 1
+		task.SetDefaults()
+	}
+
+	if isAfterExecution {
+		if task.Started == nil {
+			startTime, err := time.Parse(time.RFC1123, res.StartedAt)
```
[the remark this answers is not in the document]
bartam1: Good question! I have to look after to is it can be configured or not.

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,508 @@
...
+	if err != nil {
+		if apiErrors.IsNotFound(err) {
+			if !currentCCOperation.ObjectMeta.DeletionTimestamp.IsZero() {
+				log.Info("kafka cluster has been removed, now removing finalizer from CruiseControlOperation")
+				controllerutil.RemoveFinalizer(currentCCOperation, ccOperationFinalizerGroup)
+				if err := r.Update(ctx, currentCCOperation); err != nil {
+					return requeueWithError(log, "failed to remove finalizer from CruiseControlOperation", err)
+				}
+			}
+			return reconciled()
```
panyuenlau, line 130: Why do we just ignore the not found error here and return reconciled()?

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,508 @@
...
+		}
+	// Forth prio: execute the first element in the FirstExecutionQueue which is ordered by operation type and k8s creation timestamp
+	case len(ccOperationQueueMap[ccOperationFirstExecution]) > 0:
+		ccOperationExecution = ccOperationQueueMap[ccOperationFirstExecution][0]
+	}
+
+	return ccOperationExecution
+}
+
+// SetupCruiseControlWithManager registers cruise control controller to the manager
```
panyuenlau, line 351:
```suggestion
// SetupCruiseControlWithManager registers cruise control operation controller to the manager
```

panyuenlau (commented):
I don't really have more to add based on what have been mentioned by others, just a couple of questions / nits

controllers/cruisecontroloperation_controller.go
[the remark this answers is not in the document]
pregnor: Alternative suggestion:

```suggestion
		if len(operation.Status.FailedTasks) >= defaultFailedTasksHistoryMaxLength {
			operation.Status.FailedTasks = operation.Status.FailedTasks[1:]
		}
		operation.Status.FailedTasks = append(operation.Status.FailedTasks, *task)
```

What do you think?
(It is still better for my taste, but it resembles your approach more.)

controllers/cruisecontroloperation_controller.go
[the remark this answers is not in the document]
pregnor: Ok, then this is not just simply our choice, then I can let this go for now.

api/v1beta1/kafkacluster_types.go
```diff
@@ -966,7 +986,7 @@ func (cConfig *CruiseControlConfig) GetCCImage() string {
 	if cConfig.Image != "" {
 		return cConfig.Image
 	}
-	return "ghcr.io/banzaicloud/cruise-control:2.5.86"
+	return "ghcr.io/banzaicloud/cruise-control:2.5.101"
```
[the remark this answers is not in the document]
pregnor: Could you create an issue for this?
Then we can leave it as is for this PR and solve it later.

pregnor (commented):
Changes are looking good, couple comments left to be addressed.

Edit: went through my comments, unresolved things I didn't see fixed yet, commented everything resolved which were not marked outdated for easier review later.

pkg/scale/types.go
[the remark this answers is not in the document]
pregnor: You changed the references setting this but not the field type as I saw.

controllers/cruisecontroloperation_controller.go
```diff
@@ -0,0 +1,491 @@
...
+		return requeueAfter(defaultRequeueIntervalInSeconds)
+	}
+
+	// Filtering out CruiseControlOperation by kafka cluster ref and state
+	var ccOperationsKafkaClusterFiltered []*banzaiv1alpha1.CruiseControlOperation
+	for i := range ccOperationListClusterWide.Items {
+		operation := &ccOperationListClusterWide.Items[i]
+		ref, err := getKafkaClusterReference(operation)
+		if err != nil {
+			log.Info(err.Error())
```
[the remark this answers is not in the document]
pregnor: This was fixed in https://github.com/banzaicloud/koperator/pull/854/commits/8f3dadd1d7236af3bd5be917d4e069d985148632.

controllers/cruisecontroloperation_controller.go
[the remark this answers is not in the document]
pregnor: This was fixed in https://github.com/banzaicloud/koperator/pull/854/commits/ef3a31b054210a4555f64b93bf7a4066fda114b5

api/v1beta1/kafkacluster_types.go
```diff
@@ -268,6 +269,25 @@ type CruiseControlConfig struct {
...
+	// When TTLSecondsAfterFinished is specified, the created and finished (completed successfully or completedWithError and errorPolicy: ignore)
+	// cruiseControlOperation custom resource will be deleted after the given time elapsed.
+	// When it is 0 then the resource is going to be deleted instantly after the operation is finished.
+	// When it is not specified the resource is not going to be removed.
+	// Value can be only zero and positive integers.
+	// +kubebuilder:validation:Minimum=0
+	TTLSecondsAfterFinished *int `json:"ttlSecondsAfterFinished,omitempty"`
+}
+
+// GetTTLSecondsAfterFinished returns NIL when TTLSecondsAfterFinished is not specified otherwise it returns itself
```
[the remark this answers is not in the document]
bartam1: When CruiseControlOperationSpec is not specified then the TTLSecondsAfterFinished also is not specified. I agree with you this description is not the best. I will also change the function name to TTLSecondsAfterFinished()

controllers/cruisecontroloperation_controller.go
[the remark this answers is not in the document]
bartam1: I like this one. Thanks!

controllers/cruisecontroloperation_controller.go
bartam1: The referenced kafkaCluster CR has been deleted thus there is no Kafka cluster anymore we can remove the finalizer from the current CruiseControlOperation when it is under deletion and we are done with the current CCOperation. There is no execution needed. This is not an error. is my explanation reasonable?

api/v1beta1/kafkacluster_types.go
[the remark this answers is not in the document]
bartam1: https://jira-eng-sjc4.cisco.com/jira/browse/STB-858

controllers/cruisecontroloperation_controller.go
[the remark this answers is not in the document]
bartam1: I haven't found a configuration property yet for that.

controllers/tests/cruisecontroloperation_controller_test.go
bartam1: It is hard to do this kind of test because I cannot add multiple cruiseControlOperations in the same time.  I rather add a unit test for this part.

panyuenlau (approved):
Went through my comments, all of them have been addressed -> ✅

bartam1:
Thank you for the detailed reviews! I have re-test everything and It works well. I think Koperator is much better with this change.
Thank you all!

controllers/tests/cruisecontroloperation_controller_test.go
pregnor: I am satisfied with the unit tests. thanks.

pregnor (approved):
LGTM

PR #861, opened 2022-09-07
Refactor cruisecontroltask_controller for CruiseControlOperationController usage

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no |
| New feature?    | yes |
| API breaks?     | no |
| Deprecations?   | no |

### What's in this PR?
The cruisecontroltask_controller creates CruiseControlOperation CRs to execute CC user tasks.

### Why?
This way Cruise Control operations are more reliable.

### Checklist

- [ ] Implementation tested
- [x ] Error handling code meets the [guideline](https://github.com/banzaicloud/developer-guide/blob/master/docs/coding-style/error-handling-guide.md)
- [ x] Logging code meets the guideline
- [ x] User guide and development docs updated (if needed)


api/v1beta1/common_types.go
```diff
@@ -160,27 +154,19 @@ type IstioControlPlaneReference struct {
...
-	// CruiseControlTaskId holds info about the task id ran by CC
-	CruiseControlTaskId string `json:"cruiseControlTaskId,omitempty"`
-	// TaskStarted hold the time when the execution started
-	TaskStarted string `json:"TaskStarted,omitempty"`
-	// CruiseControlState holds the information about CC state
+	// CruiseControlState holds the information about graceful action state
 	CruiseControlState CruiseControlState `json:"cruiseControlState"`
-	// VolumeStates holds the information about the CC disk rebalance states and tasks
+	// CruiseControlOperationReference referencing to the created CruiseControlOperation to execute CC task
+	CruiseControlOperationReference *corev1.LocalObjectReference `json:"cruiseControlOperationReference,omitempty"`
```
amuraru, line 160: How are these external `CruiseControlOperation` CRs being managed? Say there are multiple of them created throughout the operation of kafka - how are the old CruiseControlOperation CRs deleted?

amuraru: Not reclaiming them would create a lot of waste in time. +1 for that flag - similar to Job resources they should be by defaukt deleted after a while since they completed successfully

bartam1: Dear @amuraru! Please let me know when you could test it

amuraru: Do you think it would be better to have something like:
ttlAfterCompleted similar to:
https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/#ttl-after-finished-controller


I am thinking at the usecase where these CRs would be monitored and give the external monitoring a chance to notice a completed operation before beibg deleted

bartam1: It can be solved. Logic also needed in the CruiseControlOperation_controller.

bartam1: I have implemented a new controller for this:
https://github.com/banzaicloud/koperator/pull/865

config/base/crds/kafka.banzaicloud.io_kafkaclusters.yaml
```diff
@@ -12303,6 +12303,17 @@ spec:
...
                   cruiseControlEndpoint:
                     type: string
+                  cruiseControlOperatonSpec:
+                    description: CruiseControlOperatonSpec specifies the configuration
+                      of the CruiseControlOperaton handling
+                    properties:
+                      ttlSecondsAfterFinished:
+                        description: TTLSecondsAfterFinished when is specified, the
+                          created and succeeded CruiseControlOperation custom resource
+                          will be deleted after the given time elapsed.
```
Kuvesz, line 12313:
```suggestion
                        description: TTLSecondsAfterFinished specifies when the
                          created and succeeded CruiseControlOperation custom resource
                          will be deleted in seconds. When not specified it will not be deleted.
```

config/base/crds/kafka.banzaicloud.io_kafkaclusters.yaml
```diff
@@ -19574,51 +19585,48 @@ spec:
                       description: GracefulActionState holds info about cc action
                         status
                       properties:
-                        TaskStarted:
-                          description: TaskStarted hold the time when the execution
-                            started
-                          type: string
+                        cruiseControlOperationReference:
+                          description: CruiseControlOperationReference referencing
+                            to the created CruiseControlOperation to execute CC task
```
Kuvesz, line 19590:
```suggestion
                          description: CruiseControlOperationReference references
                            the created CruiseControlOperation to execute a CC task
```

controllers/cruisecontroltask_controller.go
```diff
@@ -70,121 +81,218 @@ func (r *CruiseControlTaskReconciler) Reconcile(ctx context.Context, request ctr
...
 	if err != nil {
 		log.Error(err, "requeue event as updating state of active tasks failed")
 		return requeueAfter(DefaultRequeueAfterTimeInSec)
 	}
 
-	// Check if CruiseControl is ready as we cannot perform any operation until it is in ready state
-	if status, _ := scaler.Status(); status.InExecution() {
-		log.Info("updating status of Kafka Cluster and requeue event as Cruise Control is in execution")
-		if err := r.UpdateStatus(ctx, instance, tasksAndStates); err != nil {
-			log.Error(err, "failed to update Kafka Cluster status")
```
Kuvesz, line 94: May be stupid question: Is it okay to have this part removed? To me it seems kinda important.

controllers/cruisecontroltask_controller.go
```diff
@@ -70,121 +81,218 @@ func (r *CruiseControlTaskReconciler) Reconcile(ctx context.Context, request ctr
...
+				if !found {
+					unavailableBrokerIDs = append(unavailableBrokerIDs, task.BrokerID)
 				}
-			} else {
-				task.Err = "log dir is not available in Cruise Control"
 			}
 		}
+		if len(unavailableBrokerIDs) > 0 {
+			log.Info("requeue as there are offline broker log dirs for rebalance", "brokerIDs", unavailableBrokerIDs)
+			// This requeue is not necessary because the cruisecontrloperation controller retry the errored task
```
Kuvesz, line 190:
```suggestion
			// This requeue is not necessary because the cruiseContrlOperation controller retries the errored task
```

controllers/cruisecontroltask_controller.go
```diff
@@ -70,121 +81,218 @@ func (r *CruiseControlTaskReconciler) Reconcile(ctx context.Context, request ctr
...
 		log.Error(err, "failed to update Kafka Cluster status")
 	}
-	return requeueAfter(DefaultRequeueAfterTimeInSec)
+
+	return reconciled()
+}
+
+func checkBrokersAvailability(scaler scale.CruiseControlScaler, brokerIDs []string) ([]string, error) {
+	states := []scale.KafkaBrokerState{scale.KafkaBrokerAlive, scale.KafkaBrokerNew}
+	// This can result NullPointerException when the capacity calculation is missing for a broker in the cruisecontrol configmap
```
Kuvesz, line 217: Can we fix that somehow? I guess we should if we can just in case. :D

Kuvesz (commented):
I may need to have another look at the cruisecontrol_operations.go later on

controllers/cruisecontroltask_controller.go
bartam1: Yes it is needed to be removed because this part is in the cruisecontroloperation_controller

bartam1: Yes it is needed to be removed because this part is in the cruisecontroloperation_controller

config/base/crds/kafka.banzaicloud.io_kafkaclusters.yaml
bartam1: I have modified this.

controllers/cruisecontroltask_controller.go
bartam1: I have done it more nicely

Kuvesz (approved):
Looks good to me really, would appreciate a look from someone else as well. :)

controllers/cruisecontroltask_controller.go
```diff
@@ -195,55 +309,71 @@ func (r *CruiseControlTaskReconciler) UpdateStatus(ctx context.Context, instance
...
+		},
+	}
 
-	return builder
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&banzaiv1beta1.KafkaCluster{}).
+		WithEventFilter(SkipClusterRegistryOwnedResourcePredicate{}).
+		WithEventFilter(kafkaClusterPredicate).
+		Owns(&banzaiv1alpha1.CruiseControlOperation{}, builder.WithPredicates(cruiseControlOperationPredicate)).
+		Named("CruiseControl")
```
panyuenlau, line 371: What about having this to match the controller name?
```suggestion
		Named("CruiseControlTask")
```

controllers/cruisecontroltask_types.go
```diff
@@ -15,44 +15,34 @@
...
-	Operation CruiseControlOperation
+	BrokerID                        string
+	BrokerState                     kafkav1beta1.CruiseControlState
+	Volume                          string
+	VolumeState                     kafkav1beta1.CruiseControlVolumeState
+	Operation                       kafkav1alpha1.CruiseControlTaskOperation
+	CruiseControlOperationReference *corev1.LocalObjectReference
 }
 
 // IsDone returns true if the task is considered finished.
```
panyuenlau, line 34: Please update this comment to reflect the changes in this method

panyuenlau (commented):
It's great that you include integration test(s) to cover the new controllers. 

And it looks like we don't have unit tests to cover the new functions / methods created in this PR, do you think we can include unit tests in the PR as well?

controllers/cruisecontroltask_controller.go
```diff
@@ -70,121 +82,223 @@ func (r *CruiseControlTaskReconciler) Reconcile(ctx context.Context, request ctr
...
+func (r *CruiseControlTaskReconciler) removeBroker(ctx context.Context, kafkaCluster *banzaiv1beta1.KafkaCluster, ttlSecondsAfterFinished *int, brokerID string) (corev1.LocalObjectReference, error) {
+	return r.createCCOperation(ctx, kafkaCluster, banzaiv1alpha1.ErrorPolicyRetry, ttlSecondsAfterFinished, banzaiv1alpha1.OperationRemoveBroker, []string{brokerID})
+}
+
+func (r *CruiseControlTaskReconciler) rebalanceDisks(ctx context.Context, kafkaCluster *banzaiv1beta1.KafkaCluster, ttlSecondsAfterFinished *int, bokerIDs []string) (corev1.LocalObjectReference, error) {
+	return r.createCCOperation(ctx, kafkaCluster, banzaiv1alpha1.ErrorPolicyRetry, ttlSecondsAfterFinished, banzaiv1alpha1.OperationRebalance, bokerIDs)
+}
+
+func (r *CruiseControlTaskReconciler) createCCOperation(ctx context.Context, kafkaCluster *banzaiv1beta1.KafkaCluster, errorPolicy banzaiv1alpha1.ErrorPolicyType, ttlSecondsAfterFinished *int,
+	operationType banzaiv1alpha1.CruiseControlTaskOperation, bokerIDs []string) (corev1.LocalObjectReference, error) {
```
pregnor, line 253: **Recommended**: I think in such long function name cases we **SHOULD** use the convention of splitting one parameter per line for readability/maintainability purposes. We might lose the omitted type possibility, but it makes function signatures more readable.

```suggestion
func (r *CruiseControlTaskReconciler) createCCOperation(
	ctx context.Context, 
	kafkaCluster *banzaiv1beta1.KafkaCluster, 
	errorPolicy banzaiv1alpha1.ErrorPolicyType,
	ttlSecondsAfterFinished *int,
	operationType banzaiv1alpha1.CruiseControlTaskOperation, 
	bokerIDs []string,
) (corev1.LocalObjectReference, error) {
```

controllers/cruisecontroltask_controller.go
```diff
@@ -195,55 +309,71 @@ func (r *CruiseControlTaskReconciler) UpdateStatus(ctx context.Context, instance
...
 			taskAndStates.SyncState(instance)
-			err = r.Client.Status().Update(ctx, instance)
-			if err != nil {
-				return err
-			}
-		} else {
-			return err
 		}
+		return err
+	}); err != nil {
```
pregnor, line 323: **Recommended**: I think we **SHOULD NOT** put an inline function definition argumented function call into a short-statemented-if statement, because it makes everything much more difficult to read and thus maintain properly - the code logic/control flow is very difficult to understand at a glance and we don't really spare that much for it.

```golang
	conflictRetryFunction := func() error {
		log.Info("Updating status....")
		err := r.Status().Update(ctx, instance)
		if apiErrors.IsConflict(err) {
			err := r.Client.Get(ctx, types.NamespacedName{Name: instance.Name, Namespace: instance.Namespace}, instance)
			if err != nil {
				return errors.WithMessage(err, "failed to get updated Kafka Cluster CR before updating its status")
			}
			taskAndStates.SyncState(instance)
		}
		return err
	}
	
	if err := util.RetryOnConflict(util.DefaultBackOffForConflict, conflictRetryFunction); err != nil {
		return err
	}
```

I know we are "wasting" the variable, but I do believe - not really in the mood to search for and prove - the Go compiler is smart enough to detect that if a variable is only used once in the function then its value can be inlined into the single usage place when generating assembly code. I do not want to trade readability for (almost certainly premature) optimization.

(Apologies, this is a big pet peeve of mine.)

api/v1beta1/common_types.go
```diff
@@ -98,23 +105,20 @@ func (r CruiseControlState) IsRequiredState() bool {
...
-		return r
-	default:
-		return r
-	}
+// IsSucceeded returns true if CruiseControlState is succeeded
+func (r CruiseControlState) IsSucceeded() bool {
+	return r == GracefulDownscaleSucceeded || r == GracefulUpscaleSucceeded
+}
+
+// IsSucceeded returns true if CruiseControlState is succeeded
```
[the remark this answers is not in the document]
bartam1: Yes :)

controllers/cruisecontroltask_controller.go
```diff
@@ -292,28 +416,21 @@ func getActiveTasksFromCluster(instance *kafkav1beta1.KafkaCluster) *CruiseContr
...
-
-	taskResultsByID := make(map[string]*scale.Result, len(tasks))
-	for _, task := range tasks {
-		taskResultsByID[task.TaskID] = task
+// status from CruiseControlOperation
+func updateActiveTasks(tasksAndStates *CruiseControlTasksAndStates, ccOperations []*banzaiv1alpha1.CruiseControlOperation) error {
+	ccOperationMap := make(map[string]*banzaiv1alpha1.CruiseControlOperation)
+	for i := range ccOperations {
+		operation := ccOperations[i]
+		ccOperationMap[operation.Name] = operation
```
pregnor, line 424: **Recommended**: I think we **SHOULD** do this in one step, and comment here a note to avoid accidental regression by someone hastily changing this into a direct assignment of an implicit iteration pointer instead of doing it this way.

```suggestion
		ccOperationMap[operation.Name] = ccOperations[i] // Note: deep copying pointer instead of iteration pointer intentionally because the implicit iteration pointer would point to a changing value.
```

Here the regular pointer assignment rules apply copying the pointer, the main thing is to not use the implicit iteration pointer because that has a changing underlying value.

controllers/cruisecontroltask_types.go
```diff
@@ -62,71 +52,112 @@ func (t *CruiseControlTask) Apply(instance *kafkav1beta1.KafkaCluster) {
 	if t == nil || instance == nil {
 		return
 	}
-
+	//nolint:exhaustive
```
pregnor, line 55: **Recommended**: We **SHOULD** also leave a comment here why it is required to override the linter for future readers.

Example:
```suggestion
	// nolint:exhaustive // Note: Not all CC operations require updates to the KafkaCluster status.
```

controllers/cruisecontroltask_types.go
```diff
@@ -62,71 +52,112 @@ func (t *CruiseControlTask) Apply(instance *kafkav1beta1.KafkaCluster) {
...
 
+func (t *CruiseControlTask) SetCruiseControlOperationRef(ref corev1.LocalObjectReference) {
+	if t == nil {
+		return
+	}
+	t.CruiseControlOperationReference = &ref
+}
+
+func (t *CruiseControlTask) SetStateScheduled() {
+	//nolint:exhaustive
```
pregnor, line 82: **Recommended**: We **SHOULD** also leave a comment here why it is required to override the linter for future readers.

Example:
```suggestion
	// nolint:exhaustive // Note: Not all CC operations require updates to the CruiseControlTask state.
```

controllers/cruisecontroltask_controller.go
```diff
@@ -70,121 +82,223 @@ func (r *CruiseControlTaskReconciler) Reconcile(ctx context.Context, request ctr
...
+		}
+		if len(unavailableBrokers) > 0 {
+			log.Info("requeue as broker(s) are not ready for upscale", "brokerIDs", unavailableBrokers)
+			// This requeue is not necessary because the cruisecontrloperation controller retries the errored task
+			return requeueAfter(DefaultRequeueAfterTimeInSec)
+		}
+
+		cruiseControlOpRef, err := r.addBrokers(ctx, instance, operationTTLSecondsAfterFinished, brokerIDs)
+		if err != nil {
+			log.Error(err, "creating CruiseControlOperation for upscale has failed", details...)
```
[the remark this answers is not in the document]
bartam1: Nice catch! You have right!

controllers/cruisecontroltask_types.go
```diff
@@ -15,44 +15,34 @@
...
-)
+	corev1 "k8s.io/api/core/v1"
 
-type CruiseControlOperation int8
-
-const (
-	OperationAddBroker CruiseControlOperation = iota
-	OperationRemoveBroker
-	OperationRebalanceDisks
+	kafkav1alpha1 "github.com/banzaicloud/koperator/api/v1alpha1"
```
pregnor, line 20: **Recommended**: This **SHOULD** be aliased as `koperatorv1alpha1` IMO to distinguish from upstream Kafka types.

```suggestion
	koperatorv1alpha1 "github.com/banzaicloud/koperator/api/v1alpha1"
```

This requires updates throughout the file.

controllers/cruisecontroltask_types.go
```diff
@@ -15,44 +15,34 @@
...
+	kafkav1beta1 "github.com/banzaicloud/koperator/api/v1beta1"
```
pregnor, line 21: **Recommended**: This **SHOULD** be aliased as `koperatorv1beta1` IMO to distinguish from upstream Kafka types.

```suggestion
	koperatorv1beta1 "github.com/banzaicloud/koperator/api/v1beta1"
```

This requires updates throughout the file.

controllers/tests/cruisecontroltask_controller_test.go
```diff
@@ -64,56 +70,136 @@ var _ = Describe("CruiseControlTaskReconciler", func() {
...
+					if len(operationList.Items) != 1 {
+						return false
+					}
+					operation := &operationList.Items[0]
+					return actionState.CruiseControlOperationReference.Name == operation.Name && operation.GetCurrentTaskOp() == v1alpha1.OperationAddBroker && actionState.CruiseControlState == v1beta1.GracefulUpscaleScheduled
+				}
+				return false
+			}, 10*time.Second, 500*time.Millisecond).Should(BeTrue())
+		})
+		When("created CruiseControlOperation state is inExecutuin", func() {
```
pregnor, line 114: **Optional**: This **MAY** be a typo.

```suggestion
		When("created CruiseControlOperation state is inExecution", func() {
```

pregnor (approved):
Apologies for multiple comments referring to similar change proposals, but because the actual code is different I wanted to separately comment proposed new code.

controllers/cruisecontroltask_controller.go
```diff
@@ -70,121 +82,223 @@ func (r *CruiseControlTaskReconciler) Reconcile(ctx context.Context, request ctr
...
+					unavailableBrokerIDs = append(unavailableBrokerIDs, task.BrokerID)
 				}
-			} else {
-				task.Err = "log dir is not available in Cruise Control"
 			}
 		}
+		if len(unavailableBrokerIDs) > 0 {
+			log.Info("requeue as there are offline broker log dirs for rebalance", "brokerIDs", unavailableBrokerIDs)
+			// This requeue is not necessary because the cruiseContrlOperation controller retries the errored task
+			return requeueAfter(DefaultRequeueAfterTimeInSec)
```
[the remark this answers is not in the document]
bartam1: In this case there will be completedWithError in the kafkaCluster status until the all of the brokers become available.
There will be requeue in the CruiseControlOperation reconciler or here.
I think better to do here because there will be no completedWithError status

bartam1:
> It's great that you include integration test(s) to cover the new controllers.
> 
> And it looks like we don't have unit tests to cover the new functions / methods created in this PR, do you think we can include unit tests in the PR as well?

I think there is no complex logic in this PR to write a  unit test for that. What do you think?

pregnor:
> > It's great that you include integration test(s) to cover the new controllers.
> > And it looks like we don't have unit tests to cover the new functions / methods created in this PR, do you think we can include unit tests in the PR as well?
> 
> I think there is no complex logic in this PR to write a unit test for that. What do you think?

I think the ones worth unit testing are the ones difficult to unit test thus using integration test rather.

But I agree with the general notion of writing unit tests for everything we can and is worth writing unit tests for.
Darren if you can point out unit-testable but not tested functions that can help Márton identify what you would require more concretely.

controllers/cruisecontroloperation_controller.go
```diff
@@ -240,6 +239,19 @@ func (r *CruiseControlOperationReconciler) Reconcile(ctx context.Context, reques
...
+// In this case the CR should be skipped because the scale mock interference
+func (r *CruiseControlOperationReconciler) isCCTaskTest(operation *banzaiv1alpha1.CruiseControlOperation) bool {
+	_, ok := r.Scaler.(*scale.MockCruiseControlScaler)
+	for _, ownerRef := range operation.GetOwnerReferences() {
+		if ownerRef.Name == CruiseControlTaskTestKafkaClusterName {
+			return ok
+		}
+
+	}
+	return false
```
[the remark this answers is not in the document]
bartam1: r.Scaler.(*scale.MockCruiseControlScaler) returns 2 value and this function is waiting one bool. 
This does not work.
I have modified it a bit.

controllers/cruisecontroltask_controller.go
```diff
@@ -70,121 +82,223 @@ func (r *CruiseControlTaskReconciler) Reconcile(ctx context.Context, request ctr
...
 		log.Error(err, "requeue event as updating state of active tasks failed")
 		return requeueAfter(DefaultRequeueAfterTimeInSec)
 	}
 
-	// Check if CruiseControl is ready as we cannot perform any operation until it is in ready state
-	if status, _ := scaler.Status(); status.InExecution() {
-		log.Info("updating status of Kafka Cluster and requeue event as Cruise Control is in execution")
-		if err := r.UpdateStatus(ctx, instance, tasksAndStates); err != nil {
-			log.Error(err, "failed to update Kafka Cluster status")
+	if _, ok := r.Scaler.(*scale.MockCruiseControlScaler); !ok {
```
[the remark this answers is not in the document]
bartam1: For the first question:
Sometimes when the cached client is used the requested resource does not contain the modifications in the status  because of the fast requeue. This can lead duplicated task execution :( I found It safer to use here not caching directClient here.

controllers/cruisecontroltask_controller.go
```diff
@@ -70,121 +82,223 @@ func (r *CruiseControlTaskReconciler) Reconcile(ctx context.Context, request ctr
...
+			return requeueAfter(DefaultRequeueAfterTimeInSec)
+		}
 
 		details := []interface{}{"operation", "rebalance disks", "brokers", brokerIDs}
-
-		result, err := scaler.RebalanceDisks(brokerIDs...)
+		cruiseControlOpRef, err := r.rebalanceDisks(ctx, instance, operationTTLSecondsAfterFinished, brokerIDs)
 		if err != nil {
-			log.Error(err, "re-balancing disk(s) in Kafka cluster via Cruise Control failed", details...)
+			log.Error(err, "creating CruiseControlOperation for re-balancing disks has failed", details...)
```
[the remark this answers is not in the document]
bartam1: It is a bug. I have fixed it.

controllers/cruisecontroloperation_controller.go
```diff
@@ -92,8 +92,7 @@ func (r *CruiseControlOperationReconciler) Reconcile(ctx context.Context, reques
 		}
 	}
 
-	// When the resource has been removed before reconcile.
-	if currentCCOperation == nil {
+	if currentCCOperation == nil || r.isCCTaskTest(currentCCOperation) {
```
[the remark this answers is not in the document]
bartam1: I cant see many options to solve this so far. I think we should create a separate ticket to update ginkgo v1 to v2 and fix this test.

controllers/cruisecontroltask_controller.go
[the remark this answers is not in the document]
panyuenlau: hmmmm if that is the case, I think we should update the comment to reflect this so people that look at this piece of code will have some context

controllers/cruisecontroloperation_controller.go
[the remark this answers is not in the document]
pregnor: Oh, that is right, sorry about that, I didn't write it properly, I will check your variant then.

controllers/cruisecontroloperation_controller.go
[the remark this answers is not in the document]
pregnor: I agree, we should create a ticket for this and move on.

controllers/cruisecontroloperation_controller.go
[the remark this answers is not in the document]
pregnor: Looks perfect the way you changed it.

controllers/cruisecontroltask_controller.go
pregnor: This is the last change request I posted, it is a minor thing.

controllers/cruisecontroltask_controller.go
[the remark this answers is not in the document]
pregnor: Please also create a refactor ticket for the abstract factory thing.

controllers/cruisecontroltask_controller.go
```diff
@@ -16,31 +16,43 @@ package controllers
...
+	DefaultRequeueAfterTimeInSec          = 20
+	CruiseControlTaskTestKafkaClusterName = "cruisecontroltask-test"
+	nullPointerExceptionErrString         = "NullPointerException"
 )
 
 // CruiseControlTaskReconciler reconciles a kafka cluster object
 type CruiseControlTaskReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	DirectClient client.Reader
```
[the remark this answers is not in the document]
panyuenlau: @bartam1 can you provide some info to this question?

pregnor (approved):
LGTM aside from the 3 unresolved threads from my previous review (2 tickets and 1 minor change proposal).

controllers/cruisecontroltask_controller.go
```diff
@@ -70,121 +82,225 @@ func (r *CruiseControlTaskReconciler) Reconcile(ctx context.Context, request ctr
...
 	if err = r.UpdateStatus(ctx, instance, tasksAndStates); err != nil {
 		log.Error(err, "failed to update Kafka Cluster status")
 	}
-	return requeueAfter(DefaultRequeueAfterTimeInSec)
+
+	return reconciled()
 }
 
-// UpdateStatus updates the Status of the provided kafkav1beta1.KafkaCluster instance with the status of the tasks
+func getBrokersAvailability(scaler scale.CruiseControlScaler, brokerIDs []string) ([]string, error) {
```
panyuenlau, line 214: Can we please use `getUnavailableBrokers` instead? It is a bit misleading using `getBrokersAvailability` to get the unavailable broker IDs

controllers/cruisecontroltask_controller.go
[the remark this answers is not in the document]
panyuenlau: @bartam1 please update the comment in the source code to reflect your response in this thread

controllers/cruisecontroltask_controller.go
```diff
@@ -70,121 +82,223 @@ func (r *CruiseControlTaskReconciler) Reconcile(ctx context.Context, request ctr
...
+		unavailableBrokers, err := checkBrokersAvailability(r.Scaler, brokerIDs)
 		if err != nil {
-			log.Error(err, "adding broker(s) to Kafka cluster via Cruise Control failed", details...)
+			log.Error(err, "could not get unavailable brokers for upscale")
+			return requeueAfter(DefaultRequeueAfterTimeInSec)
+		}
+		if len(unavailableBrokers) > 0 {
+			log.Info("requeue as broker(s) are not ready for upscale", "brokerIDs", unavailableBrokers)
+			// This requeue is not necessary because the cruisecontrloperation controller retries the errored task
+			return requeueAfter(DefaultRequeueAfterTimeInSec)
```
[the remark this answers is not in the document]
panyuenlau: @bartam1 can you provide some info here?

panyuenlau (commented):
Looks like some of the comments from previous review got left out (probably because I left too many comments yesterday lol...)

Please provide some info to the un-addressed comments / questions before we can merge this big PR

controllers/cruisecontroltask_controller.go
bartam1: Lol I just change the check to get :)

bartam1:
Sure, yesterday I skipped this last one for reason.

controllers/cruisecontroltask_controller.go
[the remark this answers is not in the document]
bartam1: I have experienced few times that the status update was not in what I got with the cached client.Reader.
This is why I had to use DirectClient here. https://github.com/banzaicloud/koperator/blob/4e9d8d63298533aff1a33373d78dda130dc1f6e4/main.go#L187

PR #865, opened 2022-09-18
Cruisecontroloperaton ttl controller to auto remove unnecessary resources

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no |
| New feature?    | yes |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets | mentioned in #861   |

### What's in this PR?
When TTLSecondsAfterFinished is set then the Koperator will clean up the finished CruiseControlOperation CR when the finished time + TTLSecondsAfterFinished has elapsed.

### Why?
Cleaning up unnecessary CruiseControlOperation CRs, makes user's life easier

### Checklist

- [x] Implementation tested


controllers/cruisecontroloperation_ttl_controller.go
```diff
@@ -0,0 +1,116 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/banzaicloud/koperator/api/v1alpha1"
+	banzaiv1alpha1 "github.com/banzaicloud/koperator/api/v1alpha1"
+)
+
+const (
+	defaultRequeueCruiseControlOperatonTTL = 30
```
pregnor, line 35: **Recommended**: Because this is an internal default value, I think we **SHOULD** be using a structured type for this to make it easier to use.

```suggestion
	defaultRequeueCruiseControlOperatonTTL = 30 * time.Seconds
```

controllers/cruisecontroloperation_ttl_controller.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+		// Error reading the object - requeue the request.
+		return requeueWithError(log, err.Error(), err)
+	}
+
+	if ccOperation.GetTTLSecondsAfterFinished() == nil || ccOperation.GetCurrentTask().Finished == nil {
+		return reconciled()
+	}
+
+	operationTTL := time.Duration(*ccOperation.GetTTLSecondsAfterFinished())
```
pregnor, line 66: **Required**: We **MUST** handle the case where the TTL is nil, because it is a valid and often use-case.

```suggestion
	operationTTL := defaultRequeueCruiseControlOperatonTTL * time.Seconds // todo(Márton): if you change the type of the default from int to time.Duration, then upgrade this one as well to drop the conversion.

	setTTLSeconds := ccOperation.GetTTLSecondsAfterFinished()
	if setTTLSeconds != nil {
		operationTTL = *setTTLSeconds * time.Seconds
	}
```

controllers/cruisecontroloperation_ttl_controller.go
```diff
@@ -0,0 +1,116 @@
...
+		WithEventFilter(SkipClusterRegistryOwnedResourcePredicate{}).
+		WithEventFilter(cruiseControlOperationTTLPredicate).
+		Named("CruiseControlOperationTTL")
+
+	return builder
+}
+
+func IsExpired(ttl time.Duration, finishedAt time.Time) bool {
+	return finishedAt.Add(ttl).Before(time.Now())
+}
```
pregnor, line 106: **Optional**: I **MAY** consider replacing the implementation with `time.Since(finishedAt) < ttl`, because it is easier to read/understand (and that is what I am used to comparing elapsed times/durations), but could be opinionated on my end.

controllers/cruisecontroloperation_ttl_controller.go
```diff
@@ -0,0 +1,116 @@
...
+
+	operationTTL := time.Duration(*ccOperation.GetTTLSecondsAfterFinished())
+	finishedAt := ccOperation.GetCurrentTask().Finished
+
+	if IsExpired(operationTTL, finishedAt.Time) {
+		log.Info("cleaning up finished CruiseControlOperation", "name", ccOperation.GetName(), "namespace", ccOperation.GetNamespace())
+		return r.delete(ctx, ccOperation)
+	} else {
+		return requeueAfter(int(finishedAt.Time.Add(operationTTL).Sub(time.Now())))
+	}
```
pregnor, line 74: **Recommended**: I think we **SHOULD** put this logic on function top level, because it is [more idiomatic (regular case on top level)](https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow) and thus more familiar to Go developers.

controllers/cruisecontroloperation_ttl_controller.go
```diff
@@ -0,0 +1,116 @@
...
+
+// SetupCruiseControlWithManager registers cruise control controller to the manager
+func SetupCruiseControlOperationTTLWithManager(mgr ctrl.Manager) *ctrl.Builder {
+	cruiseControlOperationTTLPredicate := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return false
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			newObj := e.ObjectNew.(*banzaiv1alpha1.CruiseControlOperation)
+			if newObj.IsFinished() && newObj.GetTTLSecondsAfterFinished() != nil && newObj.GetDeletionTimestamp().IsZero() {
```
pregnor, line 85: **Question**: Shouldn't we also return true in case `newObj.GetTTLSecondsAfterFinished() == nil` in order to do the default cleanup when there is no explicit TTL set? 
In the current reconciliation logic we never update the TTL to the default in the K8s.

pregnor (changes requested):

controllers/tests/cruisecontroloperation_ttl_controller_test.go
```diff
@@ -0,0 +1,166 @@
...
+				Name: namespace,
+			},
+		}
+		kafkaClusterCRName = controllers.CruiseControlTaskTestKafkaClusterName
+		kafkaCluster = createMinimalKafkaClusterCR(kafkaClusterCRName, namespace)
+	})
+
+	JustBeforeEach(func() {
+		By("creating namespace " + namespace)
+		err := k8sClient.Create(context.TODO(), namespaceObj)
```
pregnor, line 61: **Recommended**: We also **SHOULD** use [`context.Background()`](https://pkg.go.dev/context#Background) in these cases unless we intend to change these in the foreseeable future, because we don't expect to time out in these tests.

> Background returns a non-nil, empty Context. It is never canceled, has no values, and has no deadline. It is typically used by the main function, initialization, and tests, and as the top-level Context for incoming requests.

The same holds true for the reset of the cases.

api/v1alpha1/cruisecontroloperation_types.go
```diff
@@ -60,6 +60,9 @@ type CruiseControlOperationSpec struct {
 	// +kubebuilder:default=retry
 	// +optional
 	ErrorPolicy ErrorPolicyType `json:"errorPolicy,omitempty"`
+	// TTLSecondsAfterFinished when is specified, the created succeeded CruiseControlOperation custom resource will be deleted after the given time elapsed.
+	// +kubebuilder:validation:Minimum=0
```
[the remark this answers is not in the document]
bartam1: In this case, it will be deleted instantly. I will add a comment about it.

api/v1alpha1/cruisecontroloperation_types.go
```diff
@@ -60,6 +60,9 @@ type CruiseControlOperationSpec struct {
 	// +kubebuilder:default=retry
 	// +optional
 	ErrorPolicy ErrorPolicyType `json:"errorPolicy,omitempty"`
+	// TTLSecondsAfterFinished when is specified, the created succeeded CruiseControlOperation custom resource will be deleted after the given time elapsed.
```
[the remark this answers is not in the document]
bartam1: The user cant gives a negative value because of k8s crd api validation.

controllers/cruisecontroloperation_ttl_controller.go
bartam1: This can be remove because it is not used anywhere.

controllers/cruisecontroloperation_ttl_controller.go
bartam1: I have fixed this bug in a commit later.

controllers/cruisecontroloperation_ttl_controller.go
bartam1: When it is not specified, there is no cleanup at all.

controllers/tests/cruisecontroloperation_ttl_controller_test.go
bartam1: context.Background() and context.TODO() is completely same.
https://go.dev/src/context/context.go  line 199

api/v1alpha1/cruisecontroloperation_types.go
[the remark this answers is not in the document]
bartam1: I added an extra description about this.

api/v1alpha1/cruisecontroloperation_types.go
```diff
@@ -60,6 +60,13 @@ type CruiseControlOperationSpec struct {
 	// +kubebuilder:default=retry
 	// +optional
 	ErrorPolicy ErrorPolicyType `json:"errorPolicy,omitempty"`
+	// TTLSecondsAfterFinished when is specified, the created completed successfully or completedWithError and errorPolicy: ignore
```
Kuvesz, line 63: I don't understand this sentence?

controllers/cruisecontroloperation_ttl_controller.go
```diff
@@ -0,0 +1,117 @@
...
+
+func IsExpired(ttl time.Duration, finishedAt time.Time) bool {
+	return time.Since(finishedAt) > ttl
+}
+
+func (r *CruiseControlOperationTTLReconciler) delete(ctx context.Context, ccOperation *v1alpha1.CruiseControlOperation) (reconcile.Result, error) {
+	log := logr.FromContextOrDiscard(ctx)
+	err := r.Delete(ctx, ccOperation)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return requeueWithError(log, "error is occured when deleting finished CruiseControlOperation ", err)
```
Kuvesz, line 113:
```suggestion
		return requeueWithError(log, "error occured when deleting finished CruiseControlOperation ", err)
```

Kuvesz (changes requested):
Looks good aside from that very weird sentence.

api/v1alpha1/cruisecontroloperation_types.go
bartam1: Please check it now. Is it better?

controllers/tests/cruisecontroloperation_ttl_controller_test.go
```diff
@@ -0,0 +1,166 @@
...
+				ID:        "12345",
+				Operation: v1alpha1.OperationAddBroker,
+				State:     v1beta1.CruiseControlTaskCompleted,
+				Finished:  &metav1.Time{Time: time.Now().Add(-time.Second*v1alpha1.DefaultRetryBackOffDurationSec - 10)},
+			}
+			err = k8sClient.Status().Update(context.Background(), &operation)
+			Expect(err).NotTo(HaveOccurred())
+
+		})
+		FIt("it should  remove the finished CruiseControlOperation", func() {
```
panyuenlau, line 87:
```suggestion
		FIt("it should remove the finished CruiseControlOperation", func() {
```
**Question**: why this is `FIt()`, shouldn't it be just `It()`?

controllers/tests/cruisecontroloperation_ttl_controller_test.go
```diff
@@ -0,0 +1,166 @@
...
+					Name:      opName,
+				}, &operation)
+				if err != nil && apierrors.IsNotFound(err) {
+					return true
+				}
+				return false
+			}, 10*time.Second, 500*time.Millisecond).Should(BeTrue())
+		})
+	})
+	When("there is finished (completedWithError and errorPolicy: ignore) operation with TTL", func() {
```
panyuenlau, line 101:
```suggestion
	When("there is a finished (completedWithError and errorPolicy: ignore) operation with TTL", func() {
```

controllers/cruisecontroloperation_ttl_controller.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	operationTTL := time.Duration(*ccOperation.GetTTLSecondsAfterFinished() * int(time.Second))
+	finishedAt := ccOperation.GetCurrentTask().Finished
+
+	if IsExpired(operationTTL, finishedAt.Time) {
+		log.Info("cleaning up finished CruiseControlOperation", "finished", finishedAt.Time, "clean-up time", finishedAt.Time.Add(operationTTL))
+		return r.delete(ctx, ccOperation)
+	}
+	reqSec := int(finishedAt.Time.Add(operationTTL).Sub(time.Now()).Seconds() + 1)
```
panyuenlau, line 69: Seems like we always requeue after 1 second, I think this is worth of making it a const variable either on top of the Reconcile function of as a global variable to make it more visible, what do you think?

panyuenlau (commented):
Implementation of the controller LGTM, just some places that I'v found to make the code look cleaner

controllers/cruisecontroloperation_ttl_controller.go
bartam1: This + 1 sec is needed because of the double to int conversion.  When double a = 1.8 the int(a) = 1. This is why I added +1, So when it will requeue the TTL must be elapsed .

bartam1: I think it is ok to not use const in this case

Kuvesz (approved):
LGTM

panyuenlau (approved):

controllers/cruisecontroloperation_ttl_controller.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+		// Error reading the object - requeue the request.
+		return requeueWithError(log, err.Error(), err)
+	}
+
+	if ccOperation.GetTTLSecondsAfterFinished() == nil || ccOperation.GetCurrentTask().Finished == nil {
+		return reconciled()
+	}
+
+	operationTTL := time.Duration(*ccOperation.GetTTLSecondsAfterFinished() * int(time.Second))
```
pregnor, line 62: **Note**: this could be done without double conversion, if I don't forget I'm going to comment it on the downstream PR.

pregnor (commented):
LGTM, apologies for being late

Issue #790, opened 2022-04-08
make install always exec kubectl replace logic


inetkiller:
**Describe the bug**
make install always exec kubectl replace logic, evenif the CRD  is't existing in kubernetes cluster，

**Steps to reproduce the issue:**
exec the command: make install

**Expected behavior**
Install CRDs into a cluster by creating when the CRD  is't existing

**Additional context**
os: CentOS Linux release 7.9.2009
make: GNU Make 3.82

**when delete .SHELLFLAGS in the makefile, things becomes correct.**

inetkiller:
bash version: GNU bash 4.2.46(2)-release (x86_64-redhat-linux-gnu)

bartam1:
Hello @inetkiller!
Thank you for the report!
I have created a PR to fix it.

PR #871, opened 2022-10-05
Fix for checking CRD existance on the cluster when installing CRDs

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |

### What's in this PR?
It fixes the Kubernetes CRD installation through Makefile in case when there are no Koperator CRDs on the cluster

### Why?
When "make install" (or deploy) is used the current check for CRD existence does not work.

### Checklist

- [x] Implementation tested

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,11 +104,7 @@ run: generate fmt vet manifests
 # Install CRDs into a cluster by manually creating or replacing the CRD depending on whether is currently existing
 # Apply is not applicable as the last-applied-configuration annotation would exceed the size limit enforced by the api server
 install: manifests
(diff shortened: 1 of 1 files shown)


pregnor (approved):

hi-im-aren (approved):

panyuenlau:
This should fix https://github.com/banzaicloud/koperator/issues/790, right?

bartam1:
> Do you think the changes would fix #790?

Yes, this one is fix it. It is the same. Thanks!

PR #893, opened 2022-11-17
update copyright statements to Cisco

mihaialexandrescu:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |

### What's in this PR?
Updating copyright statements to Cisco Systems branding.
* .go files that previously had a Copyright Banzai statement  (mass replace with sed)
* scripts/check-header.sh script  (ran it after changes and got no error or output)
* 2 CopyrightManager files (under the `.idea` dir) that I'm not so sure about (please advise during review)


README.md
```diff
@@ -73,7 +73,7 @@ When you are opening a PR to Koperator the first time we will require you to sig
 
 ## License
 
-Copyright (c) 2019 [Banzai Cloud, Inc.](https://banzaicloud.com)
+Copyright (c) 2019 [Cisco Systems, Inc.](https://www.cisco.com)
```
panyuenlau, line 76: I would change this to the following as well
```suggestion
Copyright (c) 2019 [Cisco Systems, Inc.](https://www.cisco.com) and/or its affiliates
```

mihaialexandrescu: done. thanks.

panyuenlau (approved):

pregnor (approved):
LGTM

bartam1:
Is it ok to have both? 
https://github.com/banzaicloud/koperator/blob/f77e9f7365ca2f7054168cf84ef5c74b84022887/NOTICE#L1-L2

mihaialexandrescu:
> Is it ok to have both?
> 
> https://github.com/banzaicloud/koperator/blob/f77e9f7365ca2f7054168cf84ef5c74b84022887/NOTICE#L1-L2

I have no idea.  Found it odd myself but the diff in this case is the ending of the second line, not its existence.

Do other reviewers have some thoughts on this ?

pregnor:
> Do other reviewers have some thoughts on this ?

IANAL, but I think it is okay to have both, but also okay to have the Cisco one with the original date as legally Cisco took over the Banzai Cloud copyrights at the acquisition, we can update that if we want to only have Cisco with the original date or we can leave it this way, Git history will preserve the states anyway.

PR #894, opened 2022-11-18
Support kafka cluster rebalance between broker disks when JBOD config is used

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no |
| New feature?    | yes |
| API breaks?     | no |
| Deprecations?   | no |

### What's in this PR?
Support disk rebalance at the rebalance CC endpoint when JBOD capacity configuration is available for all brokers 

### Why?
To rebalance kafka cluster between broker log dirs

### Checklist

- [x] Implementation tested


controllers/tests/cruisecontroltask_controller_test.go
```diff
@@ -72,13 +77,237 @@ var _ = Describe("CruiseControlTaskReconciler", func() {
...
+				brokerState1.GracefulActionState.VolumeStates = volumeState
+				kafkaCluster.Status.BrokersState["1"] = brokerState1
+
+				return k8sClient.Status().Update(context.Background(), kafkaCluster)
+			})
 
+			Expect(err).NotTo(HaveOccurred())
+
+		})
+		It("should creates not JBOD rebalance CruiseControlOperation", func() {
```
pregnor, line 276: **Recommended**: typo

```suggestion
		It("should create one not JBOD rebalance CruiseControlOperation", func() {
```

controllers/tests/cruisecontroltask_controller_test.go
```diff
@@ -72,13 +77,237 @@ var _ = Describe("CruiseControlTaskReconciler", func() {
...
+
+				brokerState.GracefulActionState.VolumeStates = volumeState
+				kafkaCluster.Status.BrokersState["0"] = brokerState
+				return k8sClient.Status().Update(context.Background(), kafkaCluster)
+			})
+
+			Expect(err).NotTo(HaveOccurred())
+
+		})
+		It("should creates one not JBOD rebalance CruiseControlOperation", func() {
```
pregnor, line 192: **Recommended**: typo

```suggestion
		It("should create one not JBOD rebalance CruiseControlOperation", func() {
```

controllers/tests/cruisecontroltask_controller_test.go
```diff
@@ -72,13 +77,237 @@ var _ = Describe("CruiseControlTaskReconciler", func() {
...
+
+				brokerState.GracefulActionState.VolumeStates = volumeState
+				kafkaCluster.Status.BrokersState["0"] = brokerState
+				return k8sClient.Status().Update(context.Background(), kafkaCluster)
+			})
+
+			Expect(err).NotTo(HaveOccurred())
+
+		})
+		It("should creates one JBOD rebalance CruiseControlOperation", func() {
```
pregnor, line 112: **Recommended**: typo

```suggestion
		It("should create one JBOD rebalance CruiseControlOperation", func() {
```

controllers/cruisecontroltask_controller.go
```diff
@@ -302,6 +348,69 @@ func (r *CruiseControlTaskReconciler) createCCOperation(
...
+	var capacityConfig koperatorccconf.JBODInvariantCapacityConfig
+	err = json.Unmarshal([]byte(capacityConfigJSON), &capacityConfig)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "could not unmarshal the user-provided broker capacity config")
+	}
+	for _, brokerCapacity := range capacityConfig.Capacities {
+		brokerCapacityMap, ok := brokerCapacity.(map[string]interface{})
+		if !ok {
+			continue
+		}
```
Kuvesz, line 371: Are we sure that just continuing here won't cause any issues? Should we at least log something?

controllers/cruisecontroltask_controller.go
```diff
@@ -302,6 +348,69 @@ func (r *CruiseControlTaskReconciler) createCCOperation(
 	}, nil
 }
 
+// brokersJBODSelector filters out the JBOD and not JBOD brokers from a broker list based on the capacityConfig
+func brokersJBODSelector(brokerIDs []string, capacityConfigJSON string) (brokersJBOD []string, brokersNotJBOD []string, err error) {
+	// JBOD is generated by default
+	if capacityConfigJSON == "" {
+		return brokerIDs, nil, nil
+	}
+	brokerIsJBOD := make(map[string]bool)
```
pregnor, line 357: **Note**: I think we could have spared this map by directly appending to one of the returned slices and on -1 case return the `brokerIDs` slice as either the first or second return value depending on the case, but I don't force this optimization if we feel the current solution improves clarity.

pregnor (dismissed):
LGTM, couple small things.

controllers/cruisecontroltask_controller.go
```diff
@@ -302,6 +348,69 @@ func (r *CruiseControlTaskReconciler) createCCOperation(
...
+					brokerIsJBOD[brokerID] = false
+				}
+			}
+			if _, ok := brokerIsJBOD[brokerId]; ok {
+				brokerIsJBOD[brokerId] = false
+			}
+			continue
+		}
+
+		// this cover that case when there was a -1 default capacity config but there is an override for a specific broker
```
Kuvesz, line 398:
```suggestion
		// this covers the case when there was a -1 default capacity config but there is an override for a specific broker
```

controllers/cruisecontroltask_controller.go
```diff
@@ -194,15 +190,38 @@ func (r *CruiseControlTaskReconciler) Reconcile(ctx context.Context, request ctr
 			return requeueAfter(DefaultRequeueAfterTimeInSec)
 		}
 
-		cruiseControlOpRef, err := r.rebalanceDisks(ctx, instance, operationTTLSecondsAfterFinished, brokerIDs)
+		allBrokerIDs := make([]string, 0, len(instance.Spec.Brokers))
+		for i := range instance.Spec.Brokers {
+			allBrokerIDs = append(allBrokerIDs, fmt.Sprint(instance.Spec.Brokers[i].Id))
+		}
+		// we can do rebalance between the broker's disks when JBOD capacity config is used
+		// this selector distinguish the JBOD brokers from the not JBOD brokers
```
Kuvesz, line 198:
```suggestion
		// this selector distinguishes the JBOD brokers from the not JBOD brokers
```

controllers/cruisecontroltask_controller.go
```diff
@@ -194,15 +190,38 @@ func (r *CruiseControlTaskReconciler) Reconcile(ctx context.Context, request ctr
...
+		// we need to search in all brokers to find out is there any not JBOD broker because
```
Kuvesz, line 199:
```suggestion
		// we need to search in all brokers to find out if there are any not JBOD brokers because
```

controllers/cruisecontroltask_controller.go
```diff
@@ -194,15 +190,38 @@ func (r *CruiseControlTaskReconciler) Reconcile(ctx context.Context, request ctr
...
+		// CC cannot do disk rebalance when one of the brokers has not JBOD capacity configuration
```
Kuvesz, line 200:
```suggestion
		// CC cannot do disk rebalance when at least one of the brokers has not JBOD capacity configuration
```

controllers/cruisecontroltask_controller.go
```diff
@@ -194,15 +190,38 @@ func (r *CruiseControlTaskReconciler) Reconcile(ctx context.Context, request ctr
...
+		_, brokersNotJBOD, err := brokersJBODSelector(allBrokerIDs, instance.Spec.CruiseControlConfig.CapacityConfig)
 		if err != nil {
-			return requeueWithError(log, fmt.Sprintf("creating CruiseControlOperation for re-balancing disks has failed, brokerIDs: %s", brokerIDs), err)
+			return requeueWithError(log, "failed to determine which broker using JBOD or not JBOD capacity configuration at rebalance operation", err)
+		}
+
+		var cruiseControlOpRef corev1.LocalObjectReference
+		// when there is one not JBOD broker in the kafka cluster CC cannot do the disk rebalance :(
```
Kuvesz, line 207:
```suggestion
		// when there is at least one not JBOD broker in the kafka cluster CC cannot do the disk rebalance :(
```

Kuvesz (commented):
Quick question and some suggestions on comments

controllers/cruisecontroltask_controller.go
bartam1: Here we use this: https://github.com/banzaicloud/koperator/blob/db78e506f76d23f6425f4a00867ef995f288ad6d/pkg/resources/cruisecontrol/configmap.go#L162 which comes from the past. When there is a syntax error in the kafkaCluster.Spec.CruiseControlConfig.CapacityConfig and we don't do any validation the Cruise Control will do and in the Cruise Control pod there will be error message and the pod will not be ready state. Because of that, we will not do any CC operation. That is why is safe to use here continue I think.

controllers/cruisecontroltask_controller.go
bartam1: When I first made this function I thought that CC can do disk rebalance for the JBOD brokers. That is why I made this selector and I created different CCOperation for JBOD and not JBOD brokers. Unfortunately, later I found that CC does not capable to do disk rebalance when there is one not JBOD broker in the KafkaCluster because this rebalance endpoint is designed in another way. Basically, it would be enough to do a return false when we found a not JBOD broker. I kept that function in the first form because who knows the future maybe this will be changed on the CC side.

pregnor: Yeah, it is perfectly fine to return a more detailed answer with the broker break down.
I just wanted to highlight using the slices would be enough, the map is only a convenience/clarity thing at most, but it is fine for me anyway.

pregnor (approved):
LGTM

Kuvesz (approved):
LGTM :)

pregnor (approved):

Issue #895, opened 2022-11-22
KafkaCluster validation fails when custom webhook port is provided


balassai:
**Describe the bug**
KafkaCluster webhook is not accessible when custom webhook port is provided.
The validations fails because the webhook is not reachable. 

**Steps to reproduce the issue:**
Set the webhook.serverPort in the chart's values yaml.
```
webhook:
  enabled: true
  serverPort: 8444
#  tls:
#    certDir: ""
  certs:
    generate: true
    secret: "kafka-operator-serving-cert"
```
Deploy the operator with helm.
```
helm install kafka charts/kafka-operator
```
Edit attempt to an existing KafkaCluster resource fails, because the webhook is unreachable.
```
k edit kafkacluster kafka -n kafka    
error: kafkaclusters.kafka.banzaicloud.io "kafka" could not be patched: Internal error occurred: failed calling webhook "kafkaclusters.kafka.banzaicloud.io": Post "[https://koperator-kafka-operator-operator.kafka.svc:443/validate-kafka-banzaicloud-io-v1beta1-kafkacluster?timeout=10s](https://koperator-kafka-operator-operator.kafka.svc/validate-kafka-banzaicloud-io-v1beta1-kafkacluster?timeout=10s)": dial tcp 10.254.29.128:443: connect: connection refused
You can run `kubectl replace -f /var/folders/8l/6jws9zmx1ng710v4v252b9jm0000gr/T/kubectl-edit-664393602.yaml` to try this update again.
```
**Expected behavior**
The webhook shall be reachable when custom webhook port is provided.

**Additional context**
version: 0.22

pregnor:
Resolved by #896, thanks.

PR #896, opened 2022-11-22
Configurable webhook target port

balassai:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets | #895 |

### What's in this PR?
Set the configured webhook port from the values.yaml explicitly on the operator's service.

### Why?
When the user wants to use custom webhook port and change the KafkaCluster resource the ValidatingWebhookConfiguration call fails.

### Checklist
- [x] Implementation tested

### DIFF
--- a/charts/kafka-operator/templates/operator-service.yaml
+++ b/charts/kafka-operator/templates/operator-service.yaml
@@ -28,8 +28,9 @@ spec:
   ports:
   - name: https
     port: 443
-    targetPort: {{ default 443 .Values.webhook.serverPort }}
+    targetPort: {{ (.Values.webhook).serverPort | default 443 }}
   {{- if and .Values.prometheusMetrics.enabled (not .Values.prometheusMetrics.authProxy.enabled) }}
   - name: metrics
     port: 8080
+    targetPort: {{ (.Values.metricEndpoint).port | default 8080 }}
   {{- end }}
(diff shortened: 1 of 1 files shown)


pregnor (approved):

bartam1 (approved):

charts/kafka-operator/templates/operator-service.yaml
```diff
@@ -28,6 +28,7 @@ spec:
   ports:
   - name: https
     port: 443
+    targetPort: {{ default 443 .Values.webhook.serverPort }}
```
hi-im-aren, line 31:
```suggestion
    targetPort: {{ .Values.webhook.serverPort | default 443 }}
```
I would use it this way to stay in line with how we use it in the rest of the project

hi-im-aren (changes requested):
Hey, thanks for the contribution!

Since we are here, could you also add `targetPort: {{ (.Values.metricEndpoint).port | default 8080 }}` to the metrics endpoint below? It will have the same issue. 🙂

charts/kafka-operator/templates/operator-service.yaml
balassai: Hello @hi-im-aren, thank you for the quick review. 
I've updated the code according to the recommendations.

akatona84 (approved):

hi-im-aren (approved):

pregnor (approved):
Thanks for the contribution.

PR #903, opened 2022-11-30
Fix missing nil pointer check

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |

### What's in this PR?

It checks a variable nilness

### Why?

Sometimes a JBOD test runs into nil pointer dereference exceptions.

### Checklist

- ~[ ] Implementation tested~
- ~[ ] Error handling code meets the [guideline](https://github.com/banzaicloud/developer-guide/blob/master/docs/coding-style/error-handling-guide.md)~
-~ [ ] Logging code meets the guideline~
- ~[ ] User guide and development docs updated (if needed)~

### DIFF
--- a/controllers/tests/cruisecontroltask_controller_test.go
+++ b/controllers/tests/cruisecontroltask_controller_test.go
@@ -122,7 +122,7 @@ var _ = Describe("CruiseControlTaskReconciler", func() {
 					return false
 				}
 				volumeState, ok := brokerState.GracefulActionState.VolumeStates[mountPath]
-				if !ok {
+				if !ok || volumeState.CruiseControlOperationReference == nil {
 					return false
 				}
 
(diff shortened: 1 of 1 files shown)


hi-im-aren (approved):

controllers/tests/cruisecontroltask_controller_test.go
```diff
@@ -286,7 +286,7 @@ var _ = Describe("CruiseControlTaskReconciler", func() {
 					return false
 				}
 				volumeState, ok := brokerState.GracefulActionState.VolumeStates[mountPath]
-				if !ok {
+				if !ok || volumeState.CruiseControlOperationReference == nil {
```
pregnor, line 289: **Question**: Won't this still fail on the same test just not with a nil pointer dereference, but earlier on this `return false` ending up failing the test case?

pregnor (approved):

PR #904, opened 2022-12-02
Added imagePullSecrets to deployment template

Kuvesz:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |

### What's in this PR?
I added imagePullSecrets to the deployment template and updated the chart version. Also updated release.yml to not include dev tags for now in releases.

### Why?
Not sure when there would be a need for this as the images are publicly accessible but just in case it doesn't hurt if it's there.

### Checklist
- [x] Implementation tested

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,11 +1,11 @@
-name: CI test
+name: Release
 
 on:
   push:
     tags:
       - 'v?[0-9]+.[0-9]+.[0-9]+'
       - "v?[0-9]+.[0-9]+.[0-9]+-rc.[0-9]+"
-      - "v?[0-9]+.[0-9]+.[0-9]+-dev.[0-9]+"
+      #- "v?[0-9]+.[0-9]+.[0-9]+-dev.[0-9]+" Disabled for now for dev tags
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/charts/kafka-operator/Chart.yaml
+++ b/charts/kafka-operator/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: kafka-operator
-version: 0.22.0
+version: 0.23.0-dev.0
 description: kafka-operator manages Kafka deployments on Kubernetes
 sources:
   - https://github.com/banzaicloud/koperator
-appVersion: v0.22.0
+appVersion: v0.23.0-dev.0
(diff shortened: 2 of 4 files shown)


pregnor (commented):
The change itself looks okay even though I would argue we should add commented out example code into the `values.yaml`.

The only use-case I could come up with would possibly be an air-gapped system with access control over the images.
I'm hesitant to approve, because I don't want additional unused code in the code base, but also don't want to block progress or uniforming things.

Kuvesz:
> The change itself looks okay even though I would argue we should add commented out example code into the `values.yaml`.
> 
> The only use-case I could come up with would possibly be an air-gapped system with access control over the images. I'm hesitant to approve, because I don't want additional unused code in the code base, but also don't want to block progress or uniforming things.

To be honest I'm not exactly sure we need this either as I've stated in my original comment. Everything else will be fine without it really and we can come back to this if someone from the community feels like they need it?

pregnor (dismissed):

pregnor:
We can let it fly as the values.yaml already contains the corresponding config.

bartam1 (changes requested):
I think we can remove this part after that change:
https://github.com/banzaicloud/koperator/blob/c30ef4c02685fd79ed8fcba6cbe337acf3715ad0/charts/kafka-operator/templates/operator-rbac.yaml#L4

Kuvesz:
> I think we can remove this part after that change:
> 
> https://github.com/banzaicloud/koperator/blob/c30ef4c02685fd79ed8fcba6cbe337acf3715ad0/charts/kafka-operator/templates/operator-rbac.yaml#L4

I don't quite get why? Would you mind explaining it a bit further?

bartam1:
> > I think we can remove this part after that change:
> > https://github.com/banzaicloud/koperator/blob/c30ef4c02685fd79ed8fcba6cbe337acf3715ad0/charts/kafka-operator/templates/operator-rbac.yaml#L4
> 
> I don't quite get why? Would you mind explaining it a bit further?

When you add an imagepullsecret into a serviceaccount then a pod which use that serviceaccount will contain the imagepullsecret what is in the serviceaccount. Here you add that imagepullsecret into the deployment anyway. Our serviceaccount imagepullsecret solution is wrong I think because when a user disable (https://github.com/banzaicloud/koperator/blob/c30ef4c02685fd79ed8fcba6cbe337acf3715ad0/charts/kafka-operator/values.yaml#L73) the serviceaccount creation then the imagepullsecret will not be in the default serviceaccount and in the  koperator's pod.

pregnor (approved):

Issue #795, opened 2022-04-14
Support Koperator in Openshift


cniackz:
**Is your feature request related to a problem? Please describe.**
Yes, I am trying to install Koperator in Openshift but due Security several issues I have faced. Last issue is:
```sh
$ oc logs kafka-operator-operator-8446494586-b8frg -n kafka manager
{
  "level": "info",
  "ts": "2022-04-14T01:57:10.700Z",
  "logger": "controller.KafkaCluster",
  "msg": "creating resource failed: configmaps \"kafka-kafka-jmx-exporter\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>",
  "reconciler group": "kafka.banzaicloud.io",
  "reconciler kind": "KafkaCluster",
  "name": "kafka",
  "namespace": "kafka"
}
```

**Describe the solution you'd like to see**
In short: Add documentation (Installation Steps) on how to get `Koperator` running on `OpenShift` as other companies does like `Strimzi`.
Bit more of story:
I am following the instructions provided at: https://banzaicloud.com/docs/supertubes/kafka-operator/install-kafka-operator/
Specifically I am not using SuperTubes but Installing Koperator and the requirements independently, when I try this in `minikube` no issue whatsoever but I can't find the instructions for `OpenShift` and then I feel I am on my own solving all sort of problems. If we compare your documentation with [Strimzi](https://strimzi.io/docs/0.5.0/), I find much easier to follow up with Strimzi.

**Describe alternatives you've considered**
I was able to use [Strimzi](https://strimzi.io/docs/0.5.0/) and it is working fine in my OpenShift Environment, less steps and online support in general. I will try to keep using Strimzi for now but would be nice to have Banzai solution working as well.

**Additional context**
If I have to provide a bit more of context, the reason behind this is to have `MinIO` creating notifications in `Kafka` or `Kafka` saving files/logs in `MinIO` as the `S3` solution. The Kafka Operator is more as a secondary step for me but still required so that I can help customers to setup different environments. Also notice I got to run the `operator` at last after many changes in the `SecurityContext`:
```sh
oc get pods -n kafka
NAME                                       READY   STATUS    RESTARTS   AGE
kafka-operator-operator-8446494586-b8frg   2/2     Running   10         171m
prometheus-kafka-prometheus-0              2/2     Running   1          8h
```
But the problem is that I still don't get the brokers and maybe related to the issue posted at the start of the conversation.

[...]

PR #912, opened 2022-12-15
Add Redhat Openshift support

hi-im-aren:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no |
| New feature?    | yes |
| API breaks?     | no |
| Deprecations?   | no |
| Issues   | #795 |

### What's in this PR?
Add openshift support by:
- adding finalizer rbac
- changing default webhook port from protected 443 to 9443

### Checklist
- [x] Implementation tested
  - https://github.com/banzaicloud/koperator-docs/pull/89

### ToDo
- ~Kafka broker image is updated~ (We will customize the broker images in a later step because it seems to work fine with the current ones and the modification is a bigger effort because of the init containers.)
  - https://github.com/banzaicloud/docker-kafka/pull/29
- [X] Envoy ingress external listener fix API extension: https://github.com/banzaicloud/koperator/pull/956

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/banzaicloud/istio-client-go v0.0.17
 	github.com/banzaicloud/istio-operator/api/v2 v2.15.1
 	github.com/banzaicloud/k8s-objectmatcher v1.8.0
-	github.com/banzaicloud/koperator/api v0.24.0
+	github.com/banzaicloud/koperator/api v0.25.0
 	github.com/banzaicloud/koperator/properties v0.4.1
 	github.com/cert-manager/cert-manager v1.9.1
 	github.com/cisco-open/cluster-registry-controller/api v0.2.5
--- a/go.sum
+++ b/go.sum
@@ -100,8 +100,8 @@ github.com/banzaicloud/istio-operator/api/v2 v2.15.1 h1:BZg8COvoOJtfx/dgN7KpoOnc
 github.com/banzaicloud/istio-operator/api/v2 v2.15.1/go.mod h1:5qCpwWlIfxiLvBfTvT2mD2wp5RlFCDEt8Xql4sYPNBc=
 github.com/banzaicloud/k8s-objectmatcher v1.8.0 h1:Nugn25elKtPMTA2br+JgHNeSQ04sc05MDPmpJnd1N2A=
 github.com/banzaicloud/k8s-objectmatcher v1.8.0/go.mod h1:p2LSNAjlECf07fbhDyebTkPUIYnU05G+WfGgkTmgeMg=
-github.com/banzaicloud/koperator/api v0.24.0 h1:RwhKWy8umzpKhKEa0J6xgvv5wOU37ti3A9JqIjCHrDk=
-github.com/banzaicloud/koperator/api v0.24.0/go.mod h1:qvpewvjdELAnfO70vg9397CXZ4K4uHxpiWtf5fhKSrQ=
+github.com/banzaicloud/koperator/api v0.25.0 h1:cRfoWRUThrAEVnszeeXJkz42gNGezonl3+bGdvbxkNQ=
+github.com/banzaicloud/koperator/api v0.25.0/go.mod h1:qvpewvjdELAnfO70vg9397CXZ4K4uHxpiWtf5fhKSrQ=
 github.com/banzaicloud/koperator/properties v0.4.1 h1:SB2QgXlcK1Dc7Z1rg65PJifErDa8OQnoWCCJgmC7SGc=
 github.com/banzaicloud/koperator/properties v0.4.1/go.mod h1:TcL+llxuhW3UeQtVEDYEXGouFLF2P+LuZZVudSb6jyA=
 github.com/banzaicloud/operator-tools v0.28.0 h1:GSfc0qZr6zo7WrNxdgWZE1LcTChPU8QFYOTDirYVtIM=


dobrerazvan:
Yeah, those seem to be the changes required for rbacs to make the operator work (just tried them today). But there are some changes required to the kafka build to not use the root account otherwise the broker attached pvcs cannot be used (permission denied).

charts/kafka-operator/templates/operator-deployment-with-webhook.yaml
```diff
@@ -212,7 +212,7 @@ spec:
           {{- end }}
           ports:
           {{- if .Values.webhook.enabled }}
-            - containerPort: {{ .Values.webhook.serverPort | default 443 }}
+            - containerPort: {{ .Values.webhook.serverPort | default 9443 }}
```
panyuenlau, line 215: Can you provide some reference for the reasoning behind this change?

panyuenlau (approved):
lgtm - although I don't have access to OpenShift so can't really play with it myself

charts/kafka-operator/templates/operator-deployment-with-webhook.yaml
dobrerazvan: It looks like openshift, by default, due to its security constraints, it does not allow binding on ports lower than 1024.

hi-im-aren:
> Yeah, those seem to be the changes required for rbacs to make the operator work (just tried them today). But there are some changes required to the kafka build to not use the root account otherwise the broker attached pvcs cannot be used (permission denied).

Yeah I made those changes as well but forgot to create a PR for it in our kafka docker repo, I will add it in a sec.

charts/kafka-operator/templates/operator-deployment-with-webhook.yaml
panyuenlau: Thanks for the info! 

Do you guys think this change will cause any backward compatibility issues for the existing users when they upgrade the Koperator?

controllers/kafkatopic_controller.go
```diff
@@ -60,6 +60,7 @@ type KafkaTopicReconciler struct {
 
 // +kubebuilder:rbac:groups=kafka.banzaicloud.io,resources=kafkatopics,verbs=get;list;watch;create;update;patch;delete;deletecollection
 // +kubebuilder:rbac:groups=kafka.banzaicloud.io,resources=kafkatopics/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=kafka.banzaicloud.io,resources=kafkatopics/finalizer,verbs=create;update;patch;delete
```
amuraru, line 63:
```suggestion
// +kubebuilder:rbac:groups=kafka.banzaicloud.io,resources=kafkatopics/finalizers,verbs=create;update;patch;delete
```

controllers/kafkauser_controller.go
```diff
@@ -154,6 +154,7 @@ type KafkaUserReconciler struct {
 
 // +kubebuilder:rbac:groups=kafka.banzaicloud.io,resources=kafkausers,verbs=get;list;watch;create;update;patch;delete;deletecollection
 // +kubebuilder:rbac:groups=kafka.banzaicloud.io,resources=kafkausers/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=kafka.banzaicloud.io,resources=kafkausers/finalizer,verbs=create;update;patch;delete
```
amuraru, line 157:
```suggestion
// +kubebuilder:rbac:groups=kafka.banzaicloud.io,resources=kafkausers/finalizers,verbs=create;update;patch;delete
```

pregnor:
Thanks for the reviews, I will take over this PR and fix the commented things.

charts/kafka-operator/templates/operator-deployment-with-webhook.yaml
tiswanso: the compatibility would be an issue because of the service definition using numerical values for targetPort rather than container port name references.  Added a comment on the service template diff.  

k8s doc ref:  https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service
> Port definitions in Pods have names, and you can reference these names in the targetPort attribute of a Service. ...
> This works even if there is a mixture of Pods in the Service using a single configured name, with the same network protocol available via different port numbers. This offers a lot of flexibility for deploying and evolving your Services. For example, you can change the port numbers that Pods expose in the next version of your backend software, without breaking clients.

tiswanso:
fyi, we're also going to need to set `runAsUser` and `runAsGroup` in the container security context to match the nonroot ID used in the docker-kafka build.

charts/kafka-operator/templates/operator-deployment-with-webhook.yaml
pregnor: ->Darren: I think the webhook port where the webhook is available is mainly used by K8s API server to send objects to, especially in the case of the container port itself (CMIIW). The service port is a bit more exposed, I will address that a bit further below.
A user - technically speaking - could have built on top of the default behavior (default 443 port), but we can signal the change of that in the release notes.
It's arguable whether we should classify this as major (breaking) change. I'm suggesting to stay on the safe side and chalk it up as a breaking (default behavior) change in the notes and give it an extra highlight, but otherwise we shouldn't be too concerned, it is easily fixable by code building on top of this - and hopefully people didn't hard code it but derived the dynamic value from the manifest.
In addition we are still on 0.x exactly because of these things, the API can change including open port numbers.

I will address Tim's comment on a later chain where it also popped up.

pregnor:
I force-with-lease-pushed the branch to clean up redundant commits I manually rebased, I will fix the comments in subsequent pushes, stay tuned.

controllers/kafkatopic_controller.go
pregnor: Fixed in https://github.com/banzaicloud/koperator/compare/50b41f0bbe5458899769a98dc20153c011f7f752..1df5df1d3d9248a0dff0b6dc4cb311acf210045a.

controllers/kafkauser_controller.go
pregnor: Fixed in https://github.com/banzaicloud/koperator/compare/50b41f0bbe5458899769a98dc20153c011f7f752..1df5df1d3d9248a0dff0b6dc4cb311acf210045a.

charts/kafka-operator/templates/operator-deployment-with-webhook.yaml
tiswanso: to clarify, this is only a breaking change because new upgraded endpoints with different containerPort numerical values won't automatically have their ports mapped when under the non-upgraded service.  Essentially if the service definition is changed to use the port name then the k8s service port to containerPort mapping works for all cases.  So for upgrade, compatibility is preserved as long as the k8s service is upgraded concurrently with or before the koperator endpoint using the new port number for the webhook port--assuming both new and old versions of koperator deployments used the same port NAME.

charts/kafka-operator/templates/operator-service.yaml
```diff
@@ -28,7 +28,7 @@ spec:
   ports:
   - name: https
     port: 443
-    targetPort: {{ (.Values.webhook).serverPort | default 443 }}
+    targetPort: {{ (.Values.webhook).serverPort | default 9443 }}
```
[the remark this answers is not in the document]
pregnor: Fixed in https://github.com/banzaicloud/koperator/compare/1df5df1d3d9248a0dff0b6dc4cb311acf210045a..3cb22dd65370093ee382d57ad2402cf05ac5fd93.

tiswanso:
for the koperator Dockerfile, I think the distroless base image should change to `gcr.io/distroless/static-debian11:nonroot`

and possibly need to ensure the deployment of koperator sets the securitycontext:
```
  # capabilities:
  #   drop:
  #   - ALL
  # readOnlyRootFilesystem: true
  # runAsNonRoot: true
  ```

pregnor:
Converted to draft, because there is still stuff to figure out unfortunately, working on it.

tiswanso:
I think this needs to change to `webhook-server` as well for the test instructions to work:  https://github.com/banzaicloud/koperator/blob/a62b459ea3c23056536fdfa1ce06c28a91de0a48/config/base/webhook/service.yaml#L10

pregnor:
For the external listeners with envoy ingress I added the podSecurityContext propagation which is needed to run fine.
With this we should have covered all basic use-cases.

This last one will build on top of an API change that will be merged in #956 , I will update this PR once that is merged and the new API is tagged. That is what breaks the CI now.

pregnor (dismissed):

Kuvesz (dismissed):

pregnor:
I added the new Koperator API version containing the envoy podSecurityContext, this is the final form of this PR.

Kuvesz (approved):

PR #914, opened 2022-12-19
Allow creating KafkaTopic CR for topic that already present on the Kafka cluster

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no |
| New feature?    | yes |
| API breaks?     | no  |
| Deprecations?   | no  |

### What's in this PR?
It allows to create a KafkaTopic CR when the referenced topic is already present on the Kafka cluster.

### Why?

Let the user create KafkaTopic resources for that topic which is already on the Kafka cluster 
Users can modify the topic configuration and partition number for that topic through the created KafkaTopic CR

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/webhooks/kafkatopic_validator.go
+++ b/pkg/webhooks/kafkatopic_validator.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/go-logr/logr"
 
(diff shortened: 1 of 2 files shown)


hi-im-aren (changes requested):
What's the use case behind this? 
- If the topic was created without a CR then it's either an internal topic, which means there is probably an other system managing this topic already
-  or the user created it manually inside kafka which means they know what they are doing and skipped the CR creation on purpose 

Also I think just because we remove the validation from the webhook, kafka will still come back with an error that the topic already exists because our reconciler will try to create the topic in kafka.

panyuenlau:
> What's the use case behind this?

@hi-im-aren  One example use case is for the replicated/migrated Kafka topics: for example, if users want to migrate their legacy Kafka clusters to under Koperator's management, they will need to create the `KafkaTopic` CRs after the replication/migration is done so Koperator can start managing the actual underline Kafka resources

hi-im-aren:
Not sure that the correct way of doing this is removing the validation from the webhook. We are removing a feature because it gets in the way 1% of the time while it's useful in 99% of other use cases.

bartam1:
1. If the topic was created without a CR then it's either an internal topic, which means there is probably an other system managing this topic already
- It will not try to create I checked it. It checks its configuration and when it differs from the actual topic then the Koperator will sync the new configuration to the actual one.
https://github.com/banzaicloud/koperator/blob/635bab34ad604caf9f5b8a6afbab251196fb13d4/controllers/kafkatopic_controller.go#L123
2. 
3. Not sure that the correct way of doing this is removing the validation from the webhook. We are removing a feature because it gets in the way 1% of the time while it's useful in 99% of other use cases:
- Im not sure about this why so useful is. For me It is much better to get all of the topic into KafkaTopic CR so somebody who doesn't like dashboards can check all of the presented topic through kubernetes.  
I also can change partition count and topic configuration for those topics.
Can you collect your points against this modification so we can argue on those points in parking lot or on a meeting to find what would be the best. (I know this is some kind of protection against noob users)

play-io:
> What's the use case behind this?
> 
> * If the topic was created without a CR then it's either an internal topic, which means there is probably an other system managing this topic already
> * or the user created it manually inside kafka which means they know what they are doing and skipped the CR creation on purpose
> 
> Also I think just because we remove the validation from the webhook, kafka will still come back with an error that the topic already exists because our reconciler will try to create the topic in kafka.

This is also my case:
- I started using koperator in 2018 or so
- There was no KafkaTopic cr or maybe I overlooked it
- I'm managing topics with a custom scripts
- However want to migrate to KafkaTopic cr without deletion anything

bartam1:
The problem is when the topic is internal or it is managed by 3rd party application and we let the user to create KafkaTopic CR for it 
1. then the user can modify topic configs (partition number ...). 
2. when it is modified and the kafka or 3rd party application change its configuration then corresponding KafkaTopic will be out of sync.

Solution A
1. We should add some kind of confirmation protection when the user want to create KafkaTopic cr for a present topic. When a KafkaTopic has a specific label like: unmanaged=true,  then the webhook checks is the present topic differs from the KafkaTopic CR and if yes, then the webhook reject that KafkaTopic with error message.
When there is an annotation allowconfigmodify=true then the user can modify the topic values.
2. We cant guarantee that the settings what are in the KafkaTopic is the same as the present topic's settings

Solution B
- We can make a controller which purpose to create KafkaTopic for the kafka topics and do a sync time by time.
There would be unmanaged=true label by default for the internal topics. Users can only change those KafkaTopic CR settings when they specify allowmodifyconfig=true.
- There would be a timestamp when the configuration is synced by the controller.
- the unmanaged=true label cannot be removed (webhook check) from those KafkaTopics that reflect to those topics that are not created by the user.

bartam1:
This PR has been closed because it is re-created here: https://github.com/banzaicloud/koperator/pull/934

PR #920, opened 2023-01-26
Augment error behaviour

mihaialexandrescu:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no|
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets |  |

### What's in this PR?

This PR aims to improve the errors implementations (in the **errorfactory** and **webhooks** packages).  
Along the way, I'm also renaming a couple of variables and functions to better reflect their usage.

The custom error types from the **errorfactory** package can be improved and brought up to date (to Go > 1.13 recommendations) by adding an `Unwrap()` method for each one of these types returning the **embedded** `error` so that errors that could be wrapped inside can be inspected and matched on with all of their details, not just the string representation, with `errors.Is()` and `errors.As()`.  
I'm also adding a unit test for the `Unwrap()` behaviour.

The constant error message strings from the **webhooks** package don't need to be exported, but, because we still have to do some string matching, we can add exported matching functions for all them much like we have right now for 2 of them (`IsAdmissionCantConnect(error) bool`  and  `IsInvalidReplicationFactor(error) bool`). This will facilitate and hide the details of the string matching that we need to do.  
I'm also adding unit tests for these new functions.

The changes made to the `controllers` package reflect the updated usage pattern of the errors from the `errorfactory` package. Without these, that code no longer works correctly.    
The previous model that performed a `type switch` on the error returned by `emperror/errors.Cause(err)` no longer works because the error types from the `errorfactory` package can now Unwrap() and return the errors inside, not just stopping the unwrap immediately and returning the outer type. Because we no longer break the error chain, the "last" error in the chain (deemed as "Cause") is no longer the type from the `errorfactory` package but whatever it wraps or further.  
Now, we can and should use the standard `errors.As(...)` function instead which doesn't require the type we seek to be at the very end of the chain but anywhere inside the chain of wrapped errors. Yes, we currently use the `As(...)` function from the emperror/errors package but that translates directly to the `errors.As(...)` from the **standard library**.

### Additional context

Currently, in koperator we have 2 patterns for our own errors:

1. Custom error types in koperator/pkg/errorfactory/errorfactory.go

These embed the builtin "error" type and wrap external errors passed in to the errorfactory.New function alongside lists of "details". We make use of the `emperror/errors` module which returns error types that can be chained (have both Error() and Unwrap() methods).  
Right now errors like APIFailure (used like in koperator/pkg/k8sutil/resource.go) can be inspected or matched on for type but we lose the details of the error(s) wrapped inside. This can be improved by adding an `Unwrap()` method for each one of these types returning the embedded error.

2. The error strings in koperator/pkg/webhooks/errors.go

These are unexported constant strings used during the validating webhook functions. These are used as parts of the strings of some log and error messages.

These end up used to create either an `error` or a `field.Error` (k8s.io/apimachinery/pkg/util/validation/field). In both situations, the ValidateCreate and ValidateUpdate sets of methods return errors that, underneath, are of `StatusError` type (apimachinery/pkg/api/errors/errors.go) from `apimachinery`. This type hides all other details from the wrapped errors except for the accumulated string. Sometimes the error chain is broken even earlier by other error types from apimachinery.   
The downside to the error types we end up returning from the webhook functions is that we don't have a reasonable solution to avoid doing string matching to discriminate between errors.

### DIFF
--- a/pkg/errorfactory/errorfactory.go
+++ b/pkg/errorfactory/errorfactory.go
@@ -19,6 +19,7 @@ import "emperror.dev/errors"
 // ResourceNotReady states that resource is not ready
 type ResourceNotReady struct{ error }
 
+// Unwrap() allows compliance with Go1.13+ error chaining behavior
 func (e ResourceNotReady) Unwrap() error { return e.error }
 
 // APIFailure states that something went wrong with the api
--- a/pkg/errorfactory/errorfactory.go
+++ b/pkg/errorfactory/errorfactory.go
@@ -19,59 +19,97 @@ import "emperror.dev/errors"
 // ResourceNotReady states that resource is not ready
 type ResourceNotReady struct{ error }
 
+func (e ResourceNotReady) Unwrap() error { return e.error }
+
 // APIFailure states that something went wrong with the api
 type APIFailure struct{ error }
+
+func (e APIFailure) Unwrap() error { return e.error }
 
 // StatusUpdateError states that the operator failed to update the Status
 type StatusUpdateError struct{ error }
 
+func (e StatusUpdateError) Unwrap() error { return e.error }
+
 // BrokersUnreachable states that the given broker is unreachable
 type BrokersUnreachable struct{ error }
+
+func (e BrokersUnreachable) Unwrap() error { return e.error }
 
 // BrokersNotReady states that the broker is not ready
 type BrokersNotReady struct{ error }
 
+func (e BrokersNotReady) Unwrap() error { return e.error }
+
 // BrokersRequestError states that the broker could not understand the request
 type BrokersRequestError struct{ error }
+
+func (e BrokersRequestError) Unwrap() error { return e.error }
 
 // CreateTopicError states that the operator could not create the topic
 type CreateTopicError struct{ error }
 
+func (e CreateTopicError) Unwrap() error { return e.error }
+
 // TopicNotFound states that the given topic is not found
 type TopicNotFound struct{ error }
+
+func (e TopicNotFound) Unwrap() error { return e.error }
 
 // GracefulUpscaleFailed states that the operator failed to update the cluster gracefully
 type GracefulUpscaleFailed struct{ error }
 
+func (e GracefulUpscaleFailed) Unwrap() error { return e.error }
+
 // TooManyResources states that too many resource found
 type TooManyResources struct{ error }
+
+func (e TooManyResources) Unwrap() error { return e.error }
 
 // InternalError states that internal error happened
 type InternalError struct{ error }
 
+func (e InternalError) Unwrap() error { return e.error }
+
 // FatalReconcileError states that a fatal error happened
 type FatalReconcileError struct{ error }
+
+func (e FatalReconcileError) Unwrap() error { return e.error }
 
 // ReconcileRollingUpgrade states that rolling upgrade is reconciling
 type ReconcileRollingUpgrade struct{ error }
 
+func (e ReconcileRollingUpgrade) Unwrap() error { return e.error }
+
 // CruiseControlNotReady states that CC is not ready to receive connection
 type CruiseControlNotReady struct{ error }
+
+func (e CruiseControlNotReady) Unwrap() error { return e.error }
 
 // CruiseControlTaskRunning states that CC task is still running
 type CruiseControlTaskRunning struct{ error }
 
+func (e CruiseControlTaskRunning) Unwrap() error { return e.error }
+
 // CruiseControlTaskRunning states that CC task timed out
 type CruiseControlTaskTimeout struct{ error }
+
+func (e CruiseControlTaskTimeout) Unwrap() error { return e.error }
 
 // CruiseControlTaskFailure states that CC task was not found (CC restart?) or failed
 type CruiseControlTaskFailure struct{ error }
 
+func (e CruiseControlTaskFailure) Unwrap() error { return e.error }
+
 // PerBrokerConfigNotReady states that per-broker configurations has been updated for a broker
 type PerBrokerConfigNotReady struct{ error }
 
+func (e PerBrokerConfigNotReady) Unwrap() error { return e.error }
+
 // LoadBalancerIPNotReady states that the LoadBalancer IP is not yet created
 type LoadBalancerIPNotReady struct{ error }
+
+func (e LoadBalancerIPNotReady) Unwrap() error { return e.error }
 
 // New creates a new error factory error
 func New(t interface{}, err error, msg string, wrapArgs ...interface{}) error {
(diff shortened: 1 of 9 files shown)


pkg/webhooks/errors_test.go
```diff
@@ -19,25 +19,28 @@ import (
 	"testing"
 
 	"emperror.dev/errors"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
```
pregnor, line 22: **Recommended**: Here package names/aliases should also follow the [Golang package naming convention](https://go.dev/blog/package-names) of lowercasing instead of camelCasing. Similarly to `metav1` or `banzaicloudv1alpha1`.
I know we break this rule a thousand times around the code, but at least I want to prevent increasing the count of those.

pregnor (approved):
LGTM with a minor recommendation.

pkg/webhooks/errors_test.go
mihaialexandrescu: It ended up like that because I wanted to standardize that name inside the package and I saw that [kafkacluster_validator.go](https://github.com/banzaicloud/koperator/blob/master/pkg/webhooks/kafkacluster_validator.go#L23) and [kafkatopic_validator.go](https://github.com/banzaicloud/koperator/blob/master/pkg/webhooks/kafkatopic_validator.go#L23) use it camel-cased.

I'll change this instance back.
Would you like me to update the other files in the package too during this PR ?  (to have `apierrors`, not `apiErrors`)

panyuenlau: > Would you like me to update the other files in the package too during this PR ? (to have apierrors, not apiErrors)

+1 It'd be great so we have a cleaner code base

pkg/webhooks/errors.go
```diff
@@ -37,9 +39,44 @@ func IsAdmissionCantConnect(err error) bool {
 	return false
 }
 
+func IsCantConnectAPIServer(err error) bool {
+	if apierrors.IsInternalError(err) && strings.Contains(err.Error(), cantConnectAPIServerMsg) {
+		return true
+	}
+	return false
```
panyuenlau, line 46: Would this be more straightforward?  (same for the other ones that follow this pattern)
```suggestion
	return apierrors.IsInternalError(err) && strings.Contains(err.Error(), cantConnectAPIServerMsg)
```

pkg/webhooks/errors.go
```diff
@@ -37,9 +39,44 @@ func IsAdmissionCantConnect(err error) bool {
 	return false
 }
 
+func IsCantConnectAPIServer(err error) bool {
```
panyuenlau, line 42: The `IsCant` part of the function name seems a bit strange, but since we are following the naming convention as the other functions in this package so I guess it's fine to have it like this

pkg/webhooks/errors_test.go
mihaialexandrescu: Done now in commit 8e3fa9a261b6ba43006fd6f49052f9623f605177

pkg/errorfactory/errorfactory.go
```diff
@@ -19,60 +19,98 @@ import "emperror.dev/errors"
 // ResourceNotReady states that resource is not ready
 type ResourceNotReady struct{ error }
 
+func (e ResourceNotReady) Unwrap() error { return e.error }
```
panyuenlau, line 22: I think it would be helpful to add a short comment on top of this exported method about why we need this so it is more self-explanatory

pkg/webhooks/errors_test.go
```diff
@@ -66,3 +69,126 @@ func TestIsInvalidReplicationFactor(t *testing.T) {
 		t.Error("Expected is invalid replication error to be false, got true")
 	}
 }
+
+func TestIsCantConnectAPIServer(t *testing.T) {
+	testcases := []struct {
+		testname string
```
panyuenlau, line 75: nit - `test` and `name` are two words
```suggestion
		testName string
```

panyuenlau (commented):
Looks great and loving the newly added tests! Just some quick comments

pkg/webhooks/errors.go
mihaialexandrescu: The other "cantConnect" message (that I didn't touch) is `cantConnectErrorMsg` and for that the function name is `IsAdmissionCantConnect`.

I don't necessarily like the `IsCant...` particle but : 
* not sure that `...Admission...` is all that accurate in the context so I avoided copying that part (correct me if I'm wrong!)
* in the k8s repos we often see function names for error comparison like `IsNotFound` which is not really different to this in my opinion so, following that logic, it doesn't feel like a code smell to have that name

pkg/webhooks/errors.go
mihaialexandrescu: I don't know which one, if any, is more idiomatic / Go-like.

I'm down to update those function definitions if it's deemed better / easier to read and understand.

pkg/webhooks/errors_test.go
mihaialexandrescu: I tried to respect the names already in use in some other file but I can't find where I looked.  
I'll update this in all the instances in the file.

pkg/errorfactory/errorfactory.go
mihaialexandrescu: We'd be adding a lot of identical comments (one for each Unwrap method) for what is optional standard library-related behaviour (to be able to do what is described here: https://go.dev/blog/go1.13-errors).

I took a look at this example in the standard library: https://cs.opensource.google/go/go/+/refs/tags/go1.19.5:src/io/fs/fs.go;l=252

I'm not opposed to adding this but I'm not convinced it helps and I struggle to imagine a nice comment ... like "Unwrap() allows compliance with Go1.13+ error chaining behaviour" (many times) ?

panyuenlau: Well, you don't have to add the same comment in every `Unwrap()` function IMO, just adding the comment to the first one in the file would be enough and I think "Unwrap() allows compliance with Go1.13+ error chaining behavior" is a great comment to add - at least people would know what this is for without chasing back which interface we are trying to implement with adding this method

mihaialexandrescu: Sounds good to me. Thanks!

Done in a99daa5a12a2e2d172c964f4e7f4b360a6ae1816.

pkg/webhooks/errors.go
mihaialexandrescu: Brevity is probably good in this case. Thanks!

Done in a99daa5a12a2e2d172c964f4e7f4b360a6ae1816.

pkg/webhooks/errors.go
panyuenlau: I don't really have a good suggestion either, so I am fine with leaving it like this

panyuenlau (approved):

pregnor (approved):

PR #923, opened 2023-01-31
Fix Kafka super.users override

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |

### What's in this PR?

Merges custom super.users configuration with the Koperator auto generated one.

### Why?

When there is custom super.users configuration in the KafkaCluster CR and there is SSL listener, Koperator generated super.user configuration overrides the custom one. 

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/resources/kafka/configmap.go
+++ b/pkg/resources/kafka/configmap.go
@@ -378,6 +378,8 @@ func mergeSuperUsersPropertyValue(source *properties.Properties, target *propert
 
 	if inserted {
 		mergedSuperUsers := strings.Join(targetSuperUsers, ";")
+		// Setting string value for a property is not going to run into error, also we don't have error handling at upper levels
+		//nolint:errcheck
 		target.Set("super.users", mergedSuperUsers)
 	}
 }
--- a/pkg/resources/kafka/configmap.go
+++ b/pkg/resources/kafka/configmap.go
@@ -346,15 +346,16 @@ func generateListenerSSLConfig(config *properties.Properties, name string, sslCl
 	}
 }
 
-// mergeSuperUsersPropertyValue merges the super.users property value from the source into the target properties
-func mergeSuperUsersPropertyValue(source *properties.Properties, target *properties.Properties) {
+// mergeSuperUsersPropertyValue merges the target and source super.users property value, and returns it as string.
+// It returns empty string when there were no updates or any  of the super.users property value was empty.
+func mergeSuperUsersPropertyValue(source *properties.Properties, target *properties.Properties) string {
 	sourceVal, foundSource := source.Get("super.users")
 	if !foundSource || sourceVal.IsEmpty() {
-		return
+		return ""
 	}
 	targetVal, foundTarget := target.Get("super.users")
 	if !foundTarget || targetVal.IsEmpty() {
-		return
+		return ""
 	}
 
 	sourceSuperUsers := strings.Split(sourceVal.Value(), ";")
@@ -377,11 +378,10 @@ func mergeSuperUsersPropertyValue(source *properties.Properties, target *propert
 	}
 
 	if inserted {
-		mergedSuperUsers := strings.Join(targetSuperUsers, ";")
-		// Setting string value for a property is not going to run into error, also we don't have error handling at upper levels
-		//nolint:errcheck
-		target.Set("super.users", mergedSuperUsers)
+		return strings.Join(targetSuperUsers, ";")
 	}
+
+	return ""
 }
 
 func (r Reconciler) generateBrokerConfig(id int32, brokerConfig *v1beta1.BrokerConfig, extListenerStatuses,
@@ -394,9 +394,13 @@ func (r Reconciler) generateBrokerConfig(id int32, brokerConfig *v1beta1.BrokerC
 
 	// Merge operator generated configuration to the final one
 	if opGenConf != nil {
-		// When there is custom super.users configuration we merge its value into the Koperator generated one
-		// thus finalBrokerConfig contains merged super.users value.
-		mergeSuperUsersPropertyValue(finalBrokerConfig, opGenConf)
+		// When there is custom super.users configuration we merge its value with the Koperator generated one
+		// to avoid overwrite that happens when the finalBrokerConfig.Merge(opGenConf) is called.
+		if suMerged := mergeSuperUsersPropertyValue(finalBrokerConfig, opGenConf); suMerged != "" {
+			// Setting string value for a property is not going to run into error, also we don't return error in this function
+			//nolint:errcheck
+			opGenConf.Set("super.users", suMerged)
+		}
 		finalBrokerConfig.Merge(opGenConf)
 	}
 
--- a/pkg/resources/kafka/configmap.go
+++ b/pkg/resources/kafka/configmap.go
@@ -392,8 +392,8 @@ func (r Reconciler) generateBrokerConfig(id int32, brokerConfig *v1beta1.BrokerC
 
 	// Merge operator generated configuration to the final one
 	if opGenConf != nil {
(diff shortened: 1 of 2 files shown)


pkg/resources/kafka/configmap.go
```diff
@@ -346,6 +346,44 @@ func generateListenerSSLConfig(config *properties.Properties, name string, sslCl
 	}
 }
 
+// mergeSuperUsersPropertyValue merges the super.users property value from the source into the target properties
+func mergeSuperUsersPropertyValue(source *properties.Properties, target *properties.Properties) {
```
pregnor, line 350: **Recommended**: In Go it is more idiomatic to pass two values and return a third value which can be used to overwrite any of the inputs at the caller side if needed (or assign to a new variable) instead of creating side-effect ridden functions with pseudo-references (pointers).
For example append works similarly `newSlice = append(oldSlice, newValue)`.

Is there a reason to prefer writing pointers as a side effect in a method here compared to using a proper function and returning a new value?

pkg/resources/kafka/configmap.go
```diff
@@ -356,6 +394,9 @@ func (r Reconciler) generateBrokerConfig(id int32, brokerConfig *v1beta1.BrokerC
 
 	// Merge operator generated configuration to the final one
 	if opGenConf != nil {
+		// When there is custom super.users configuration we merge its value into the Koperator generated one
+		// thus finalBrokerConfig contains merged super.users value.
+		mergeSuperUsersPropertyValue(finalBrokerConfig, opGenConf)
```
pregnor, line 399: **Required**: We **MUST** either update the comment here or update the function logic because with the current implementation the `target` contains the final value (line 383) which among the arguments here is `opGenConf`, not `finalBrokerConfig`, but the last line of the comment makes it sound it should be `finalBrokerConfig` that contains the merged super.users value.
(Whichever the following logic builds upon should be the consistent truths in both logic and comment.)

pregnor: Oh I see the next line we merge the other way, that spins it a little.
I think we should try to find a better phrasing still, but I understand where it went wrong here first.

pkg/resources/kafka/configmap.go
```diff
@@ -346,6 +346,44 @@ func generateListenerSSLConfig(config *properties.Properties, name string, sslCl
...
+	}
+
+	if inserted {
+		mergedSuperUsers := strings.Join(targetSuperUsers, ";")
+		// Setting string value for a property is not going to run into error, also we don't have error handling at upper levels
+		//nolint:errcheck
+		target.Set("super.users", mergedSuperUsers)
+	}
+}
+
```
pregnor, line 386: **Recommended**: I think the current implementation is a little more complex than it should be.
For a `len(source) == n` and `len(target) == m` and `intersect(source, target) == o` the complexity is somewhere between `O(n*m)` and `O(n*(m-o))`.
1. If the super.Users lists were sorted (or could be sorted and not taking into consideration the complexity of the sorting) then a simple merging algorithm would work faster here (iterating both slices with one cursor taking the earliest item), namely `O(n+m)`.
2. or If the super.Users could not be sorted easily then we could still use a map to merge the unique values into one comprehensive list which would make the complexity `O(n+m)` with an extra `union(n, m) < n+m` memory requirement.

pkg/resources/kafka/configmap.go
bartam1: I wanted to put everything in that mergeSuperUsersPropertyValue function thus it works almost the same as the finalBrokerConfig.Merge(opGenConf).

It can be a good idea to follow the append convention. Like  mergeSuperUsersPropertyValue(source *properties.Properties, superusers) return *properties.Properties

pkg/resources/kafka/configmap.go
bartam1: On paper you have right regarding performance and complexity, but in practically in our case my method is the fastest.
The hash function also has complexity and more memory allocated when u create a map data structure.
Sorting also has complexity as you said. 
In our case because of the small sizes of the source and target slices my method is the fastest and it has the lowest memory requirements.

BenchmarkMapMethod-8             2700618               447.6 ns/op           432 B/op          5 allocs/op
BenchmarkMyMethod-8             11876641                99.88 ns/op          128 B/op          1 allocs/op
BenchmarkSortMethod-8            2892297               412.7 ns/op           544 B/op          7 allocs/op

Other question which one is the most beautiful solution :)

pregnor: Try with the following mapMethod pls:

```golang
func mapMethod(a []string, b []string) (res []string) {
	check := make(map[string]interface{}, len(a)+len(b))
	res = make([]string, 0, len(a)+len(b))
	
	for _, item := range a {
		check[item] = nil
	}
	
	for _, item := range b {
		check[item] = nil
	}

	for letter, _ := range check {
		res = append(res, letter)
	}

	return res
}
```

If the result is similar then I let this go on the basis of being opinionated.

bartam1: When the input is:
a = []string{"alma", "korte", "User:CN=kafka-kafka-cluster", "User:CN=kafka-kafka-operator", "User:CN=supertubes-system-supertubes", "User:CN=supertubes-system-supertubes-ui"}
b = []string{"User:CN=kafka-kafka-cluster", "User:CN=kafka-kafka-operator", "User:CN=supertubes-system-supertubes", "User:CN=supertubes-system-supertubes-ui"}
result:
BenchmarkMapMethod-8            33648338               351.9 ns/op           432 B/op          5 allocs/op
BenchmarkMapMethod2-8           33422253               356.9 ns/op           736 B/op          2 allocs/op
BenchmarkMyMethod-8             148691840               80.95 ns/op          128 B/op          1 allocs/op
BenchmarkSortMethod-8           35608514               332.0 ns/op           544 B/op          7 allocs/op

When 
a =  []string{"alma", "korte"}
b = []string{"User:CN=kafka-kafka-cluster", "User:CN=kafka-kafka-operator", "User:CN=supertubes-system-supertubes", "User:CN=supertubes-system-supertubes-ui"}

BenchmarkMapMethod-8            38433975               315.2 ns/op           336 B/op          5 allocs/op
BenchmarkMapMethod2-8           70905468               167.6 ns/op            96 B/op          1 allocs/op
BenchmarkMyMethod-8             218245809               54.87 ns/op          128 B/op          1 allocs/op
BenchmarkSortMethod-8           51449380               234.7 ns/op           288 B/op          6 allocs/op

pregnor: Fine, I let this go.

pregnor (dismissed):

pkg/resources/kafka/configmap.go
```diff
@@ -346,6 +346,44 @@ func generateListenerSSLConfig(config *properties.Properties, name string, sslCl
 	}
 }
 
+// mergeSuperUsersPropertyValue merges the target and source super.users property value, and returns it as string.
+// It returns empty string when there were no updates or any  of the super.users property value was empty.
```
panyuenlau, line 350: nit - unintentional extra space in comment

```suggestion
// It returns empty string when there were no updates or any of the super.users property value was empty.
```

panyuenlau (dismissed):

pkg/resources/kafka/configmap.go
panyuenlau: You don't have to fix it now - I can fix it in my refactoring PR https://github.com/banzaicloud/koperator/pull/917 after yours is merged

bartam1: Thank you! I think it is better to fix it here.

pregnor (approved):

PR #925, opened 2023-02-02
Rename webhooks errors exported funcs

mihaialexandrescu:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no |
| New feature?    | no |
| API breaks?     | yes |
| Deprecations?   | no |
| Related tickets |  |

### What's in this PR?

This PR renames the exported error checking functions from pkg `webhooks` to names that include the `IsAdmission...` prefix, rather than `Is...`, to hopefully make it clearer that they relate to admission webhooks errors and results.

The change is initially done in 2 commits : 
1. the non-breaking renames -- those functions were added very recently in https://github.com/banzaicloud/koperator/pull/920 (after the v0.22.0 release)
2. a breaking change  --  to keep the naming scheme consistent in the package, the `IsInvalidReplicationFactor` function (which already exists in release v0.22.0) is **renamed** following the same pattern

### Why?

More clarity and consistency in the naming scheme in the "webhooks" package

### API-breaking markers ?
I am not sure what markers I need to use to correctly flag the breaking change in API, other than the table at the top of the PR description.

### Still one question about renaming ...
Should the `IsErrorDuringValidation` function follow the same pattern since it includes the word `validation` already ... ? (aka :  `IsAdmissionErrorDuringValidation`)   
Currently, I favour renaming throughout and it is done for this one too.

### DIFF
--- a/pkg/resources/cruisecontrol/topicManager.go
+++ b/pkg/resources/cruisecontrol/topicManager.go
@@ -99,7 +99,7 @@ func generateCCTopic(cluster *v1beta1.KafkaCluster, client client.Client, log lo
 					return errorfactory.New(errorfactory.ResourceNotReady{}, err, "topic admission failed to connect to kafka cluster")
 				}
 				// If less than the required brokers are available - return not ready
-				if webhooks.IsInvalidReplicationFactor(err) {
+				if webhooks.IsAdmissionInvalidReplicationFactor(err) {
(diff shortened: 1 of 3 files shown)


pregnor:
> Should the IsErrorDuringValidation function follow the same pattern since it includes the word validation already ... ? (aka : IsAdmissionErrorDuringValidation)

I vote yes

pregnor (approved):

Kuvesz (approved):

PR #928, opened 2023-02-09
Prevent kafka controller from running into NodePort service deletion and re-creation cycles indefinitely

### LINKED ISSUE #922: External NodePort services got killed and recreated for every several seconds.
**Describe the bug**
We are using koperator 0.22.0. After installed a kafka cluster with two borkers, the kafka-cluster is stuck in ClusterReCounciling state
```
NAME            CLUSTER STATE        CLUSTER ALERT COUNT   LAST SUCCESSFUL UPGRADE   UPGRADE ERROR COUNT   AGE
kafka-cluster   ClusterReconciling   0                                               0                     3d19h
```
And the two nodePort services restart every several seconds. 
```
NAME                              TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)                                 AGE
kafka-cluster-0-external          NodePort    10.233.5.235    <none>        9094:30091/TCP                          8s
kafka-cluster-1-external          NodePort    10.233.20.3     <none>        9094:30092/TCP                          8s
kafka-cluster-cruisecontrol-svc   ClusterIP   10.233.35.58    <none>        8090/TCP,9020/TCP                       3d19h
kafka-cluster-headless            ClusterIP   None            <none>        29092/TCP,29093/TCP,9094/TCP,9020/TCP   3d19h
kafka-operator-alertmanager       ClusterIP   10.233.11.110   <none>        9001/TCP                                3d19h
kafka-operator-authproxy          ClusterIP   10.233.63.33    <none>        8443/TCP                                3d19h
kafka-operator-operator           ClusterIP   10.233.10.195   <none>        443/TCP                                 3d19h
```
The client got disconnected constantly from the broker.

** Steps to reproduce the issue**
Installed the kafka-operator and kafka-cluster with version 0.22.0 

**Additional context**
We suspect it's caused by https://github.com/banzaicloud/koperator/commit/28a116898cb89264a761a2bfe1d82c40a9528344, the two services got deleted and recreated with every reconciling flow.

panyuenlau:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets | fixes #922 |

### What's in this PR?

To fix #922

### Why?

When using `NodePort` as the service type for external listeners, the corresponding NodePort services get into deletion-recreation cycles indefinitely

### Additional context

Also refactor the relevant logic - moving the headless service related logic from `pkg/resources/kafka/allBrokerService.go` to `pkg/resources/kafka/headlessService.go` along with additional unit tests 

### Checklist

- [x] Implementation tested
~- [x] User guide and development docs updated (if needed)~

### DIFF
--- a/pkg/resources/kafka/allBrokerService.go
+++ b/pkg/resources/kafka/allBrokerService.go
@@ -15,18 +15,7 @@
 package kafka
 
 import (
-	"context"
 	"fmt"
-
-	"emperror.dev/errors"
-
-	"github.com/banzaicloud/koperator/api/v1beta1"
-
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/selection"
-
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -63,68 +52,4 @@ func (r *Reconciler) allBrokerService() runtime.Object {
 			Ports:           usedPorts,
 		},
 	}
-}
-
-// deleteNonHeadlessServices deletes the all-broker service that was created for the current KafkaCluster
-// if there is any and also the service of each broker
-func (r *Reconciler) deleteNonHeadlessServices() error {
-	ctx := context.Background()
-
-	svc := corev1.Service{
-		TypeMeta: metav1.TypeMeta{
-			APIVersion: corev1.SchemeGroupVersion.String(),
-			Kind:       "Service",
-		},
-		ObjectMeta: metav1.ObjectMeta{
-			Namespace: r.KafkaCluster.GetNamespace(),
-			Name:      fmt.Sprintf(kafkautils.AllBrokerServiceTemplate, r.KafkaCluster.GetName()),
-		},
-	}
-
-	err := r.Client.Delete(ctx, &svc)
-	if err != nil && client.IgnoreNotFound(err) != nil {
-		return err
-	}
-
-	// delete broker services
-	labelSelector := labels.NewSelector()
-	for k, v := range apiutil.LabelsForKafka(r.KafkaCluster.GetName()) {
-		req, err := labels.NewRequirement(k, selection.Equals, []string{v})
-		if err != nil {
-			return err
-		}
-		labelSelector = labelSelector.Add(*req)
-	}
-
-	// add "has label 'brokerId' to matching labels selector expression
-	req, err := labels.NewRequirement(v1beta1.BrokerIdLabelKey, selection.Exists, nil)
-	if err != nil {
-		return err
-	}
-	labelSelector = labelSelector.Add(*req)
-
-	var services corev1.ServiceList
-	err = r.Client.List(ctx, &services,
-		client.InNamespace(r.KafkaCluster.GetNamespace()),
-		client.MatchingLabelsSelector{Selector: labelSelector},
-	)
-
-	if err != nil {
-		return errors.WrapIfWithDetails(err, "failed to list services",
-			"namespace", r.KafkaCluster.GetNamespace(),
-			"label selector", labelSelector.String())
-	}
-
-	for i := range services.Items {
-		svc = services.Items[i]
-		if !svc.GetDeletionTimestamp().IsZero() {
-			continue
-		}
-		err = r.Client.Delete(ctx, &svc)
-		if err != nil && client.IgnoreNotFound(err) != nil {
-			return err
-		}
-	}
-
-	return nil
 }
--- a/pkg/resources/kafka/configmap.go
+++ b/pkg/resources/kafka/configmap.go
@@ -129,7 +129,7 @@ func (r *Reconciler) getConfigProperties(bConfig *v1beta1.BrokerConfig, id int32
 	mountPathsMerged, isMountPathRemoved := mergeMountPaths(mountPathsOld, mountPathsNew)
 
 	if isMountPathRemoved {
-		log.Error(errors.New("removed storage is found in the KafkaCluster CR"), "removing storage from broker is not supported", v1beta1.BrokerIdLabelKey, id, "mountPaths", mountPathsOld, "mountPaths in externalListeners CR ", mountPathsNew)
+		log.Error(errors.New("removed storage is found in the KafkaCluster CR"), "removing storage from broker is not supported", v1beta1.BrokerIdLabelKey, id, "mountPaths", mountPathsOld, "mountPaths in kafkaCluster CR ", mountPathsNew)
 	}
 
 	if len(mountPathsMerged) != 0 {
--- a/pkg/resources/kafka/configmap.go
+++ b/pkg/resources/kafka/configmap.go
@@ -346,6 +346,44 @@ func generateListenerSSLConfig(config *properties.Properties, name string, sslCl
 	}
 }
 
+// mergeSuperUsersPropertyValue merges the target and source super.users property value, and returns it as string.
+// It returns empty string when there were no updates or any of the super.users property value was empty.
+func mergeSuperUsersPropertyValue(source *properties.Properties, target *properties.Properties) string {
+	sourceVal, foundSource := source.Get("super.users")
+	if !foundSource || sourceVal.IsEmpty() {
+		return ""
+	}
+	targetVal, foundTarget := target.Get("super.users")
+	if !foundTarget || targetVal.IsEmpty() {
+		return ""
+	}
+
+	sourceSuperUsers := strings.Split(sourceVal.Value(), ";")
+	targetSuperUsers := strings.Split(targetVal.Value(), ";")
+
+	inserted := false
+	for _, sourceSuperUser := range sourceSuperUsers {
+		found := false
+		for _, targetSuperUser := range targetSuperUsers {
+			if sourceSuperUser == targetSuperUser {
+				found = true
+				break
+			}
+		}
+
+		if !found {
+			inserted = true
+			targetSuperUsers = append(targetSuperUsers, sourceSuperUser)
+		}
+	}
+
+	if inserted {
+		return strings.Join(targetSuperUsers, ";")
+	}
+
+	return ""
+}
+
 func (r Reconciler) generateBrokerConfig(id int32, brokerConfig *v1beta1.BrokerConfig, extListenerStatuses,
 	intListenerStatuses, controllerIntListenerStatuses map[string]v1beta1.ListenerStatusList,
 	serverPasses map[string]string, clientPass string, superUsers []string, log logr.Logger) string {
@@ -356,6 +394,13 @@ func (r Reconciler) generateBrokerConfig(id int32, brokerConfig *v1beta1.BrokerC
 
 	// Merge operator generated configuration to the final one
 	if opGenConf != nil {
+		// When there is custom super.users configuration we merge its value with the Koperator generated one
+		// to avoid overwrite that happens when the finalBrokerConfig.Merge(opGenConf) is called.
+		if suMerged := mergeSuperUsersPropertyValue(finalBrokerConfig, opGenConf); suMerged != "" {
+			// Setting string value for a property is not going to run into error, also we don't return error in this function
+			//nolint:errcheck
+			opGenConf.Set("super.users", suMerged)
+		}
 		finalBrokerConfig.Merge(opGenConf)
 	}
 
--- a/pkg/resources/kafka/configmap.go
+++ b/pkg/resources/kafka/configmap.go
@@ -129,7 +129,7 @@ func (r *Reconciler) getConfigProperties(bConfig *v1beta1.BrokerConfig, id int32
 	mountPathsMerged, isMountPathRemoved := mergeMountPaths(mountPathsOld, mountPathsNew)
 
 	if isMountPathRemoved {
-		log.Error(errors.New("removed storage is found in the KafkaCluster CR"), "removing storage from broker is not supported", v1beta1.BrokerIdLabelKey, id, "mountPaths", mountPathsOld, "mountPaths in kafkaCluster CR ", mountPathsNew)
+		log.Error(errors.New("removed storage is found in the KafkaCluster CR"), "removing storage from broker is not supported", v1beta1.BrokerIdLabelKey, id, "mountPaths", mountPathsOld, "mountPaths in externalListeners CR ", mountPathsNew)
 	}
 
 	if len(mountPathsMerged) != 0 {
--- a/pkg/resources/kafka/headlessService.go
+++ b/pkg/resources/kafka/headlessService.go
@@ -153,8 +153,7 @@ func (r *Reconciler) deleteNonHeadlessServices(ctx context.Context) error {
 		filteredSvcsToDelete = nonNodePortServices(services)
 	}
 
-	for i := range filteredSvcsToDelete.Items {
-		svc = filteredSvcsToDelete.Items[i]
+	for _, svc := range filteredSvcsToDelete.Items {
 		if !svc.GetDeletionTimestamp().IsZero() {
 			continue
 		}
--- a/pkg/resources/kafka/headlessService.go
+++ b/pkg/resources/kafka/headlessService.go
@@ -18,11 +18,16 @@ import (
 	"context"
 	"fmt"
 
+	"emperror.dev/errors"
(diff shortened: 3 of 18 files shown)


panyuenlau:
Also found another minor issue during development - if the external listeners were set to `Loadbalancer` and then changed to `Nodeport`, the operator doesn't remove the unused loadbalancer service

Will open another PR to address this issue since it's not part of #922

pkg/resources/kafka/headlessService_test.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	for _, test := range testCases {
+		testReconciler.KafkaCluster = test.kafkaCluster
+		got := testReconciler.checkIfNodePortSvcNeeded()
+		t.Run(test.testName, func(t *testing.T) {
+			if got != test.isNodePortSvcNeeded {
+				t.Errorf("Expected: %v  Got: %v", test.isNodePortSvcNeeded, got)
+			}
+		})
```
pregnor, line 98: **Recommended**: We **SHOULD** consider using [`testify/require`](https://pkg.go.dev/github.com/stretchr/testify/require) for improved readability of the test check. I'm going to combine this with my previous comments in the code suggestion.

```suggestion
		t.Run(test.testName, func(t *testing.T) {
			actualHasAnyNodePortAccessMethodExternalListener := testReconciler.checkIfNodePortSvcNeeded()
			require.Equals(t, test.expectedHasAnyNodePortAccessMethodExternalListener, actualHasAnyNodePortAccessMethodExternalListener)
		})
```

Testify also provides a similar error message you manually constructed using the `Equal(...)` function:
```
# https://go.dev/play/p/FqfUImA461T
=== RUN   TestLastIndex/single_item_slice,_last_index_is_first_and_last
    prog_test.go:39: 
        	Error Trace:	/prog_test.go:39
        	Error:      	Not equal: 
        	            	expected: 0
        	            	actual  : -1
        	Test:       	TestLastIndex/single_item_slice,_last_index_is_first_and_last
```

pkg/resources/kafka/headlessService_test.go
```diff
@@ -0,0 +1,242 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/banzaicloud/koperator/api/v1beta1"
+)
+
```
pregnor, line 26: **Note**: most of my comments in this file stand true for both test functions, but I only commented them on the first.

pkg/resources/kafka/headlessService_test.go
```diff
@@ -0,0 +1,242 @@
...
+						},
+					},
+					{
+						ObjectMeta: metav1.ObjectMeta{
+							Name: "NodePort service 2",
+						},
+						Spec: corev1.ServiceSpec{
+							Type: corev1.ServiceTypeNodePort,
+						},
+					},
```
pregnor, line 127: **Optional**: I **MAY** have considered moving one of node port services to a later point or maybe even create a 3rd one and have a `[firstNodePort, nonNodePort, middleNodePort, nonNodePort2, lastNodePort]` setup in the slice to make sure we handle items uniformly regardless of their position in the slice (and may have created a `[nonNodePort, middleNodePort, nonNodePort2]` case as well) because those logic errors in the function implementation always come unexpected when the initial implementation is modified and the logic is made more complex.
Not necessary here as we can pay attention to it the next time the function gets modified, just a highlight.

pregnor (commented):
Don't be discouraged by the comments, as those are just recommendations and optional stuff, I would be okay approving the PR as is, I just feel like we could still add some worthwhile improvements at relatively low cost.

Also I don't keep my suggestions in me, rather write it out even if it is just a personal practice (some optionals), but if you disagree with anything - that is not arequired change of which there is non right now - I'm not attached to these and am ready to drop those.

pkg/resources/kafka/headlessService.go
```diff
@@ -92,3 +97,96 @@ func (r *Reconciler) deleteHeadlessService() error {
...
+		if externalListener.GetAccessMethod() == corev1.ServiceTypeNodePort {
+			return true
+		}
+	}
+	return false
+}
+
+func getNonNodePortSvc(services corev1.ServiceList) corev1.ServiceList {
+	var svc corev1.Service
+	var nonNodePortSvc corev1.ServiceList
```
pregnor, line 182: **Recommended**: We **SHOULD** consider calling this plural because it refers to a list of services - or alternatively call it a list, but I think that is too verbose and unnecessary, it is a convenience that we use a list struct and not the slice directly - because currently the name itself would imply this designates a single service while that is not the case.

Apologies I missed this on the first review, only caught it while reviewing my comments.

pkg/resources/kafka/headlessService.go
```diff
@@ -92,3 +97,96 @@ func (r *Reconciler) deleteHeadlessService() error {
...
+			continue
+		}
+		err = r.Client.Delete(ctx, &svc)
+		if err != nil && client.IgnoreNotFound(err) != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
panyuenlau: By refactoring, do you mean to abstract the some of the logic into a (or multiple) functions? (I thought about this as well)

pkg/resources/kafka/headlessService_test.go
```diff
@@ -0,0 +1,242 @@
...
+				},
+			},
+			isNodePortSvcNeeded: true,
+		},
+	}
+
+	for _, test := range testCases {
+		testReconciler.KafkaCluster = test.kafkaCluster
+		got := testReconciler.checkIfNodePortSvcNeeded()
```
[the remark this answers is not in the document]
panyuenlau: Thanks for catching this, I meant to put it in the subcase run, but for got

pkg/resources/kafka/headlessService.go
```diff
@@ -92,3 +97,96 @@ func (r *Reconciler) deleteHeadlessService() error {
...
+func (r *Reconciler) checkIfNodePortSvcNeeded() bool {
+	for _, externalListener := range r.KafkaCluster.Spec.ListenersConfig.ExternalListeners {
+		if externalListener.GetAccessMethod() == corev1.ServiceTypeNodePort {
+			return true
+		}
+	}
+	return false
+}
+
+func getNonNodePortSvc(services corev1.ServiceList) corev1.ServiceList {
```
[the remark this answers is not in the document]
panyuenlau: renamed to `nonNodePortServices()`

pkg/resources/kafka/headlessService.go
```diff
@@ -92,3 +97,96 @@ func (r *Reconciler) deleteHeadlessService() error {
...
+	var nonNodePortSvc corev1.ServiceList
+
+	for i := range services.Items {
+		svc = services.Items[i]
+		if svc.Spec.Type != corev1.ServiceTypeNodePort {
+			nonNodePortSvc.Items = append(nonNodePortSvc.Items, svc)
+		}
+	}
+
+	return nonNodePortSvc
```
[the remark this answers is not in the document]
panyuenlau: Thx! Updated

pkg/resources/kafka/headlessService.go
```diff
@@ -92,3 +97,96 @@ func (r *Reconciler) deleteHeadlessService() error {
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// checkIfNodePortSvcNeeded returns true when users specify any of the external listeners to use NodePort
+func (r *Reconciler) checkIfNodePortSvcNeeded() bool {
+	for _, externalListener := range r.KafkaCluster.Spec.ListenersConfig.ExternalListeners {
```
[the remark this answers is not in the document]
panyuenlau: Updated the function signature to take a slice of `ExternalListenerConfig`

pkg/resources/kafka/headlessService.go
```diff
@@ -92,3 +97,96 @@ func (r *Reconciler) deleteHeadlessService() error {
...
+		if err != nil && client.IgnoreNotFound(err) != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// checkIfNodePortSvcNeeded returns true when users specify any of the external listeners to use NodePort
+func (r *Reconciler) checkIfNodePortSvcNeeded() bool {
```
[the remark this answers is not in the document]
panyuenlau: Renamed it to `isNodePortAccessMethodInUseAmongExternalListeners` if this sounds good to you

pkg/resources/kafka/headlessService_test.go
```diff
@@ -0,0 +1,242 @@
...
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/banzaicloud/koperator/api/v1beta1"
+)
+
+func TestCheckIfNodePortSvcNeeded(t *testing.T) {
+	testReconciler := Reconciler{}
```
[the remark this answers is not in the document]
panyuenlau: Fixed as since the function signature has been updated

pkg/resources/kafka/headlessService_test.go
panyuenlau: I forgot to use `require.Equal` here as I updated the second function to use it rather than `t.Errorf`

pkg/resources/kafka/configmap.go
```diff
@@ -129,7 +129,7 @@ func (r *Reconciler) getConfigProperties(bConfig *v1beta1.BrokerConfig, id int32
 	mountPathsMerged, isMountPathRemoved := mergeMountPaths(mountPathsOld, mountPathsNew)
 
 	if isMountPathRemoved {
-		log.Error(errors.New("removed storage is found in the KafkaCluster CR"), "removing storage from broker is not supported", v1beta1.BrokerIdLabelKey, id, "mountPaths", mountPathsOld, "mountPaths in kafkaCluster CR ", mountPathsNew)
+		log.Error(errors.New("removed storage is found in the KafkaCluster CR"), "removing storage from broker is not supported", v1beta1.BrokerIdLabelKey, id, "mountPaths", mountPathsOld, "mountPaths in externalListeners CR ", mountPathsNew)
```
bartam1, line 132: I think this is a mistake :)

pkg/resources/kafka/allBrokerService.go
```diff
@@ -64,67 +53,3 @@ func (r *Reconciler) allBrokerService() runtime.Object {
 		},
 	}
 }
-
-// deleteNonHeadlessServices deletes the all-broker service that was created for the current KafkaCluster
-// if there is any and also the service of each broker
-func (r *Reconciler) deleteNonHeadlessServices() error {
```
bartam1, line 70: I would not move this function from the allBrokerService because basicly deleteNonHeadlessServices = deleteAllBrokerService and it is more connected to the func (r *Reconciler) allBrokerService() function.
I would rename the deleteNonHeadlessServices to deleteAllBrokerService.

We can se the same for the headlessService - deleteHeadlessService() in the headlessService.go

panyuenlau: Well, if you look into the details of this function, it is not just to delete the all broker service, it also deletes the services like the `NodePort` services, and LoadBalancer (the implementation for this will be needed in another PR)

pkg/resources/kafka/configmap.go
panyuenlau: Thanks for catching this! This was not an intended change, not sure how the change somehow appeared 😅

Kuvesz (approved):
Looks good to me, thanks for also adding tests!

pkg/resources/kafka/allBrokerService.go
bartam1: You have right regarding this function deletes not only the all broker service but deletes those services which belongs to all broker services (https://github.com/banzaicloud/koperator/blob/71000a41df4fe3c1565217df9ff75ae9b9adbec6/pkg/resources/kafka/kafka.go#L385)
Then I would say deleteAllBrokerServices not deleteAllBrokerService.
I understand your logic.  I think both can be good.

panyuenlau: Actually I think the `allBrokerService.go` is meant for the `"%s-all-broker"` service only, not the "services for all the brokers". That's why I don't think it makes sense to put anything that is not related to the `"%s-all-broker"` service to that file

> I know this is confusing 😕

bartam1: It is a little bit weird "non headless serices for all brokers" :)

pkg/resources/kafka/headlessService.go
```diff
@@ -92,3 +97,95 @@ func (r *Reconciler) deleteHeadlessService() error {
...
+	}
+
+	// if NodePort is used for any of the external listeners, the corresponding services need to remain
+	// so that clients from outside the Kubernetes cluster can reach the brokers
+	filteredSvcsToDelete := services
+	if isNodePortAccessMethodInUseAmongExternalListeners(r.KafkaCluster.Spec.ListenersConfig.ExternalListeners) {
+		filteredSvcsToDelete = nonNodePortServices(services)
+	}
+
+	for i := range filteredSvcsToDelete.Items {
```
bartam1, line 156: Here we can use _, svc = range filteredSvcsToDelete.Items

pkg/resources/kafka/allBrokerService.go
panyuenlau: agreed, I can update it to "failed to delete non-headless services for all of the brokers" to make it less confusing

panyuenlau: log message updated

pkg/resources/kafka/headlessService.go
```diff
@@ -92,3 +97,95 @@ func (r *Reconciler) deleteHeadlessService() error {
...
+
+	if err != nil {
+		return errors.WrapIfWithDetails(err, "failed to list services",
+			"namespace", r.KafkaCluster.GetNamespace(),
+			"label selector", labelSelector.String())
+	}
+
+	// if NodePort is used for any of the external listeners, the corresponding services need to remain
+	// so that clients from outside the Kubernetes cluster can reach the brokers
+	filteredSvcsToDelete := services
```
bartam1, line 151: I think we dont have to use another variable for this.
we can use the services all way long.
later: services = nonNodePortServices(services)
...

pkg/resources/kafka/allBrokerService.go
bartam1: Thank you! :)

pkg/resources/kafka/headlessService.go
panyuenlau: yea...but having this little extra variable named `filteredSvcsToDelete` would make the implementation more readable

pkg/resources/kafka/headlessService.go
panyuenlau: right... I tried not to touch the implementation that was copied over. But updated in commit a4a8fb53ccba2bb8dd6be6d2c809ca778f005bb0

pkg/resources/kafka/headlessService.go
bartam1: It is clear enough to me services = nonNodePortServices(services).
It is just my opinion and a suggestion. Your solution is ok for me.

bartam1 (approved):

pkg/resources/kafka/headlessService.go
bartam1: We can also use here (https://github.com/banzaicloud/koperator/blob/71000a41df4fe3c1565217df9ff75ae9b9adbec6/pkg/resources/kafka/headlessService.go#L137)  name like: brokerServices
and  brokerServices = nonNodePortServices(brokerServices).

This https://github.com/banzaicloud/koperator/blob/71000a41df4fe3c1565217df9ff75ae9b9adbec6/pkg/resources/kafka/headlessService.go#L151

"services" already filtered by label selector so it is filteredSvcsToDelete

panyuenlau: > "services" already filtered by label selector so it is equal filteredSvcsToDelete

I guess I have a slightly different interpretation - we use the label selectors to query all of the services that match the labels (and they are stored in the `services` variable), then we filter them based on the external listener configuration, which results in `filteredSvcsToDelete` and these are the services that we will actually remove at the end

pregnor (approved):
LGTM at a glance

Kuvesz (approved):

PR #934, opened 2023-02-17
Allow creating KafkaTopic CR for topic that already present on the Kafka cluster

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no |
| New feature?    | yes |
| API breaks?     | no |
| Deprecations?   | no |

### What's in this PR?
It allows to create a KafkaTopic CR when the referenced topic is already present on the Kafka cluster. 
It allows when KafkaTopic CR:
-  has managedBy: koperator annotation 
-  configuration is the same as the present kafka topic configuration
otherwise the KafkaTopic CR will be rejected by validation webhook.
There is a new status field "managedBy" which value is the manager of the kafka topic

When the KafkaTopic CR has "managedBy" annotation with not "koperator" value and the KafkaTopic CR is deleted or modified then the present kafka topic will not be affected in the kafka cluster. 

### Why?

Let the user create KafkaTopic resources for that topic which is already on the Kafka cluster
Users can modify the topic configuration and partition number for that topic through the created KafkaTopic CR

### Checklist

- [x] Implementation tested

### More details
When the KafkaTopic rejected the above error message is an example:
```code
The KafkaTopic "ccmetrics" is invalid:
* spec.name: Invalid value: "__CruiseControlMetrics": topic "__CruiseControlMetrics" already exists on kafka cluster and it is not managed by Koperator,
					when you want to be managed by Koperator to be able to modify its configuration through KafkaTopic CR,
					add this "managedBy: koperator" annotation for this KafkaTopic CR
* spec.partitions: Invalid value: 3: initial KafkaTopic partition number must be the same as the already exist kafka topic has  (given: 3 present: 12)
* spec.replicationfactor: Invalid value: 3: initial KafkaTopic replication factor must be the same as the already exist kafka topic has (given: 3 present: 2)
* spec.config: Invalid value: 3: initial KafkaTopic configuration must be the same as the already exist kafka topic configuration (difference:   map[string]*string{
  	"cleanup.policy":      &"delete",
- 	"min.insync.replicas": &"1",
- 	"retention.ms":        &"18000000",
+ 	"retention.ms":        &"604800000",
  }
```

### DIFF
--- a/charts/kafka-operator/templates/crds.yaml
+++ b/charts/kafka-operator/templates/crds.yaml
@@ -20101,10 +20101,10 @@ spec:
             properties:
               managedBy:
                 description: 'ManagedBy describes who is the manager of the Kafka
-                  topic. When its value is not ''koperator'' then modifications to
-                  the KafkaTopic CR''s topic configurations will be unaffected on
+                  topic. When its value is not "koperator" then modifications to the
+                  topic configurations of the KafkaTopic CR will be unaffected on
                   the Kafka topic. Manager of the Kafka topic can be changed by adding
-                  the ''managedBy: <manager>'' annotation to the KafkaTopic CR.'
+                  the "managedBy: <manager>" annotation to the KafkaTopic CR.'
                 type: string
               state:
                 description: TopicState defines the state of a KafkaTopic
--- a/charts/kafka-operator/templates/crds.yaml
+++ b/charts/kafka-operator/templates/crds.yaml
@@ -3907,6 +3907,21 @@ spec:
                         external listener advertise address according to the description
                         of the "hostnameOverride" field.
                       type: object
+                    nodePortNodeAddressType:
+                      description: When "hostNameOverride" and brokerConfig.nodePortExternalIP
+                        are empty and NodePort access method is selected for an external
+                        listener the NodePortNodeAdddressType defines the Kafka broker's
+                        Kubernetes node's address type that shall be used in the advertised.listeners
+                        property. https://kubernetes.io/docs/concepts/architecture/nodes/#addresses
+                        The NodePortNodeAddressType's possible values can be Hostname,
+                        ExternalIP, InternalIP, InternalDNS,ExternalDNS
+                      enum:
+                      - Hostname
+                      - ExternalIP
+                      - InternalIP
+                      - InternalDNS
+                      - ExternalDNS
+                      type: string
                     nodeSelector:
                       additionalProperties:
                         type: string
@@ -10034,6 +10049,21 @@ spec:
                             the broker's external listener advertise address according
                             to the description of the "hostnameOverride" field.
                           type: object
+                        nodePortNodeAddressType:
+                          description: When "hostNameOverride" and brokerConfig.nodePortExternalIP
+                            are empty and NodePort access method is selected for an
+                            external listener the NodePortNodeAdddressType defines
+                            the Kafka broker's Kubernetes node's address type that
+                            shall be used in the advertised.listeners property. https://kubernetes.io/docs/concepts/architecture/nodes/#addresses
+                            The NodePortNodeAddressType's possible values can be Hostname,
+                            ExternalIP, InternalIP, InternalDNS,ExternalDNS
+                          enum:
+                          - Hostname
+                          - ExternalIP
+                          - InternalIP
+                          - InternalDNS
+                          - ExternalDNS
+                          type: string
                         nodeSelector:
                           additionalProperties:
                             type: string
@@ -20102,9 +20132,9 @@ spec:
               managedBy:
                 description: 'ManagedBy describes who is the manager of the Kafka
                   topic. When its value is not "koperator" then modifications to the
-                  topic configurations of the KafkaTopic CR will be unaffected on
-                  the Kafka topic. Manager of the Kafka topic can be changed by adding
-                  the "managedBy: <manager>" annotation to the KafkaTopic CR.'
+                  topic configurations of the KafkaTopic CR will not be propagated
+                  to the Kafka topic. Manager of the Kafka topic can be changed by
+                  adding the "managedBy: <manager>" annotation to the KafkaTopic CR.'
                 type: string
               state:
                 description: TopicState defines the state of a KafkaTopic
--- a/charts/kafka-operator/templates/crds.yaml
+++ b/charts/kafka-operator/templates/crds.yaml
@@ -20100,6 +20100,11 @@ spec:
             description: KafkaTopicStatus defines the observed state of KafkaTopic
             properties:
               managedBy:
+                description: 'ManagedBy describes who is the manager of the Kafka
+                  topic. When its value is not ''koperator'' then modifications to
+                  the KafkaTopic CR''s topic configurations will be unaffected on
+                  the Kafka topic. Manager of the Kafka topic can be changed by adding
+                  the ''managedBy: <manager>'' annotation to the KafkaTopic CR.'
                 type: string
               state:
                 description: TopicState defines the state of a KafkaTopic
--- a/charts/kafka-operator/templates/crds.yaml
+++ b/charts/kafka-operator/templates/crds.yaml
@@ -20099,10 +20099,13 @@ spec:
           status:
             description: KafkaTopicStatus defines the observed state of KafkaTopic
             properties:
+              managedBy:
+                type: string
               state:
                 description: TopicState defines the state of a KafkaTopic
                 type: string
             required:
+            - managedBy
             - state
             type: object
         type: object
--- a/controllers/kafkatopic_controller.go
+++ b/controllers/kafkatopic_controller.go
@@ -17,6 +17,7 @@ package controllers
 import (
 	"context"
 	"errors"
+	"fmt"
 	"reflect"
 	"strings"
 
@@ -40,7 +41,7 @@ import (
 var topicFinalizer = "finalizer.kafkatopics.kafka.banzaicloud.io"
 
 func isTopicManagedByKoperator(topic metav1.Object) bool {
-	if manager, ok := topic.GetAnnotations()[webhooks.TopicManagedByKoperatorAnnotationKey]; !ok || strings.ToLower(manager) == webhooks.TopicManagedByKoperatorAnnotationValue {
+	if manager, ok := topic.GetAnnotations()[webhooks.TopicManagedByAnnotationKey]; !ok || strings.ToLower(manager) == webhooks.TopicManagedByKoperatorAnnotationValue {
 		return true
 	}
 	return false
@@ -105,6 +106,19 @@ func (r *KafkaTopicReconciler) Reconcile(ctx context.Context, request reconcile.
 		return requeueWithError(reqLogger, "failed to lookup referenced cluster", err)
 	}
 
+	// Set managed status based on KafkaTopic managedBy annotation
+	managedByStatus := webhooks.TopicManagedByKoperatorAnnotationValue
+	if !isTopicManagedByKoperator(instance) {
+		managedByStatus = instance.GetAnnotations()[webhooks.TopicManagedByAnnotationKey]
+	}
+
+	if instance.Status.ManagedBy != managedByStatus {
+		instance.Status.ManagedBy = managedByStatus
+		if err := r.Client.Status().Update(ctx, instance); err != nil {
+			return requeueWithError(reqLogger, "failed to update kafkatopic status", err)
+		}
+	}
+
 	// Get a kafka connection
 	broker, close, err := newKafkaFromCluster(r.Client, cluster)
 	if err != nil {
@@ -119,12 +133,7 @@ func (r *KafkaTopicReconciler) Reconcile(ctx context.Context, request reconcile.
 
 	// No need to do anything when the kafka topic is not managed by Koperator
 	if !isTopicManagedByKoperator(instance) {
-		// if instance.Status.State != v1alpha1.TopicStateUnmanaged {
-		// 	instance.Status = v1alpha1.KafkaTopicStatus{State: v1alpha1.TopicStateUnmanaged}
-		// 	if err := r.Client.Status().Update(ctx, instance); err != nil {
-		// 		return requeueWithError(reqLogger, "failed to update kafkatopic status", err)
-		// 	}
-		// }
+		reqLogger.Info(fmt.Sprintf("topic '%s' is not managed by %s it is managed by '%s' ==> reconciled", instance.Spec.Name, webhooks.TopicManagedByKoperatorAnnotationValue, instance.GetAnnotations()[webhooks.TopicManagedByAnnotationKey]))
 		return reconciled()
 	}
 
--- a/controllers/kafkatopic_controller.go
+++ b/controllers/kafkatopic_controller.go
@@ -18,9 +18,11 @@ import (
 	"context"
 	"errors"
 	"reflect"
+	"strings"
 
 	"github.com/go-logr/logr"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -36,6 +38,13 @@ import (
 )
 
 var topicFinalizer = "finalizer.kafkatopics.kafka.banzaicloud.io"
+
+func isTopicManagedByKoperator(topic metav1.Object) bool {
+	if manager, ok := topic.GetAnnotations()[webhooks.TopicManagedByKoperatorAnnotationKey]; !ok || strings.ToLower(manager) == webhooks.TopicManagedByKoperatorAnnotationValue {
+		return true
+	}
+	return false
+}
 
 // SetupKafkaTopicWithManager registers kafka topic controller with manager
 func SetupKafkaTopicWithManager(mgr ctrl.Manager, maxConcurrentReconciles int) *ctrl.Builder {
@@ -106,6 +115,17 @@ func (r *KafkaTopicReconciler) Reconcile(ctx context.Context, request reconcile.
 	// Check if marked for deletion and if so run finalizers
 	if k8sutil.IsMarkedForDeletion(instance.ObjectMeta) {
 		return r.checkFinalizers(ctx, broker, instance)
+	}
+
+	// No need to do anything when the kafka topic is not managed by Koperator
+	if !isTopicManagedByKoperator(instance) {
+		// if instance.Status.State != v1alpha1.TopicStateUnmanaged {
+		// 	instance.Status = v1alpha1.KafkaTopicStatus{State: v1alpha1.TopicStateUnmanaged}
+		// 	if err := r.Client.Status().Update(ctx, instance); err != nil {
+		// 		return requeueWithError(reqLogger, "failed to update kafkatopic status", err)
+		// 	}
+		// }
+		return reconciled()
 	}
 
 	// Check if the topic already exists
@@ -195,8 +215,11 @@ func (r *KafkaTopicReconciler) checkFinalizers(ctx context.Context, broker kafka
 	reqLogger.Info("Kafka topic is marked for deletion")
 	var err error
 	if util.StringSliceContains(topic.GetFinalizers(), topicFinalizer) {
-		if err = r.finalizeKafkaTopic(reqLogger, broker, topic); err != nil {
-			return requeueWithError(reqLogger, "failed to finalize kafkatopic", err)
+		// Remove topic from Kafka cluster when it is managed by Koperator
+		if isTopicManagedByKoperator(topic) {
+			if err = r.finalizeKafkaTopic(reqLogger, broker, topic); err != nil {
+				return requeueWithError(reqLogger, "failed to finalize kafkatopic", err)
+			}
 		}
 		if err = r.removeFinalizer(ctx, topic); err != nil {
 			return requeueWithError(reqLogger, "failed to remove finalizer from kafkatopic", err)
@@ -217,10 +240,6 @@ func (r *KafkaTopicReconciler) finalizeKafkaTopic(reqLogger logr.Logger, broker 
 		return err
 	}
 	if exists != nil {
-		// When the topic is not managed by the Koperator, it is not our responsibility to delete it.
-		if val, ok := topic.GetAnnotations()[webhooks.ManagedByAnnotationKey]; ok && val != webhooks.ManagedByAnnotationValue {
-			return nil
-		}
 		// DeleteTopic with wait to make sure it goes down fully in case of cluster
 		// deletion.
 		// TODO (tinyzimmer): Perhaps this should only wait when it's the cluster
--- a/controllers/kafkatopic_controller.go
+++ b/controllers/kafkatopic_controller.go
@@ -41,8 +41,8 @@ import (
 var topicFinalizer = "finalizer.kafkatopics.kafka.banzaicloud.io"
 
 func isTopicManagedByKoperator(topic metav1.Object) bool {
-	if manager, ok := topic.GetAnnotations()[webhooks.TopicManagedByAnnotationKey]; !ok || strings.ToLower(manager) == webhooks.TopicManagedByKoperatorAnnotationValue {
-		return true
+	if managedByAnnotation, ok := topic.GetAnnotations()[webhooks.TopicManagedByAnnotationKey]; !ok {
+		return strings.ToLower(managedByAnnotation) == webhooks.TopicManagedByKoperatorAnnotationValue
 	}
 	return false
 }
@@ -133,7 +133,7 @@ func (r *KafkaTopicReconciler) Reconcile(ctx context.Context, request reconcile.
 
 	// No need to do anything when the kafka topic is not managed by Koperator
 	if !isTopicManagedByKoperator(instance) {
-		reqLogger.Info(fmt.Sprintf("topic '%s' is not managed by %s it is managed by '%s' ==> reconciled", instance.Spec.Name, webhooks.TopicManagedByKoperatorAnnotationValue, instance.GetAnnotations()[webhooks.TopicManagedByAnnotationKey]))
+		reqLogger.Info(fmt.Sprintf("topic '%s' is not managed by %s it is managed by '%s' ==> nothing to reconcile here", instance.Spec.Name, webhooks.TopicManagedByKoperatorAnnotationValue, instance.GetAnnotations()[webhooks.TopicManagedByAnnotationKey]))
 		return reconciled()
 	}
 
--- a/controllers/kafkatopic_controller.go
+++ b/controllers/kafkatopic_controller.go
@@ -32,6 +32,7 @@ import (
 	"github.com/banzaicloud/koperator/pkg/k8sutil"
 	"github.com/banzaicloud/koperator/pkg/kafkaclient"
 	"github.com/banzaicloud/koperator/pkg/util"
+	"github.com/banzaicloud/koperator/pkg/webhooks"
 )
 
 var topicFinalizer = "finalizer.kafkatopics.kafka.banzaicloud.io"
@@ -216,6 +217,10 @@ func (r *KafkaTopicReconciler) finalizeKafkaTopic(reqLogger logr.Logger, broker 
 		return err
 	}
 	if exists != nil {
+		// When the topic is not managed by the Koperator, it is not our responsibility to delete it.
+		if val, ok := topic.GetAnnotations()[webhooks.ManagedByAnnotationKey]; ok && val != webhooks.ManagedByAnnotationValue {
+			return nil
+		}
 		// DeleteTopic with wait to make sure it goes down fully in case of cluster
 		// deletion.
 		// TODO (tinyzimmer): Perhaps this should only wait when it's the cluster
--- a/controllers/kafkatopic_controller.go
+++ b/controllers/kafkatopic_controller.go
@@ -41,10 +41,10 @@ import (
 var topicFinalizer = "finalizer.kafkatopics.kafka.banzaicloud.io"
 
 func isTopicManagedByKoperator(topic metav1.Object) bool {
-	if managedByAnnotation, ok := topic.GetAnnotations()[webhooks.TopicManagedByAnnotationKey]; !ok {
+	if managedByAnnotation, hasManagedByAnnotation := topic.GetAnnotations()[webhooks.TopicManagedByAnnotationKey]; hasManagedByAnnotation {
 		return strings.ToLower(managedByAnnotation) == webhooks.TopicManagedByKoperatorAnnotationValue
 	}
-	return false
+	return true
 }
 
 // SetupKafkaTopicWithManager registers kafka topic controller with manager
@@ -133,7 +133,7 @@ func (r *KafkaTopicReconciler) Reconcile(ctx context.Context, request reconcile.
 
 	// No need to do anything when the kafka topic is not managed by Koperator
 	if !isTopicManagedByKoperator(instance) {
-		reqLogger.Info(fmt.Sprintf("topic '%s' is not managed by %s it is managed by '%s' ==> nothing to reconcile here", instance.Spec.Name, webhooks.TopicManagedByKoperatorAnnotationValue, instance.GetAnnotations()[webhooks.TopicManagedByAnnotationKey]))
+		reqLogger.Info(fmt.Sprintf("topic '%s' is not managed by %s it is managed by '%s' ==> nothing to reconcile here", instance.Spec.Name, webhooks.TopicManagedByKoperatorAnnotationValue, managedByStatus))
 		return reconciled()
 	}
 
--- a/pkg/webhooks/kafkatopic_validator.go
+++ b/pkg/webhooks/kafkatopic_validator.go
@@ -173,22 +173,22 @@ func (s *KafkaTopicValidator) checkKafka(ctx context.Context, topic *banzaicloud
 				if manager, ok := topic.GetAnnotations()[TopicManagedByAnnotationKey]; !ok || strings.ToLower(manager) != TopicManagedByKoperatorAnnotationValue {
 					allErrs = append(allErrs, field.Invalid(field.NewPath("spec").Child("name"), topic.Spec.Name,
 						fmt.Sprintf(`topic "%s" already exists on kafka cluster and it is not managed by Koperator,
-					if you want it to be managed by Koperator making you able to modify its configuration through a KafkaTopic CR,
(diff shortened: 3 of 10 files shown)


api/v1alpha1/kafkatopic_types.go
```diff
@@ -40,7 +40,8 @@ type KafkaTopicSpec struct {
 // KafkaTopicStatus defines the observed state of KafkaTopic
 // +k8s:openapi-gen=true
 type KafkaTopicStatus struct {
-	State TopicState `json:"state"`
+	State     TopicState `json:"state"`
+	ManagedBy string     `json:"managedBy"`
```
pregnor, line 44: **Question**: What is the reason we are making the `managedBy` part of the status in addition to using it as an annotation? (Looks like double booking to me, but I'm mostly just curious, not debating it.)

controllers/kafkatopic_controller.go
```diff
@@ -32,10 +35,18 @@ import (
 	"github.com/banzaicloud/koperator/pkg/k8sutil"
 	"github.com/banzaicloud/koperator/pkg/kafkaclient"
 	"github.com/banzaicloud/koperator/pkg/util"
+	"github.com/banzaicloud/koperator/pkg/webhooks"
 )
 
 var topicFinalizer = "finalizer.kafkatopics.kafka.banzaicloud.io"
 
+func isTopicManagedByKoperator(topic metav1.Object) bool {
+	if manager, ok := topic.GetAnnotations()[webhooks.TopicManagedByAnnotationKey]; !ok || strings.ToLower(manager) == webhooks.TopicManagedByKoperatorAnnotationValue {
```
pregnor, line 44: **Recommended**: We **SHOULD NOT** return true for `!ok` case IMO, because in case there is no `managedBy` annotation on the object we should not assume Koperator created it. To my understanding what you wrote in the description (that the user needs to set the annotation if they want it to be managed) is a good way to handle this. Or am I misunderstanding something here?

```suggestion
	if manager, ok := topic.GetAnnotations()[webhooks.TopicManagedByAnnotationKey]; ok && strings.ToLower(manager) == webhooks.TopicManagedByKoperatorAnnotationValue {
```

pregnor, line 44: **Recommended**: We **SHOULD NOT** call the annotation `manager` IMO because it is misleading.

```suggestion
	if managedByAnnotation, ok := topic.GetAnnotations()[webhooks.TopicManagedByAnnotationKey]; !ok || strings.ToLower(managedByAnnotation) == webhooks.TopicManagedByKoperatorAnnotationValue {
```

controllers/kafkatopic_controller.go
```diff
@@ -32,10 +35,18 @@ import (
...
+		return true
```
pregnor, line 45: **Optional**: We **MAY** return the second condition (managed by annotation comparison) directly here without extra conditional to make it more readable.
(Can't give suggestion due to the condition recommendation I made which needs to be cleared up first.)

controllers/kafkatopic_controller.go
```diff
@@ -95,6 +106,19 @@ func (r *KafkaTopicReconciler) Reconcile(ctx context.Context, request reconcile.
 		return requeueWithError(reqLogger, "failed to lookup referenced cluster", err)
 	}
 
+	// Set managed status based on KafkaTopic managedBy annotation
+	managedByStatus := webhooks.TopicManagedByKoperatorAnnotationValue
+	if !isTopicManagedByKoperator(instance) {
+		managedByStatus = instance.GetAnnotations()[webhooks.TopicManagedByAnnotationKey]
+	}
```
pregnor, line 113: **Note**: here the logic depends on the `isTopicManagedByKoperator` function as well.

controllers/kafkatopic_controller.go
```diff
@@ -95,6 +106,19 @@ func (r *KafkaTopicReconciler) Reconcile(ctx context.Context, request reconcile.
...
+
+	if instance.Status.ManagedBy != managedByStatus {
```
pregnor, line 115: **Note**: for the CI errors you might need to release a new version of the API first and pull it into the root module, this is a recent change.

You can check with local replacement locally on your machine, also you can use a temporary dev tag or a real tag.

controllers/kafkatopic_controller.go
```diff
@@ -107,6 +131,12 @@ func (r *KafkaTopicReconciler) Reconcile(ctx context.Context, request reconcile.
 		return r.checkFinalizers(ctx, broker, instance)
 	}
 
+	// No need to do anything when the kafka topic is not managed by Koperator
+	if !isTopicManagedByKoperator(instance) {
+		reqLogger.Info(fmt.Sprintf("topic '%s' is not managed by %s it is managed by '%s' ==> reconciled", instance.Spec.Name, webhooks.TopicManagedByKoperatorAnnotationValue, instance.GetAnnotations()[webhooks.TopicManagedByAnnotationKey]))
```
pregnor, line 136: **Optional**: We **MAY** add a bit more clarity

```suggestion
		reqLogger.Info(fmt.Sprintf("topic '%s' is not managed by %s it is managed by '%s' ==> nothing to reconcile here", instance.Spec.Name, webhooks.TopicManagedByKoperatorAnnotationValue, instance.GetAnnotations()[webhooks.TopicManagedByAnnotationKey]))
```

pkg/webhooks/kafkatopic_validator.go
```diff
@@ -55,22 +63,22 @@ func (s KafkaTopicValidator) ValidateDelete(ctx context.Context, obj runtime.Obj
 
 func (s *KafkaTopicValidator) validate(ctx context.Context, obj runtime.Object) error {
 	kafkaTopic := obj.(*banzaicloudv1alpha1.KafkaTopic)
-	log := s.Log.WithValues("name", kafkaTopic.GetName(), "namespace", kafkaTopic.GetNamespace())
-	fieldErrs, err := s.validateKafkaTopic(ctx, kafkaTopic, log)
+	s.Log = s.Log.WithValues("name", kafkaTopic.GetName(), "namespace", kafkaTopic.GetNamespace())
```
pregnor, line 66: **Question**: Are we sure we want to overwrite the stored logger here? For different topics this would be different and in that case the local instance would be better IMO.

pkg/webhooks/kafkatopic_validator.go
```diff
@@ -154,37 +161,61 @@ func (s *KafkaTopicValidator) checkKafka(ctx context.Context, topic *banzaicloud
...
 	if existing != nil {
 		// Check if this is the correct CR for this topic
 		topicCR := &banzaicloudv1alpha1.KafkaTopic{}
 		if err := s.Client.Get(ctx, types.NamespacedName{Name: topic.Name, Namespace: topic.Namespace}, topicCR); err != nil {
+			// Checking that the validation request is update
 			if apierrors.IsNotFound(err) {
-				// User is trying to overwrite an existing topic - bad user
-				logMsg := fmt.Sprintf("topic already exists on kafka cluster '%s'", topic.Spec.ClusterRef.Name)
-				return field.Invalid(field.NewPath("spec").Child("name"), topic.Spec.Name, logMsg), nil
+				if manager, ok := topic.GetAnnotations()[TopicManagedByAnnotationKey]; !ok || strings.ToLower(manager) != TopicManagedByKoperatorAnnotationValue {
```
pregnor, line 172: **Recommended**: We **SHOULD** possibly move the `isTopicManagedByKoperator` here and then we could reuse it here for branching so we wouldn't need to duplicate the condition.

pkg/webhooks/kafkatopic_validator.go
```diff
@@ -154,37 +161,61 @@ func (s *KafkaTopicValidator) checkKafka(ctx context.Context, topic *banzaicloud
...
+				}
+				if existing.ReplicationFactor != int16(topic.Spec.ReplicationFactor) {
+					allErrs = append(allErrs, field.Invalid(field.NewPath("spec").Child("replicationfactor"), topic.Spec.ReplicationFactor,
+						fmt.Sprintf(`initial KafkaTopic replication factor must be the same as the already exist kafka topic has (given: %v present: %v)`, topic.Spec.ReplicationFactor, existing.ReplicationFactor)))
+				}
+
+				if diff := cmp.Diff(existing.ConfigEntries, util.MapStringStringPointer(topic.Spec.Config), cmpopts.EquateEmpty()); diff != "" {
+					allErrs = append(allErrs, field.Invalid(field.NewPath("spec").Child("config"), topic.Spec.Partitions,
+						fmt.Sprintf(`initial KafkaTopic configuration must be the same as the already exist kafka topic configuration difference: %s`, diff)))
+				}
```
pregnor, line 191: **Note**: I have been thinking how we could make this less fragile on newly added fields and also on structure, but I couldn't come up with a solution that keeps the information of the fields.

pkg/webhooks/kafkatopic_validator_test.go
```diff
@@ -168,21 +303,11 @@ func TestValidateTopic(t *testing.T) {
...
 	topic.Name = "test-topic"
-	fieldErrorList, err = kafkaTopicValidator.validateKafkaTopic(context.Background(), topic, logr.Discard())
-	if err != nil {
-		t.Errorf("err should be nil, got: %s", err)
-	}
-	if len(fieldErrorList) != 1 {
-		t.Error("Expected not allowed due to existing topic with same name, got allowed")
-	} else if !strings.Contains(fieldErrorList.ToAggregate().Error(), "topic already exists on kafka cluster") {
-		t.Error("Expected not allowed for reason: already exists")
-	}
```
pregnor, line 185: **Question**: Why are we removing this section?

api/v1alpha1/kafkatopic_types.go
bartam1: When there is no annotation the referenced kafka topic is managed by Koperator.
I think it is good to show who is the manager also in the topic status.
This can be a confirmation for the user when he checks the topic status that is managed by who.

controllers/kafkatopic_controller.go
bartam1: This is a little bit different from what we need in the kafkatopic_validation. This means that isTopicManagedByKoperator
thus when there is no annotation then it is created by koperator (because when there is existing topic and we just create kafkatopic cr we dont have to put annotation), and when it has annotation managedBy:koperator it is also managed by koperator so we need to apply changes based on the kafkatopic CR settings.

controllers/kafkatopic_controller.go
bartam1: Yes, thanks. First I want to get approve for this PR and later I will make the necessary changes on API and tagging.

pkg/webhooks/kafkatopic_validator.go
bartam1: I think it is OK in this way because the kafkatopic controller runs sequential and not parallel

pkg/webhooks/kafkatopic_validator.go
bartam1: This has to be different what is in the isTopicManagedByKoperator function.
In this case the topic already exist so when there is no managedBy annotation or the managedBy annotation is not Koperator then we reject this kafkatopic because the kafka topic is exists and need confirmation.

In the kafkatopic controller: when there is no managedBy annotation or the annotation is koperator then the manager is koperator.

pkg/webhooks/kafkatopic_validator_test.go
bartam1: I felt like it as a test duplicate because I wrote a better test for this here:  https://github.com/banzaicloud/koperator/blob/3444e5d819a4b4a9ac653307e498c6daf80bcf0c/pkg/webhooks/kafkatopic_validator_test.go#L75

api/v1alpha1/kafkatopic_types.go
Kuvesz: Since we're handling this in annotations everywhere else wouldn't it make more sense to add an annotation instead?

pkg/webhooks/kafkatopic_validator.go
```diff
@@ -154,37 +161,61 @@ func (s *KafkaTopicValidator) checkKafka(ctx context.Context, topic *banzaicloud
 		return nil, errors.WrapIff(err, fmt.Sprintf("failed to list topics for kafka cluster: %s", topic.Spec.ClusterRef.Name))
 	}
 
+	var allErrs field.ErrorList
 	// The topic exists
 	if existing != nil {
 		// Check if this is the correct CR for this topic
 		topicCR := &banzaicloudv1alpha1.KafkaTopic{}
 		if err := s.Client.Get(ctx, types.NamespacedName{Name: topic.Name, Namespace: topic.Namespace}, topicCR); err != nil {
+			// Checking that the validation request is update
```
Kuvesz, line 170:
```suggestion
			// Checking that the validation request is updated
```

pkg/webhooks/kafkatopic_validator.go
```diff
@@ -154,37 +161,61 @@ func (s *KafkaTopicValidator) checkKafka(ctx context.Context, topic *banzaicloud
...
 			if apierrors.IsNotFound(err) {
-				// User is trying to overwrite an existing topic - bad user
-				logMsg := fmt.Sprintf("topic already exists on kafka cluster '%s'", topic.Spec.ClusterRef.Name)
-				return field.Invalid(field.NewPath("spec").Child("name"), topic.Spec.Name, logMsg), nil
+				if manager, ok := topic.GetAnnotations()[TopicManagedByAnnotationKey]; !ok || strings.ToLower(manager) != TopicManagedByKoperatorAnnotationValue {
+					allErrs = append(allErrs, field.Invalid(field.NewPath("spec").Child("name"), topic.Spec.Name,
+						fmt.Sprintf(`topic "%s" already exists on kafka cluster and it is not managed by Koperator,
+					when you want to be managed by Koperator to be able to modify its configuration through KafkaTopic CR,
+					add this "%s: %s" annotation for this KafkaTopic CR`, topic.Spec.Name, TopicManagedByAnnotationKey, TopicManagedByKoperatorAnnotationValue)))
```
Kuvesz, line 176:
```suggestion
						fmt.Sprintf(`topic "%s" already exists on kafka cluster and it is not managed by Koperator,
					if you want it to be managed by Koperator making you able to modify its configuration through a KafkaTopic CR,
					add this "%s: %s" annotation to this KafkaTopic CR`, topic.Spec.Name, TopicManagedByAnnotationKey, TopicManagedByKoperatorAnnotationValue)))
```

pkg/webhooks/kafkatopic_validator.go
```diff
@@ -154,37 +161,61 @@ func (s *KafkaTopicValidator) checkKafka(ctx context.Context, topic *banzaicloud
...
+				}
+				// Comparing KafkaTopic configuration with the Kafka topic
```
Kuvesz, line 178:
```suggestion
				// Comparing KafkaTopic configuration with the existing Kafka topic
```
Did you mean this?

pkg/webhooks/kafkatopic_validator.go
```diff
@@ -154,37 +161,61 @@ func (s *KafkaTopicValidator) checkKafka(ctx context.Context, topic *banzaicloud
...
+				if existing.NumPartitions != topic.Spec.Partitions {
+					allErrs = append(allErrs, field.Invalid(field.NewPath("spec").Child("partitions"), topic.Spec.Partitions,
+						fmt.Sprintf(`initial KafkaTopic partition number must be the same as the already exist kafka topic has  (given: %v present: %v)`, topic.Spec.Partitions, existing.NumPartitions)))
```
Kuvesz, line 181:
```suggestion
						fmt.Sprintf(`initial KafkaTopic partition number must be the same as what the existing kafka topic has  (given: %v present: %v)`, topic.Spec.Partitions, existing.NumPartitions)))
```

pkg/webhooks/kafkatopic_validator.go
```diff
@@ -154,37 +161,61 @@ func (s *KafkaTopicValidator) checkKafka(ctx context.Context, topic *banzaicloud
...
+				}
+				if existing.ReplicationFactor != int16(topic.Spec.ReplicationFactor) {
+					allErrs = append(allErrs, field.Invalid(field.NewPath("spec").Child("replicationfactor"), topic.Spec.ReplicationFactor,
+						fmt.Sprintf(`initial KafkaTopic replication factor must be the same as the already exist kafka topic has (given: %v present: %v)`, topic.Spec.ReplicationFactor, existing.ReplicationFactor)))
```
Kuvesz, line 185:
```suggestion
						fmt.Sprintf(`initial KafkaTopic replication factor must be the same as what the existing kafka topic has (given: %v present: %v)`, topic.Spec.ReplicationFactor, existing.ReplicationFactor)))
```

pkg/webhooks/kafkatopic_validator.go
```diff
@@ -154,37 +161,61 @@ func (s *KafkaTopicValidator) checkKafka(ctx context.Context, topic *banzaicloud
...
+				}
+
+				if diff := cmp.Diff(existing.ConfigEntries, util.MapStringStringPointer(topic.Spec.Config), cmpopts.EquateEmpty()); diff != "" {
+					allErrs = append(allErrs, field.Invalid(field.NewPath("spec").Child("config"), topic.Spec.Partitions,
+						fmt.Sprintf(`initial KafkaTopic configuration must be the same as the already exist kafka topic configuration difference: %s`, diff)))
```
Kuvesz, line 190:
```suggestion
						fmt.Sprintf(`initial KafkaTopic configuration must be the same as the already existing kafka topic configuration. Difference: %s`, diff)))
```

pkg/webhooks/kafkatopic_validator_test.go
```diff
@@ -191,10 +316,11 @@ func TestValidateTopic(t *testing.T) {
 
 	// partition decrease attempt
 	topic.Spec.Partitions = 1
-	fieldErrorList, err = kafkaTopicValidator.validateKafkaTopic(context.Background(), topic, logr.Discard())
+	fieldErrorList, err = kafkaTopicValidator.validateKafkaTopic(context.Background(), topic)
 	if err != nil {
 		t.Errorf("err should be nil, got: %s", err)
 	}
+	// TODO BUG
```
Kuvesz, line 323: What do you mean here?

Kuvesz (commented):
I left a couple of suggestions and questions.

pkg/webhooks/kafkatopic_validator.go
bartam1: Since we're handling this in annotations everywhere else wouldn't it make more sense to add an annotation instead?
- We can add annotations for every newly created KafkaTopic with the kafkaTopic controller and for the existing KafkaTopic CR-s when there will be a reconcilation and there is no managedBy annotation. What if there is a missing managedBy annotation we every time force write the managedBy: koperator annotation?
What is the problem with to show that this topic certainly not managed by koperator as a confirmation in the status also. I would feel myself more relaxed when I delete a KafkaTopic Cr and I dont want to delete the topic itself when I can see in the status it is not managed by koperator.  This is only opinion of course.
Basicly the user set annotation to change the manager of the topic from koperator to something else, and he can get feedback from the status that this change has happened.

pkg/webhooks/kafkatopic_validator_test.go
bartam1: I left it accidentally. Thanks!

pkg/webhooks/kafkatopic_validator.go
bartam1: It gives better user experience for me.

controllers/kafkatopic_controller.go
pregnor: I left this comment to have a placeholder in case we want to modify the `!ok` handling, so it would be easier to double check the cases where it is used.

pkg/webhooks/kafkatopic_validator.go
pregnor: I can let this fly, but the stored validator logger will still point to some topic after the validation already run which is IMO misleading, because it handles multiple topics.
(Practically the result is the same, but conceptually we are abusing implementation details instead of doing this cleanly.)

pkg/webhooks/kafkatopic_validator.go
pregnor: Yeah, it makes sense with the `!ok` clearup, thanks.

pkg/webhooks/kafkatopic_validator.go
bartam1: Agree. Thank you for bringing this up. I fix this.

pkg/webhooks/kafkatopic_validator.go
```diff
@@ -172,22 +173,23 @@ func (s *KafkaTopicValidator) checkKafka(ctx context.Context, topic *banzaicloud
...
 				if existing.ReplicationFactor != int16(topic.Spec.ReplicationFactor) {
 					allErrs = append(allErrs, field.Invalid(field.NewPath("spec").Child("replicationfactor"), topic.Spec.ReplicationFactor,
-						fmt.Sprintf(`initial KafkaTopic replication factor must be the same as the already exist kafka topic has (given: %v present: %v)`, topic.Spec.ReplicationFactor, existing.ReplicationFactor)))
+						fmt.Sprintf(`initial KafkaTopic replication factor must be the same as what the existing kafka topic has (given: %v present: %v)`, topic.Spec.ReplicationFactor, existing.ReplicationFactor)))
 				}
 
 				if diff := cmp.Diff(existing.ConfigEntries, util.MapStringStringPointer(topic.Spec.Config), cmpopts.EquateEmpty()); diff != "" {
 					allErrs = append(allErrs, field.Invalid(field.NewPath("spec").Child("config"), topic.Spec.Partitions,
-						fmt.Sprintf(`initial KafkaTopic configuration must be the same as the already exist kafka topic configuration difference: %s`, diff)))
+						fmt.Sprintf(`initial KafkaTopic configuration must be the same as what the existing kafka topic configuration.
```
Kuvesz, line 191:
```suggestion
						fmt.Sprintf(`initial KafkaTopic configuration must be the same as the existing kafka topic configuration.
```

api/v1alpha1/kafkatopic_types.go
```diff
@@ -41,6 +41,10 @@ type KafkaTopicSpec struct {
 // +k8s:openapi-gen=true
 type KafkaTopicStatus struct {
 	State TopicState `json:"state"`
+	// ManagedBy describes who is the manager of the Kafka topic.
+	// When its value is not 'koperator' then modifications to the KafkaTopic CR's topic configurations will be unaffected on the Kafka topic.
```
Kuvesz, line 45: Sorry, I'm a bit confused by this still. What happens if the annotation is not set at all? Do we assume it's managed by koperator or not? I seem to remember a comment (though I can't find it) that said that previously.

Kuvesz (approved):
I can accept the managedBy status field, though I'm not a huge fan of it. :D Just left a quick question and a suggestion on a typo.

charts/kafka-operator/templates/crds.yaml
```diff
@@ -20099,10 +20099,18 @@ spec:
           status:
             description: KafkaTopicStatus defines the observed state of KafkaTopic
             properties:
+              managedBy:
+                description: 'ManagedBy describes who is the manager of the Kafka
+                  topic. When its value is not ''koperator'' then modifications to
+                  the KafkaTopic CR''s topic configurations will be unaffected on
```
panyuenlau, line 20105: The generated `KafkaTopic CR''s` is weird and looks like kubebuilder replaces every `'` with `''`, not sure if there is a way to make it not do that for us

controllers/kafkatopic_controller.go
```diff
@@ -32,10 +35,18 @@ import (
 	"github.com/banzaicloud/koperator/pkg/k8sutil"
 	"github.com/banzaicloud/koperator/pkg/kafkaclient"
 	"github.com/banzaicloud/koperator/pkg/util"
+	"github.com/banzaicloud/koperator/pkg/webhooks"
 )
 
 var topicFinalizer = "finalizer.kafkatopics.kafka.banzaicloud.io"
 
+func isTopicManagedByKoperator(topic metav1.Object) bool {
+	if managedByAnnotation, ok := topic.GetAnnotations()[webhooks.TopicManagedByAnnotationKey]; !ok {
```
panyuenlau, line 44: I believe it is `ok` instead of `!ok`? I got a bit lost following the current logic: 

- when the "managedBy" annotation doesn't exist, we are returning `"" == webhooks.TopicManagedByKoperatorAnnotationValue`, which is always false
- and when "managedBy" annotation does exist, we are hitting the "return false" statement at the end of this function

It would be great to add a unit test for this little function

controllers/kafkatopic_controller.go
```diff
@@ -107,6 +131,12 @@ func (r *KafkaTopicReconciler) Reconcile(ctx context.Context, request reconcile.
 		return r.checkFinalizers(ctx, broker, instance)
 	}
 
+	// No need to do anything when the kafka topic is not managed by Koperator
+	if !isTopicManagedByKoperator(instance) {
+		reqLogger.Info(fmt.Sprintf("topic '%s' is not managed by %s it is managed by '%s' ==> nothing to reconcile here", instance.Spec.Name, webhooks.TopicManagedByKoperatorAnnotationValue, instance.GetAnnotations()[webhooks.TopicManagedByAnnotationKey]))
```
panyuenlau, line 136:
```suggestion
		reqLogger.Info(fmt.Sprintf("topic '%s' is not managed by %s, it is managed by '%s' ==> nothing to reconcile here", instance.Spec.Name, webhooks.TopicManagedByKoperatorAnnotationValue, managedByStatus))
```

Is there a particular reason to not use the `managedByStatus` here?

pkg/webhooks/kafkatopic_validator.go
```diff
@@ -70,7 +79,7 @@ func (s *KafkaTopicValidator) validate(ctx context.Context, obj runtime.Object)
 		kafkaTopic.Name, fieldErrs)
 }
 
-func (s *KafkaTopicValidator) validateKafkaTopic(ctx context.Context, topic *banzaicloudv1alpha1.KafkaTopic, log logr.Logger) (field.ErrorList, error) {
+func (s *KafkaTopicValidator) validateKafkaTopic(ctx context.Context, log logr.Logger, topic *banzaicloudv1alpha1.KafkaTopic) (field.ErrorList, error) {
```
panyuenlau, line 82: One suggestion: it's probably better to get the logger from the context (follow what we have in the existing codebase) instead of passing it around as an argument, I believe passing the logger through argument is recommended

pkg/webhooks/kafkatopic_validator.go
```diff
@@ -154,37 +162,62 @@ func (s *KafkaTopicValidator) checkKafka(ctx context.Context, topic *banzaicloud
...
 		if err := s.Client.Get(ctx, types.NamespacedName{Name: topic.Name, Namespace: topic.Namespace}, topicCR); err != nil {
+			// Checking that the validation request is update
 			if apierrors.IsNotFound(err) {
-				// User is trying to overwrite an existing topic - bad user
-				logMsg := fmt.Sprintf("topic already exists on kafka cluster '%s'", topic.Spec.ClusterRef.Name)
-				return field.Invalid(field.NewPath("spec").Child("name"), topic.Spec.Name, logMsg), nil
+				if manager, ok := topic.GetAnnotations()[TopicManagedByAnnotationKey]; !ok || strings.ToLower(manager) != TopicManagedByKoperatorAnnotationValue {
+					allErrs = append(allErrs, field.Invalid(field.NewPath("spec").Child("name"), topic.Spec.Name,
+						fmt.Sprintf(`topic "%s" already exists on kafka cluster and it is not managed by Koperator,
+					if you want it to be managed by Koperator making you able to modify its configuration through a KafkaTopic CR,
```
panyuenlau, line 176:
```suggestion
					if you want it to be managed by Koperator so you can modify its configurations through a KafkaTopic CR,
```

pkg/webhooks/kafkatopic_validator.go
```diff
@@ -154,37 +162,62 @@ func (s *KafkaTopicValidator) checkKafka(ctx context.Context, topic *banzaicloud
...
+					add this "%s: %s" annotation to this KafkaTopic CR`, topic.Spec.Name, TopicManagedByAnnotationKey, TopicManagedByKoperatorAnnotationValue)))
+				}
+				// Comparing KafkaTopic configuration with the existing
+				if existing.NumPartitions != topic.Spec.Partitions {
+					allErrs = append(allErrs, field.Invalid(field.NewPath("spec").Child("partitions"), topic.Spec.Partitions,
+						fmt.Sprintf(`initial KafkaTopic partition number must be the same as what the existing kafka topic has  (given: %v present: %v)`, topic.Spec.Partitions, existing.NumPartitions)))
```
panyuenlau, line 182: I am a bit confused about this "initial" wording -- it is leading me to the partition number that the users specified when the kafka topic was initially created. But we are actually comparing the new partition number against the existing value

Same confusion while reading the other "initial" messages in this function

api/v1alpha1/kafkatopic_types.go
bartam1: Yes, when there is no annotation it is managedby koperator

controllers/kafkatopic_controller.go
bartam1: We can use both. Same result. But I think better to use your suggestion. Thanks!

pkg/webhooks/kafkatopic_validator.go
bartam1: I also think that passing the logger as argument is recommended and not to use ctx to get the logger.
Probably when you writing tests you can just use in the argument the log.Discard() to solve the dependency.

charts/kafka-operator/templates/crds.yaml
bartam1: Weird :) KafkaTopic CR''s ...

controllers/kafkatopic_controller.go
bartam1: Nice catch!
Actually when I changed based Pregnor suggestion. It went wrong :) (https://github.com/banzaicloud/koperator/pull/934#discussion_r1112859931)

pkg/webhooks/kafkatopic_validator.go
bartam1: I have changed it.  What you think? Is it clear now?

controllers/kafkatopic_controller.go
bartam1: I added test

pkg/webhooks/kafkatopic_validator.go
pregnor: I agree with Márton's reasoning here, and to me it makes sense to handle the "missing annotation" part with a status field that explicitly tells how it will be handled.

pkg/webhooks/kafkatopic_validator.go
panyuenlau: > I also think that passing the logger as argument is recommended and not to use ctx to get the logger.

Oh really? Can you provide a reference? 

I thought it was the other way around since there was an onboarding task (https://github.com/banzaicloud/koperator/pull/765) that I did a long time ago to replace the logger argument with the one getting from the context. 

If it is the case that it's recommended to pass the logger as argument, we might need to revert that PR

pregnor (approved):
LGTM with 1 comment rewording optional suggestion.

pkg/webhooks/kafkatopic_validator.go
panyuenlau: Perhaps having the following is clear enough:

`"When creating KafkaTopic CR for existing topic, its partition number must be the same as what the existing kafka topic has..."`

What do you think?

charts/kafka-operator/templates/crds.yaml
panyuenlau: yes, this has been taken care of with the latest commits

Kuvesz (approved):

Kuvesz (approved):

charts/kafka-operator/templates/crds.yaml
```diff
@@ -3907,6 +3907,21 @@ spec:
                         external listener advertise address according to the description
                         of the "hostnameOverride" field.
                       type: object
+                    nodePortNodeAddressType:
```
pregnor, line 3910: **Question**: are we sure this should be here?
I think this should only be in your other PR.

pregnor (approved):
LGTM with 1 comment

pkg/webhooks/kafkatopic_validator_test.go
```diff
@@ -72,6 +74,141 @@ func newMockClients(cluster *v1beta1.KafkaCluster) (runtimeClient.WithWatch, kaf
...
+				},
+			},
+			expectedErrors: []string{"its configuration must be"},
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.testName, func(t *testing.T) {
+			t.Parallel()
+			fieldErrorList, err := kafkaTopicValidator.checkKafka(context.Background(), &testCase.kafkaTopic, cluster)
```
mihaialexandrescu, line 198: Does this section of code suffer from the "loop var + closures + goroutines" gotcha from Go ?

I tried adding a `t.Log(testCase)` beneath this line and the output always has `testName` as `topic configuration is different and managedBy koperator` (the last test case).
```
=== NAME  TestCheckKafkaTopicExist/topic_configuration_is_same_and_managedBy_koperator
    kafkatopic_validator_test.go:204: {topic configuration is different and managedBy koperator {{ } {      0 {{0 0 <nil>}} <nil> <nil> map[] map[managedBy:koperator] [] [] []} {test-topic 2 1 map[testConfKey:testConfVal1] { }} { }} [its configuration must be]}
=== NAME  TestCheckKafkaTopicExist/topic_replication_factor_is_different_and_managedBy_koperator
    kafkatopic_validator_test.go:204: {topic configuration is different and managedBy koperator {{ } {      0 {{0 0 <nil>}} <nil> <nil> map[] map[managedBy:koperator] [] [] []} {test-topic 2 1 map[testConfKey:testConfVal1] { }} { }} [its configuration must be]}
=== NAME  TestCheckKafkaTopicExist/topic_partition_and_replication_is_different_and_not_managedBy_koperator
    kafkatopic_validator_test.go:204: {topic configuration is different and managedBy koperator {{ } {      0 {{0 0 <nil>}} <nil> <nil> map[] map[managedBy:koperator] [] [] []} {test-topic 2 1 map[testConfKey:testConfVal1] { }} { }} [its configuration must be]}
```

pregnor: Looks to me like a proper full name.
`TestCheckKafkaTopicExist/topic_configuration_is_same_and_managedBy_koperator`
Not sure what you mean exactly.

pregnor: Oh, I see, I think after the for there should be a 
```go
testCase := testCase
```
line which fixes this.
(I use a linter that alerts on this.)

Thanks for catching Mihai.

PR #939, opened 2023-02-27
Add Listener port validation

### LINKED ISSUE #913: Broker.Id data type is incorrect and some within range numbers would result in reconciler error
**Describe the bug**

Currently, the `KafkaCluster.spec.Broker.Id` field has type `int32`,  with supported range [-2,147,483,648, 2,147,483,647]. However, it is not possible to set the broker id to be negative, otherwise users would see the following error:
```
{"level":"error","ts":"2022-12-15T16:50:28.951Z","msg":"Reconciler error","controller":"KafkaCluster","controllerGroup":"kafka.banzaicloud.io","controllerKind":"KafkaCluster","KafkaCluster":{"name":"kafka","namespace":"kafka"},"namespace":"kafka","name":"kafka","reconcileID":"1b3299b0-a1dc-4bce-8f27-b809e7ccc374","error":"failed to reconcile resource: creating resource failed: PersistentVolumeClaim \"kafka--1-storage-0-prpfx\" is invalid: metadata.labels: Invalid value: \"-1\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')","errorVerbose":"creating resource failed: PersistentVolumeClaim \"kafka--1-storage-0-prpfx\" is invalid: metadata.labels: Invalid value: \"-1\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')\nfailed to reconcile resource\ngithub.com/banzaicloud/koperator/pkg/resources/kafka.(*Reconciler).Reconcile\n\t/workspace/pkg/resources/kafka/kafka.go:273\ngithub.com/banzaicloud/koperator/controllers.(*KafkaClusterReconciler).Reconcile\n\t/workspace/controllers/kafkacluster_controller.go:125\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/internal/controller/controller.go:121\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/internal/controller/controller.go:320\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/internal/controller/controller.go:273\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/internal/controller/controller.go:234\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1594","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/internal/controller/controller.go:326\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/internal/controller/controller.go:273\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/internal/controller/controller.go:234"}
{"level":"info","ts":"2022-12-15T17:07:09.005Z","msg":"failed to reconcile resource: creating resource failed: PersistentVolumeClaim \"kafka--1-storage-0-l95xh\" is invalid: metadata.labels: Invalid value: \"-1\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')","controller":"KafkaCluster","controllerGroup":"kafka.banzaicloud.io","controllerKind":"KafkaCluster","KafkaCluster":{"name":"kafka","namespace":"kafka"},"namespace":"kafka","name":"kafka","reconcileID":"2fb20bd0-b0db-41bf-bac8-804790ae6cab"}
```

Also, setting the broker id to be `65535`(which is within the range of `int32` and also positive) would result in the following error:
```
{"level":"info","ts":"2022-12-15T20:42:03.762Z","msg":"updating resource failed: admission webhook \"rev.validation.istio.io\" denied the request: configuration is invalid: port number 84625 must be in the range 1..65535","controller":"KafkaCluster","controllerGroup":"kafka.banzaicloud.io","controllerKind":"KafkaCluster","KafkaCluster":{"name":"kafka","namespace":"kafka"},"namespace":"kafka","name":"kafka","reconcileID":"93e69d4b-d3fa-4142-bdd7-0a8f9a905850"}
{"level":"error","ts":"2022-12-15T20:42:03.762Z","msg":"Reconciler error","controller":"KafkaCluster","controllerGroup":"kafka.banzaicloud.io","controllerKind":"KafkaCluster","KafkaCluster":{"name":"kafka","namespace":"kafka"},"namespace":"kafka","name":"kafka","reconcileID":"93e69d4b-d3fa-4142-bdd7-0a8f9a905850","error":"updating resource failed: admission webhook \"rev.validation.istio.io\" denied the request: configuration is invalid: port number 84625 must be in the range 1..65535","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/internal/controller/controller.go:326\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/internal/controller/controller.go:273\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/internal/controller/controller.go:234"}
```

Set one of the broker id in `KafkaCluster` spec to the numbers described above

**Expected behavior**

Koperator has correct range for broker id at data type definition

mihaialexandrescu:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets | fixes #913 |

### What's in this PR?

This PR aims to add some validation for the port numbers under `spec.listenersConfig` of the `KafkaCluster` CR.

Example config pointing out the fields this PR adds validation for : 
```
apiVersion: kafka.banzaicloud.io/v1beta1
kind: KafkaCluster
spec:
  brokers:
    - id: 0
      brokerConfigGroup: "default"
    ... ...
  ... ...
  listenersConfig:
    internalListeners:
      - type: "plaintext"
        name: "internal"
        containerPort: 29092     <<<  
        usedForInnerBrokerCommunication: true
      ... ...
    externalListeners:
      - type: "plaintext"
        name: "external1"
        externalStartingPort: 19090     <<< 
        containerPort: 9094    <<<
      ... ...
  ... ...
```

The validation logic added in this PR falls under 2 categories : 
1. OpenAPIv3 Schema validations (via kubebuilder markers)
  * this performs an initial sanity check (value between 1 and 65535; 0 is acceptable for externalStartingPort) directly on the input value
  * applied in this PR to both `containerPort` (via `CommonListenerSpec`; applies to both internal and external listeners) and to `externalListeners[].externalStartingPort`
2. New functions added to the validating webhook (for both `Create` and `Update` operations) for `KafkaCluster` to check that:
  * the external port we get from `externalStartingPort + brokers[].id` is sane (1 < value < 65535)
  * the `containerPort` is not duplicated between listeners because that would lead to a `Duplicate value` error when creating the Kafka broker Service (headless or "all-brokers") (just a general rule of K8s Services)
    * sample of such an error for the headless service (truncated ;  note the same `Port` reused under different `spec.ports[index]`): 
      > "errorVerbose":"creating resource failed: Service \"kafka-headless\" is invalid: [spec.ports[1]: Duplicate value: core.ServicePort{Name:\"\", Protocol:\"TCP\", AppProtocol:(*string)(nil), Port:29095, TargetPort:intstr.IntOrString{Type:0, IntVal:0, StrVal:\"\"}, NodePort:0}, spec.ports[3]: Duplicate value: core.ServicePort{Name:\"\", Protocol:\"TCP\", AppProtocol:(*string)(nil), Port:29095, TargetPort:intstr.IntOrString{Type:0, IntVal:0, StrVal:\"\"}, NodePort:0}]

This PR also includes :
* updates to the validatingwebhook CR for the RBAC to register for `Create` operations (as well as the already existing `Update` operations) and what I thought are the necessary translations of that into the helm charts.
* unit tests for the new functions used by the webhook.
* a new sentinel error for the externalStartingPort case, an exported function to help "recognize" it and the unit test for it (`IsAdmissionInvalidExternalListenerPort()`)

The new unit tests use [testify/require](https://pkg.go.dev/github.com/stretchr/testify/require).  
**Outdated:** I chose to do matching with the `ElementsMatch` function because order doesn't matter even though it is deterministic. With the current test cases the `Equal` function could be used just as well.  
**Update:** In `kafkacluster_validator_test.go` matching is done with `require.Equal()` per the advice and explanations received during review.

### Why?

I reckon this deals with most of #913 (and a bit more).

### Additional context

Tested on a real cluster (with `make deploy`) and, for the webhook, tested both Create and Update operations (with both `kubectl apply` and `kubectl edit` for `Update`).

Example inputs and results:

1. OpenAPIv3 Schema validation:

```
    externalListeners:
      - type: "plaintext"
        name: "external1"
        externalStartingPort: 65536   << illegal value
        containerPort: 9094
```
Results in :

```
The KafkaCluster "kafka" is invalid: spec.listenersConfig.externalListeners.externalStartingPort: Invalid value: 65536: spec.listenersConfig.externalListeners.externalStartingPort in body should be less than or equal to 65535
```

Also:

```
    internalListeners:
      - type: "plaintext"
        name: "internal"
        containerPort: 89092
        usedForInnerBrokerCommunication: true
```

Results in :

```
The KafkaCluster "kafka" is invalid: spec.listenersConfig.internalListeners.containerPort: Invalid value: 89092: spec.listenersConfig.internalListeners.containerPort in body should be less than or equal to 65535
```

2. Webhook validation :

```
apiVersion: kafka.banzaicloud.io/v1beta1
kind: KafkaCluster
spec:
  brokers:
    - id: 0
      brokerConfigGroup: "default"
    - id: 1
      brokerConfigGroup: "default"
    - id: 200
      brokerConfigGroup: "default"
  listenersConfig:
    internalListeners:
      - type: "plaintext"
        name: "internal"
        containerPort: 29095    <<< subsequent uses of this port number (regardless of listener type) are reported as "Duplicate value" errors
        usedForInnerBrokerCommunication: true
      - type: "plaintext"
        name: "controller"
        containerPort: 29095     <<< duplicate value
        usedForInnerBrokerCommunication: false
        usedForControllerCommunication: true
    externalListeners:
      - type: "plaintext"
        name: "external1"
        externalStartingPort: 65535   <<< causes external port for brokers 1 and 200 to be out of range
        containerPort: 9094
      - type: "plaintext"
        name: "external2"
        externalStartingPort: 65400   <<< causes external port for broker 200 to be out of range
        containerPort: 29095     <<< duplicate value
```

Results in :

```
The KafkaCluster "kafka" is invalid:
* spec.listenersConfig.internalListeners[1].containerPort: Duplicate value: 29095
* spec.listenersConfig.externalListeners[1].containerPort: Duplicate value: 29095
* spec.listenersConfig.externalListeners[0].externalStartingPort: Invalid value: 65535: invalid external listener starting port number: ExternalListener 'external1' would generate external access port numbers (externalStartingPort + Broker ID) that are out of range (not between 1 and 65535) for brokers [1 200]
* spec.listenersConfig.externalListeners[1].externalStartingPort: Invalid value: 65400: invalid external listener starting port number: ExternalListener 'external2' would generate external access port numbers (externalStartingPort + Broker ID) that are out of range (not between 1 and 65535) for brokers [200]
```

We get the same messages/errors when using `kubectl apply/create` and when using `kubectl edit` (depending on the operation Create/Update being performed).

#### externalListeners[].accessMethod: NodePort

The checks proposed in this PR address generic well-known bounds for port numbers.  
They don't address in any way the more restrictive port range at play when `externalListeners[].accessMethod: NodePort` -->  the issue I detect here is that the nodePort range is a kubeapiserver input flag/parameter and I don't know how to query for that... also, even if we could do that we can't guarantee that a value is not already taken by another service. For this case, I think we'll have to keep relying on errors from the K8s `Service` controller during deployment.

#### Naming scheme

Please also provide feedback on the name and message string of the new sentinel error in `pkg/webhooks/errors.go` and the exported function meant to "recognize it" (`IsAdmission...()`).

Please let me know if there are other cases I missed in the unit tests or, worse, the logic.


pkg/webhooks/errors_test.go
```diff
@@ -166,6 +166,41 @@ func TestIsAdmissionInvalidRemovingStorage(t *testing.T) {
...
+	for _, tc := range testCases {
+		t.Run(tc.testName, func(t *testing.T) {
+			kafkaCluster := banzaicloudv1beta1.KafkaCluster{ObjectMeta: metav1.ObjectMeta{Name: "test-KafkaCluster"}}
+
+			err := apierrors.NewInvalid(
+				kafkaCluster.GetObjectKind().GroupVersionKind().GroupKind(),
+				kafkaCluster.Name, tc.fieldErrs)
+
+			if got := IsAdmissionInvalidExternalListenerPort(err); got != tc.want {
+				t.Errorf("Check External Listener Port Error message. Expected: %t ; Got: %t", tc.want, got)
```
panyuenlau, line 198: Something that I learned from Patrik: https://github.com/banzaicloud/koperator/pull/928#discussion_r1103573247

The same comment goes to other `t.Errorf` in this PR

pkg/webhooks/kafkacluster_validator.go
```diff
@@ -39,6 +39,8 @@ func (s KafkaClusterValidator) ValidateUpdate(ctx context.Context, oldObj, newOb
 	kafkaClusterOld := oldObj.(*banzaicloudv1beta1.KafkaCluster)
 	kafkaClusterNew := newObj.(*banzaicloudv1beta1.KafkaCluster)
 	log := s.Log.WithValues("name", kafkaClusterNew.GetName(), "namespace", kafkaClusterNew.GetNamespace())
+
+	// check storage removal
```
panyuenlau, line 43: Suggestion: This comment in unnecessary. 

Something that I learned from reading a book called [Clean Code](https://books.google.com/books/about/Clean_Code.html?id=hjEFCAAAQBAJ&source=kp_book_description)(which I highly recommend) is that adding comment that is not needed to the source code is actually considered a bad coding behavior because it requires the audience to spend time on reading the comment. 

And I think this comment falls into the "unnecessary comment" category because the next line calls a function named `checkBrokerStorageRemoval`, which is self-explanatory enough

pkg/webhooks/kafkacluster_validator.go
```diff
@@ -47,17 +49,42 @@ func (s KafkaClusterValidator) ValidateUpdate(ctx context.Context, oldObj, newOb
 	if fieldErr != nil {
 		allErrs = append(allErrs, fieldErr)
 	}
+
+	// check internal and external listeners
```
panyuenlau, line 53: Same as above, I don't really see the need to leave the comment here, if there is, I would rename the function to be something more explicit, e.g. `checkInternalAndExternalListeners`

pkg/webhooks/kafkacluster_validator.go
```diff
@@ -133,3 +160,67 @@ func getMissingMounthPathLocation(mounthPath string, kafkaClusterSpec *banzaiclo
...
+// checkExternalListenerStartingPort checks the generic sanity of the resulting external port (valid number between 1 and 65535)
+func checkExternalListenerStartingPort(kafkaClusterSpec *banzaicloudv1beta1.KafkaClusterSpec) field.ErrorList {
+	// if there are no externalListeners, there is no need to perform the rest of the checks in this function
+	if kafkaClusterSpec.ListenersConfig.ExternalListeners == nil {
+		return nil
+	}
+
+	var allErrs field.ErrorList
+	const maxPort int32 = 65535
+	for i, extLstn := range kafkaClusterSpec.ListenersConfig.ExternalListeners {
```
panyuenlau, line 212: Suggestion: `extLstn` is a bit ambitious to me
```suggestion
	for i, extListener := range kafkaClusterSpec.ListenersConfig.ExternalListeners {
```
But it's fine if you feel like keeping it like this since it is just used within the for loop

pkg/webhooks/kafkacluster_validator.go
```diff
@@ -133,3 +160,67 @@ func getMissingMounthPathLocation(mounthPath string, kafkaClusterSpec *banzaiclo
...
+		var errbrokerids []int32
+		for _, broker := range kafkaClusterSpec.Brokers {
+			if extLstn.ExternalStartingPort+broker.Id < 1 || extLstn.ExternalStartingPort+broker.Id > maxPort {
+				errbrokerids = append(errbrokerids, broker.Id)
+			}
+		}
+		if len(errbrokerids) > 0 {
```
panyuenlau, line 219:
```suggestion
		if len(invalidBrokerIDs) > 0 {
```

pkg/webhooks/kafkacluster_validator_test.go
```diff
@@ -486,3 +487,221 @@ func TestCheckBrokerStorageRemoval(t *testing.T) {
 		}
 	}
 }
+
+func TestCheckExternalListenerStartingPort(t *testing.T) {
```
panyuenlau, line 491: Question about the combination of broker ID and `ExternalStartingPort` used in the test: is it possible to use some randomly generated numbers that meet the expected behavior? 

The main reason behind this ask is that the current implementation always uses the fixed combination of broker IDs and `ExternalStartingPort` no matter how many times we run the unit test, which is not very ideal from the testing perspective. The best scenario is to use some randomly generated numbers within certain range every time that the test is run.

pkg/webhooks/kafkacluster_validator_test.go
```diff
@@ -486,3 +487,221 @@ func TestCheckBrokerStorageRemoval(t *testing.T) {
...
+	testCases := []struct {
+		testName         string
+		kafkaClusterSpec v1beta1.KafkaClusterSpec
+		isValid          bool
+	}{
+		{
+			// In this test case, all resulting external port numbers should be valid
+			testName: "valid-2-brokers-2-externalListeners",
```
panyuenlau, line 499: Question: is there a particular reason to add the dashes in the `testName`? If not, removing the dashes would make it a bit more readable IMO

```suggestion
			testName: "valid configurations: 3 brokers with 2 externalListeners",
```

panyuenlau (commented):
Thanks for the contribution! Left some comments

pkg/webhooks/kafkacluster_validator_test.go
mihaialexandrescu: The reason was to keep it the same in the `go test` output (subtests replace white spaces with `_`) as in code but that's not important. It is probably better to use the spaces. I'll update it.

pkg/webhooks/kafkacluster_validator_test.go
mihaialexandrescu: > current implementation always uses the fixed combination of broker IDs and ExternalStartingPort no matter how many times we run the unit test

The test does that because that's what we do in the operator code.

> is it possible to use some randomly generated numbers that meet the expected behavior?

A longwinded reply about random numbers for those 2 user inputs (broker ids and startingExternalPort) : 
1. I don't think that's attainable because the main condition we ultimately check against in this function is `RNG_1 + RNG_2 <= 65535` which translates to nothing meaningful in limiting the range of the 2 numbers other than that both have to be < 65535. 
    * We do that in this PR via the OpenAPIv3Schema validations (markers) but because that is such a weak condition (means close to nothing, in practice) we add the validating webhook function to check the sum for more practical relevance.
    * How would you generate numbers that don't "obviously/by default" satisfy these conditions **AND know in advance** that the case should result in true/false (aka numbers and the sum are valid/invalid) ?
2. I believe the current setup proves considerably more even as documentation value because the current 3 test cases within that function translate in natural language to : 
    * [1] obviously ok / well in range
    * [2] obviously bad / well out of range
    * [3] borderline values - some results are in range, some are out of range but sometimes barely out of range which emphasises the limits of the intended interval.     

I believe addressing these cases is also part of general testing theory.

Due to the lack of a meaningful mathematical condition that we could enforce **a-priori**, I think random numbers could only replace the ones in the "obviously ok/bad" cases (never the borderline cases) and that, in my opinion, doesn't add value to our tests.   
Yes, in the "obvious" cases, I pulled those broker ids and startingPorts out of "thin air" but making them smth like (for the "obviously ok" case):
```
brokers[0-2].id = rng(0<x<999)
listenersConfig.[...int/ext...].externalStartingPort = rng(15000 < x < 50000)
```
is, imo, just a headache without a payoff.

Does that thought process make sense ?

Kuvesz (commented):
Looks good to me with the update Darren proposed regarding the `t.Errorf` vs `Require` logic!

Kuvesz (approved):
LGTM! Thanks for the work. :)

pkg/webhooks/errors_test.go
mihaialexandrescu: Like 4f1fcdc1eaa51f5764a69e39c3ad860e667b8a55 ?

One more thing:   
I chose to do matching with the `ElementsMatch` function because order of the `*field.Error` elements in the `field.ErrorList` doesn't matter (IMO) even though it is deterministic. With the current test cases the `Equal` function could be used just as well.  
Do we prefer `Equal` instead ? (to be 100% precise)

panyuenlau: > Do we prefer Equal instead ? (to be 100% precise)

Compared to `t.Errorf`, yes, since it's a bit cleaner

mihaialexandrescu: I was unclear.  
I meant : do we prefer `require.Equal()` or `require.ElementsMatch()` ?

pkg/webhooks/kafkacluster_validator_test.go
panyuenlau: I disagree that using random numbers doesn't add value to the tests, I am a bit lazy of writing a lot of text so let's look at the code...

```golang

	const lowerEnd int32 = 30000

	testCases := []struct {
		testName         string
		kafkaClusterSpec v1beta1.KafkaClusterSpec
		expected         field.ErrorList
	}{
		{
			// In this test case, all resulting external port numbers should be valid
			testName: "valid config: 3 brokers with 2 externalListeners",
			kafkaClusterSpec: v1beta1.KafkaClusterSpec{
				Brokers: []v1beta1.Broker{{Id: rand.Int31n(lowerEnd)}, {Id: rand.Int31n(lowerEnd)}, {Id: rand.Int31n(lowerEnd)}},
				ListenersConfig: v1beta1.ListenersConfig{
					ExternalListeners: []v1beta1.ExternalListenerConfig{
						{
							CommonListenerSpec:   v1beta1.CommonListenerSpec{Name: "test-external1"},
							ExternalStartingPort: rand.Int31n(65535 - lowerEnd),
						},
						{
							CommonListenerSpec:   v1beta1.CommonListenerSpec{Name: "test-external2"},
							ExternalStartingPort: rand.Int31n(65535 - lowerEnd),
						},
					},
				},
			},
			expected: nil,
		},
...

```

This way we make sure that the `checkExternalListenerStartingPort` actually behaves as we expected with different combination of numbers with the broker.id and the starting port number every time that we run the unit test, which I think it's much better

mihaialexandrescu: I think I see what you mean. I like it for this case (the previously deemed "obviously ok" case).  
Do you see it applied on other cases ?

panyuenlau: > I think I see what you mean. I like it for this case (the previously deemed "obviously ok" case aka when `expected` is always `nil`). Do you see it applied on other cases ?

yea, I think we can apply the same logic to the other two test cases under this test function by setting the range for the randomly generated numbers

pregnor: While Darren's proposal does have it's merit, it should be done during fuzzy testing IMO instead of unit testing.
In unit tests I would rather adhere to the TDD FIRST principle: https://stackoverflow.com/questions/18024785/tdd-first-principle in which the repeatability/reproducibility would be much better with non-random input data even if semantically the results should be the same. (The reason is unexpected breakings depending on the random input.)

pkg/webhooks/kafkacluster_validator.go
```diff
@@ -133,3 +157,65 @@ func getMissingMounthPathLocation(mounthPath string, kafkaClusterSpec *banzaiclo
...
+// The first time a port number is found will not be reported as duplicate; only subsequent instances using that port are.
+// (this is done to keep in tune with the way the K8s Service would report the "Duplicate value" error which ignores the first instance)
+func checkUniqueListenerContainerPort(listeners banzaicloudv1beta1.ListenersConfig) field.ErrorList {
+	var allErrs field.ErrorList
+	var containerPorts = make(map[int32]int)
+
+	for i, intListener := range listeners.InternalListeners {
+		containerPorts[intListener.ContainerPort] += 1
+		if containerPorts[intListener.ContainerPort] > 1 {
+			fldErr := field.Duplicate(field.NewPath("spec").Child("listenersConfig").Child("internalListeners").Index(i).Child("containerPort"), intListener.ContainerPort)
```
pregnor, line 183: **Question**: Wouldn't it be better for the user experience to report the listener names of the duplicate ports here? (storing the names in a `map[int32][]string` and when the length is over 1 at the end of the iteration then report all so the user knows what to fix exactly)

Same for the iteration following this one on external listeners

pregnor (dismissed):
LGTM at a glance with 1 comment.

pkg/webhooks/kafkacluster_validator_test.go
mihaialexandrescu: > I think we can apply the same logic to the other two test cases under this test function by setting the range for the randomly generated numbers

I was looking today at it and one challenge I see for the cases when `expected` is not `nil` (especially in the context of [this thread](https://github.com/banzaicloud/koperator/pull/939#discussion_r1119122154) and us going the "exact match" route) is defining the "precise" `expected` results because they would need to include most of these random numbers and the name of the listeners which fail.

For example (snippet partially defining `expected` from another test case ):  
```
expected: append(field.ErrorList{},
	field.Invalid(field.NewPath("spec").Child("listenersConfig").Child("externalListeners").Index(0).Child("externalStartingPort"), int32(65535),
		invalidExternalListenerStartingPortErrMsg+": "+fmt.Sprintf("ExternalListener '%s' would generate invalid port numbers (not between 1 and 65535) for brokers %v",
			"test-external1", []int32{11, 102})),
... ... ...
```

In that example, I don't know how to apply the random numbers idea to replace : 
* `int32(65535)`
    * this is the `value interface{}` parameter of the `field.Invalid()` function
    * this challenge applies to generating any `field.Error` aka as long as `expected` is not `nil`
 
Similar idea applies to these but they are not unavoidable in the webhook function: 
* `[]int32{11, 102}`
    * this is a bit of helping info about which broker ids are incompatible with the externalListener 
* `"test-external1"`
    * least important, just makes the log a bit nicer in my opinion but I do like having it in the log message

pkg/webhooks/kafkacluster_validator.go
```diff
@@ -133,3 +157,65 @@ func getMissingMounthPathLocation(mounthPath string, kafkaClusterSpec *banzaiclo
...
+	const maxPort int32 = 65535
+	for i, extListener := range kafkaClusterSpec.ListenersConfig.ExternalListeners {
+		var invalidBrokerIDs []int32
+		for _, broker := range kafkaClusterSpec.Brokers {
+			if extListener.ExternalStartingPort+broker.Id < 1 || extListener.ExternalStartingPort+broker.Id > maxPort {
+				invalidBrokerIDs = append(invalidBrokerIDs, broker.Id)
+			}
+		}
+		if len(invalidBrokerIDs) > 0 {
+			errmsg := invalidExternalListenerStartingPortErrMsg + ": " + fmt.Sprintf("ExternalListener '%s' would generate invalid port numbers (not between 1 and 65535) for brokers %v", extListener.Name, invalidBrokerIDs)
```
bartam1, line 215: I think we should chage a bit on this error message ""ExternalListener '%s' would generate invalid port numbers (not between 1 and 65535) for brokers %v"
Like: for externalListener '%s' the generated port numbers for external access (externalStartingPort '%d' + brokerID '%v') is not between 1 and 65535"

pkg/webhooks/kafkacluster_validator.go
mihaialexandrescu: For reference, this is what we log in the current interaction of this PR for that part of the code:
```
The KafkaCluster "kafka" is invalid:
* spec.listenersConfig.internalListeners[1].containerPort: Duplicate value: 29095
* spec.listenersConfig.externalListeners[1].containerPort: Duplicate value: 29095
```

I like the idea of having the listener names as part of what we report back (it also fits with what is done in the other validation from this PR).  
I am, however, not clear on what you're suggesting. 

Right now, we iterate through the config and as we find a repeated containerPort, we add a `field.Error` (of the `field.Duplicate()` kind) pointing to that specific `containerPort` field (e.g. spec.listenersConfig.internalListeners[1].containerPort)

I understand how we can get aggregate the info (small change; the map definition you suggested) but I don't understand to what field in the spec we would associate the error to, in this case.

<br></br>
Side note:
It is true that one downside of `field.Duplicate()` is that it doesn't allow us to add any `detail string` like `field.Invalid` does.
```
func field.Duplicate(field *field.Path, value interface{}) *field.Error

func field.Invalid(field *field.Path, value interface{}, detail string) *field.Error
```

bartam1 (commented):
LGTM! Have one suggestion.

pkg/webhooks/kafkacluster_validator_test.go
panyuenlau: Thanks @pregnor for the pointer! I wasn't aware the TDD FIRST principle in testing due to experience. Based on that, I think you can ignore my original comment now @mihaialexandrescu

pkg/webhooks/kafkacluster_validator.go
pregnor: Oh, I see, I didn't think this through.
The field would be the same, only the details would be different, but I overlooked the fact in this place we cannot define details for the duplicate error so I think you can leave it the way it is, because the listener config index is already encoded in the error message.

pkg/webhooks/errors_test.go
panyuenlau: > do we prefer `require.Equal()` or `require.ElementsMatch()` ?

To make a good decision on this, I checked what these two functions for, and found that
- `require.Equal()` is to "assert that two objects are equal", and "Pointer variable equality is determined based on the equality of the referenced values (as opposed to the memory addresses). Function equality cannot be determined and will always fail."
- And `require.ElementsMatch()` is to "assert that the specified listA(array, slice...) is equal to specified listB(array, slice...) ignoring the order of the elements. If there are duplicate elements, the number of appearances of each of them in both lists should match."

So, `require.ElementsMatch()` is a better choice if you are comparing two slices and arrays while ignoring the order of the elements. And `require.Equal()` is to compare two objects in general

But above is just my understanding after diving into a bit of the source code of those two methods, it'd be great if @pregnor can correct me if I am mistaken somewhere

mihaialexandrescu: In this line in the `errors_test.go` file `require.Equal()` is the way to go.  

The question about `require.Equal()` vs `require.ElementsMatch()` targets the `kafkacluster_validator_test.go` file where the functions we test return [`field.ErrorList`](https://pkg.go.dev/k8s.io/apimachinery@v0.26.1/pkg/util/validation/field#ErrorList) (which is a `[]*field.Error`).

I'm curious about you guys' opinion on Equal() vs ElementsMatch() because, to rephrase a part of a previous comment, order in that field error list : 
* is deterministic   -->   so using require.Equal() is an option
* but order doesn't matter and it isn't characteristic to the list or its use  -->  so I instinctively chose require.ElementsMatch()
    * also, it makes it a bit easier to write the `expected` results in the test cases since order wouldn't matter

panyuenlau: Ahhh I see, I looked into the implementation of both methods but still didn't determine which one is better for comparing `field.ErrorList`, so I tried to break the test and see what I get from them:

With `require.Equal()`:

```
=== RUN   TestCheckExternalListenerStartingPort/invalid_config:_3_brokers_with_2_at-the-limit_externalListeners
    kafkacluster_validator_test.go:692: 
        	Error Trace:	/Users/darrelau/go/src/github.com/cisco-app-networking/SDM/koperator/pkg/webhooks/kafkacluster_validator_test.go:692
        	Error:      	Not equal: 
        	            	expected: field.ErrorList{(*field.Error)(0xc0003c33c0), (*field.Error)(0xc0003c3480)}
        	            	actual  : field.ErrorList{(*field.Error)(0xc0003c3800), (*field.Error)(0xc0003c38c0)}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -11,3 +11,3 @@
        	            	   BadValue: (int32) 65434,
        	            	-  Detail: (string) (len=169) "invalid external listener starting port number: ExternalListener 'test-external2' would generate invalid port numbers (not between 1 and 65535) for brokers [102] - debug"
        	            	+  Detail: (string) (len=161) "invalid external listener starting port number: ExternalListener 'test-external2' would generate invalid port numbers (not between 1 and 65535) for brokers [102]"
        	            	  })
        	Test:       	TestCheckExternalListenerStartingPort/invalid_config:_3_brokers_with_2_at-the-limit_externalListeners
```

With `require.ElementsMatch()`:
```
    kafkacluster_validator_test.go:692: 
        	Error Trace:	/Users/darrelau/go/src/github.com/cisco-app-networking/SDM/koperator/pkg/webhooks/kafkacluster_validator_test.go:692
        	Error:      	elements differ
        	            	
        	            	extra elements in list A:
        	            	([]interface {}) (len=1) {
        	            	 (*field.Error)({
        	            	  Type: (field.ErrorType) (len=17) "FieldValueInvalid",
        	            	  Field: (string) (len=62) "spec.listenersConfig.externalListeners[1].externalStartingPort",
        	            	  BadValue: (int32) 65434,
        	            	  Detail: (string) (len=169) "invalid external listener starting port number: ExternalListener 'test-external2' would generate invalid port numbers (not between 1 and 65535) for brokers [102] - debug"
        	            	 })
        	            	}
        	            	
        	            	
        	            	extra elements in list B:
        	            	([]interface {}) (len=1) {
        	            	 (*field.Error)({
        	            	  Type: (field.ErrorType) (len=17) "FieldValueInvalid",
        	            	  Field: (string) (len=62) "spec.listenersConfig.externalListeners[1].externalStartingPort",
        	            	  BadValue: (int32) 65434,
        	            	  Detail: (string) (len=161) "invalid external listener starting port number: ExternalListener 'test-external2' would generate invalid port numbers (not between 1 and 65535) for brokers [102]"
        	            	 })
        	            	}
        	            	
        	            	
        	            	listA:
        	            	(field.ErrorList) (len=2) {
        	            	 (*field.Error)({
        	            	  Type: (field.ErrorType) (len=17) "FieldValueInvalid",
        	            	  Field: (string) (len=62) "spec.listenersConfig.externalListeners[0].externalStartingPort",
        	            	  BadValue: (int32) 65535,
        	            	  Detail: (string) (len=164) "invalid external listener starting port number: ExternalListener 'test-external1' would generate invalid port numbers (not between 1 and 65535) for brokers [11 102]"
        	            	 }),
        	            	 (*field.Error)({
        	            	  Type: (field.ErrorType) (len=17) "FieldValueInvalid",
        	            	  Field: (string) (len=62) "spec.listenersConfig.externalListeners[1].externalStartingPort",
        	            	  BadValue: (int32) 65434,
        	            	  Detail: (string) (len=169) "invalid external listener starting port number: ExternalListener 'test-external2' would generate invalid port numbers (not between 1 and 65535) for brokers [102] - debug"
        	            	 })
        	            	}
        	            	
        	            	
        	            	listB:
        	            	(field.ErrorList) (len=2) {
        	            	 (*field.Error)({
        	            	  Type: (field.ErrorType) (len=17) "FieldValueInvalid",
        	            	  Field: (string) (len=62) "spec.listenersConfig.externalListeners[0].externalStartingPort",
        	            	  BadValue: (int32) 65535,
        	            	  Detail: (string) (len=164) "invalid external listener starting port number: ExternalListener 'test-external1' would generate invalid port numbers (not between 1 and 65535) for brokers [11 102]"
        	            	 }),
        	            	 (*field.Error)({
        	            	  Type: (field.ErrorType) (len=17) "FieldValueInvalid",
        	            	  Field: (string) (len=62) "spec.listenersConfig.externalListeners[1].externalStartingPort",
        	            	  BadValue: (int32) 65434,
        	            	  Detail: (string) (len=161) "invalid external listener starting port number: ExternalListener 'test-external2' would generate invalid port numbers (not between 1 and 65535) for brokers [102]"
        	            	 })
        	            	}
        	Test:       	TestCheckExternalListenerStartingPort/invalid_config:_3_brokers_with_2_at-the-limit_externalListeners
```

I think either method works for comparing `field.ErrorList`, but I think `require.Equal()` provides a better test output on failures

pkg/webhooks/kafkacluster_validator.go
mihaialexandrescu: For reference, currently, the message in question looks like this:
```
The KafkaCluster "kafka" is invalid:
* spec.listenersConfig.externalListeners[0].externalStartingPort: Invalid value: 65535: invalid external listener starting port number: ExternalListener 'external1' would generate invalid port numbers (not between 1 and 65535) for brokers [1 200]
* spec.listenersConfig.externalListeners[1].externalStartingPort: Invalid value: 65400: invalid external listener starting port number: ExternalListener 'external2' would generate invalid port numbers (not between 1 and 65535) for brokers [200]
```

How do you feel about updating **the last part** to look like this ?   
>ExternalListener 'external1' would generate external access port numbers (externalStartingPort 65535 + Broker IDs [11 102]) that are out of range (not between 1 and 65535)



A couple of notes:
1. I like the idea of inserting the "formula" in the log because that makes it significantly clearer for the user what the problem is
2. Having the problematic externalStartingPort and broker IDs right next to each other in the log could make it easier for our users to interpret the message
3. I favour the use of the conditional "would" in a validating webhook context because using past tense ("generated") gives me, as a user, the feeling that something already happened in the background (objects were created, who knows if they were cleaned up...) in the cluster which is both disconcerting and not true in this case.
4. I don't like "brokerIDs" as part of the string because it looks like a field name (and it isn't ;  for us the actual path would be `brokers[].id`). That's why I'd go with either `brokers.id %v` or `Broker IDs %v` to not give the impression that we're referencing a hidden/unknown field from the spec.

panyuenlau: IMO the current error message is very clear

panyuenlau (dismissed):
Feel free to merge

pkg/webhooks/kafkacluster_validator.go
bartam1: We should put somehow the formula. I think it is not so clear for the user. Most of the cases the users will not know how it is generated. Also it is hard to find in our documentation. I just like to see the formula with or without the values. 
Agree with the 3 and 4

mihaialexandrescu: Here are a couple of examples that I think follow those thoughts :

1. formula added as an explanation and numbers are right in it including repeating the externalStartingPort :
    * `ExternalListener 'external1' would generate external access port numbers (externalStartingPort 65535 + Broker IDs [11 102]) that are out of range (not between 1 and 65535)`
        * this keeps relevant pieces of information (those user inputs) closer to each other 

2. formula added as an explanation but without the numbers right in it and without repeating the externalStartingPort value :
    * `ExternalListener 'external1' would generate external access port numbers (externalStartingPort + Broker ID) that are out of range (not between 1 and 65535) for brokers [11 102]`

What do you think about these ?

bartam1: I vote for  2. :)

pregnor (dismissed):

pkg/webhooks/kafkacluster_validator.go
mihaialexandrescu: Done in commit c6f0ea75642551310b64e87faa4969e6c89e5f3b.

I also updated the PR description with the new log message (taken from a real install running this code against the same input as before).

bartam1 (approved):

pkg/webhooks/errors_test.go
pregnor: Sorry for getting here late, being a bit busy.

I think both of you get the right opinion.

Theory: (strict)

If semantically the function is expected to always return the same deterministic order of elements based on the input where there is a relation between the input and the output, I would use `Equal` to signal that the function is expected to comply with that understanding (implicit contract) of returning the elements in predeterminable order - I think of a stable sort for example, but there it is intrinsically part of the semantics of the function, there are less obvious cases like filtering a slice for some elements, in these cases I would suggest using equal.

If the order is not predeterminable or the order of the resulting elements doesn't matter then I would leave it for the test implementor which function is to be used, I wouldn't comment on an `ElementsMatch`, but I personally would favor `Equal` still unless the order of the elements is dynamic in which case I would use `ElementsMatch` - so only use `ElementsMatch` when I absolutely have to, personally. The use-case for this would be when there is a non-stable resorting on elements, for example when there is a weak ordering in a slice - I don't have a concrete example apologies.

Practice: (loose)

In the first case when it is an implicit expectation on ordered elements - input commands being returned in reception order -  I would request changing `ElementsMatch` to `Equal`, in all other cases I would let any of the two fly unless the CI breaks repeatably because of ordering match issues.

Kuvesz (approved):

pkg/webhooks/errors_test.go
pregnor: Also one thing I haven't mentioned.
I really favor the concise output of the `Equal` function in error case so I would prefer that, it is easier to digest by a human afterwards.

pkg/webhooks/kafkacluster_validator.go
panyuenlau: > We should show somehow the formula. I think it is not so clear for the user. Most of the cases the users will not know how is it generated. Also it is hard to find in our documentation. I would like to see the formula with or without the values.

Then we should show how this is calculated in the CRD level, not just the error log message

pkg/webhooks/errors_test.go
mihaialexandrescu: Thank you both for the explanations!

Commit 4419acde4f82e6fcb7bcf0d23f60631367bcceb5 updates the test file to use `require.Equal()`.

panyuenlau (approved):

Kuvesz (dismissed):

pregnor (dismissed):

pkg/webhooks/kafkacluster_validator.go
bartam1: 👎🏼  It means I dont agree with that comment and because you also said that "feel free to merge"  before it has not been completely discussed

panyuenlau: > 👎🏼 It means I dont agree with that comment and because you also said that "feel free to merge" before it has not been completely discussed

"feel free to merge" it's purely from my own perspective, it's 100% ok to disagree, but please provide explanation instead of just a 👎🏼 , what if I give your 👎🏼 every time I don't agree with your opinion?

panyuenlau: On top of that, you also mentioned "LGTM", so your 👎🏼 is conflicting your own message to me

bartam1: I agree we should add the formula into the CRD. That should be in a separate PR because It is belongs to the API module.

panyuenlau: > I agree we should add the formula into the CRD. That should be in a separate PR because It is belongs to the API module.

I don't necessary agree we should do it in a separate PR because we are literally changing the API here in this PR, IMO we should just add a couple of lines here in this PR. But I will leave the decision to you @mihaialexandrescu

mihaialexandrescu: I am trying to understand what's necessary and to come up with a proposal.   

So here goes my first try :  

In `api/v1beta1/kafkacluster_types.go` we would have : 
```
// ExternalListenerConfig defines the external listener config for Kafka
type ExternalListenerConfig struct {
	... ...
	// +kubebuilder:validation:Minimum=0
	// +kubebuilder:validation:Maximum=65535
	// externalStartingPort is added to each broker ID to get the port number that will be used for external access to the broker. 
	// If accessMethod is Nodeport and externalStartingPort is set to 0 then the broker IDs are not added and the nodeport port numbers will be chosen automatically by the K8s Service controller
```
The two `+kubebuilder:validation` markers are already part of this PR.  
By the way, the reason, `0` is accepted is precisely because of that exception which I think comes from [this bit of code](https://github.com/banzaicloud/koperator/blob/master/pkg/resources/nodeportexternalaccess/service.go#L34-L37). Please correct me if I'm wrong on this.

panyuenlau: > 

@mihaialexandrescu it looks solid to me, on top of what you have, I would also add something like this:
```
the choice of broker ID and externalStartingPort must satisfy 0 < broker.ID + externalStartingPort <= 65535
```

mihaialexandrescu: So this ?

```
// ExternalListenerConfig defines the external listener config for Kafka
type ExternalListenerConfig struct {
	... ...
	// +kubebuilder:validation:Minimum=0
	// +kubebuilder:validation:Maximum=65535
	// externalStartingPort is added to each broker ID to get the port number that will be used for external access to the broker. 
         // The choice of broker ID and externalStartingPort must satisfy 0 < brokers.id + externalStartingPort <= 65535
	// If accessMethod is Nodeport and externalStartingPort is set to 0 then the broker IDs are not added and the nodeport port numbers will be chosen automatically by the K8s Service controller
```

panyuenlau: That LGTM

mihaialexandrescu: Got an ok offline from Marton on the phrasing.

Done in d1064eb3c0b18e5d4bf3da9ea67f73b070316445 (`kafkacluster_types.go` and the CRD updates for the description)

panyuenlau (approved):

Kuvesz (approved):

PR #943, opened 2023-03-03
Fix TestCheckKafkaTopicExist

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |

### What's in this PR?

Fix test for that case when the Kafka topic exist on the Kafka cluster and we want to create KafkaTopic Cr for that.

### Why?

The current implementation only checks the last case from the test cases and there is  a wrong test.

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/webhooks/kafkatopic_validator_test.go
+++ b/pkg/webhooks/kafkatopic_validator_test.go
@@ -162,7 +162,7 @@ func TestCheckKafkaTopicExist(t *testing.T) {
 			testName: "topic partition and replication is different and not managedBy koperator",
 			kafkaTopic: v1alpha1.KafkaTopic{
 				ObjectMeta: metav1.ObjectMeta{
-					Annotations: map[string]string{TopicManagedByAnnotationKey: TopicManagedByKoperatorAnnotationValue},
+					Annotations: map[string]string{TopicManagedByAnnotationKey: "other"},
 				},
 				Spec: v1alpha1.KafkaTopicSpec{
 					Name:              "test-topic",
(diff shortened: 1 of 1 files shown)


pkg/webhooks/kafkatopic_validator_test.go
```diff
@@ -193,6 +193,7 @@ func TestCheckKafkaTopicExist(t *testing.T) {
 	}
 
 	for _, testCase := range testCases {
+		testCase := testCase
```
Kuvesz, line 196: Why is this necessary?

Kuvesz (approved):

pkg/webhooks/kafkatopic_validator_test.go
pregnor: Because of how the iteration pointer works in Go together with the capture variables for the lambda function defined on the next line where the iteration pointer is a lambda capture variable and running the tests in a parallel fashion.

With `t.Parallel()` the iteration just fires the subtests but won't be blocked until it finishes, so multiple `t.Run()`s are going to be executed at the same time.
All of those are going to refer to `&testCase.KafkaTopic` for example and when the iteration moves ahead the `testCase` pointer will be updated in place to the new value.
If the iteration is at test case X and executes `t.Run()` then detaches the thread with `t.Parallel()` and the iteration moves ahead to test case Y updating the `testCase` iteration pointer then X's thread executes the the `fieldErrorList, err := kafkaTopicValidator.checkKafka(context.Background(), &testCase.kafkaTopic, cluster)` statement, it will evaluate `&testCase.kafkaTopic` to Y's `testCase.kafkaTopic` value instead of X's.
By declaring another variable inside the iteration to hold the iterator's value, we copy the original content, ensuring the lambda function always calls the iteration-local copy which means using X's values in X's lambda function.
The convention of naming the iteration local variable the same way as the iterator is to ease the developer's work because semantically the iterator changing and the lambda function using the new value of it is merely a Go optimization internal detail the developer doesn't care much about.

pregnor (approved):

pkg/webhooks/kafkatopic_validator_test.go
Kuvesz: I had no idea, thanks a lot for explaining!

PR #944, opened 2023-03-06
Use testify/require in pkg webhooks sentinel errors unit tests

mihaialexandrescu:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | no |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |
| Related tickets |  |

### What's in this PR?

This PR updates the unit test file for the sentinel errors in pkg webhooks (some tests were recently added in https://github.com/banzaicloud/koperator/pull/920) to use `testify`([`testify/require`](https://pkg.go.dev/github.com/stretchr/testify/require)).

This PR brings updates to the syntax and tooling used, not to the logic.  
The very first test is an older one and I felt it could easily lend itself to subtests so I changed that too, but with the goal of keeping the logic intact. The subtest names in that instance are my understanding of the intent of the test (please double check that as well).

### Why?

There were newer [guidelines/advice](https://github.com/banzaicloud/koperator/pull/928#discussion_r1103573247) about using `testify` ([`testify/require`](https://pkg.go.dev/github.com/stretchr/testify/require)) that had eluded me during https://github.com/banzaicloud/koperator/pull/920.

### DIFF
--- a/pkg/webhooks/errors_test.go
+++ b/pkg/webhooks/errors_test.go
@@ -30,21 +30,51 @@ import (
 )
 
 func TestIsAdmissionConnectionError(t *testing.T) {
-	err := apierrors.NewInternalError(errors.Wrap(errors.New("..."), cantConnectErrorMsg))
-
-	if !IsAdmissionCantConnect(err) {
-		t.Error("Expected is connection error to be true, got false")
+	testCases := []struct {
+		testName string
+		err      error
+		want     bool
+	}{
+		{
+			testName: "correct error type and message",
+			err:      apierrors.NewInternalError(errors.Wrap(errors.New("..."), cantConnectErrorMsg)),
+			want:     true,
+		},
+		{
+			testName: "incorrect both error type and message",
+			err:      apierrors.NewServiceUnavailable("some other reason"),
+			want:     false,
+		},
+		{
+			testName: "incorrect error type with correct message",
+			err:      apierrors.NewBadRequest(cantConnectErrorMsg),
+			want:     false,
+		},
 	}
+	for _, tc := range testCases {
+		t.Run(tc.testName, func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


Kuvesz (approved):
Looks good to me, would appreciate someone who knows testify more taking a look.

pregnor:
> Looks good to me, would appreciate someone how knows testify more taking a look.

Doing it now

pkg/webhooks/errors_test.go
```diff
@@ -91,37 +96,34 @@ func TestIsAdmissionCantConnectAPIServer(t *testing.T) {
...
+			require.Equal(t, tc.want, got)
 		})
 	}
 }
 
 func TestIsAdmissionOutOfRangeReplicationFactor(t *testing.T) {
 	kafkaTopic := banzaicloudv1alpha1.KafkaTopic{ObjectMeta: metav1.ObjectMeta{Name: "test-KafkaTopic"}}
 	var fieldErrs field.ErrorList
-	fieldErrs = append(fieldErrs, field.Invalid(field.NewPath("spec").Child("replicationFactor"), "-2", outOfRangeReplicationFactorErrMsg))
+	fieldErrs = append(fieldErrs, field.Invalid(field.NewPath("spec").Child("replicationFactor"), int32(-2), outOfRangeReplicationFactorErrMsg))
```
pregnor, line 108: **Question**: Why were these type changes necessary? Or were they just linter suggestions?

pregnor (approved):
LGTM

pkg/webhooks/errors_test.go
pregnor: No, that is fine, I just wasn't sure whether this was required by compiler, suggested by linter or had another reasoning behind, like what you explained.
Makes sense.

PR #954, opened 2023-03-22
Fix: unused ingress resources cleanup for external listeners

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |

### What's in this PR?

Removing unused ingress resources (nodePort services, istio ingress, envoy ingress) for external listeners

### Why?
Koperator uses envoy as the ingress controller by default. When the users want to expose their Kafka brokers via LoadBalancer, the Koperator deploys these necessary resources: . But it doesn’t seem to remove these resources when they are not needed, for example, the users switch from LoadBalancer to NodePort. Moreover, the brokers are still reachable via the stale loadbalancer service. Same applies when changing from NodePort to LoadBalancer and from istioingress ingresscontroller to envoy ingress controller and vice-versa.

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/resources/envoy/envoy.go
+++ b/pkg/resources/envoy/envoy.go
@@ -96,58 +96,56 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
 					}
 				}
 			}
-		} else {
-			if r.KafkaCluster.Spec.RemoveUnusedIngressResources {
-				// Cleaning up unused envoy resources when ingress controller is not envoy or externalListener access method is not LoadBalancer
-				deletionCounter := 0
-				ctx := context.Background()
-				envoyResourcesGVK := []schema.GroupVersionKind{
-					{
-						Version: corev1.SchemeGroupVersion.Version,
-						Group:   corev1.SchemeGroupVersion.Group,
-						Kind:    reflect.TypeOf(corev1.Service{}).Name(),
-					},
-					{
-						Version: corev1.SchemeGroupVersion.Version,
-						Group:   corev1.SchemeGroupVersion.Group,
-						Kind:    reflect.TypeOf(corev1.ConfigMap{}).Name(),
-					},
-					{
-						Version: appsv1.SchemeGroupVersion.Version,
-						Group:   appsv1.SchemeGroupVersion.Group,
-						Kind:    reflect.TypeOf(appsv1.Deployment{}).Name(),
-					},
-					{
-						Version: policyv1.SchemeGroupVersion.Version,
-						Group:   policyv1.SchemeGroupVersion.Group,
-						Kind:    reflect.TypeOf(policyv1.PodDisruptionBudget{}).Name(),
-					},
-				}
-				var envoyResources unstructured.UnstructuredList
-				for _, gvk := range envoyResourcesGVK {
-					envoyResources.SetGroupVersionKind(gvk)
+		} else if r.KafkaCluster.Spec.RemoveUnusedIngressResources {
+			// Cleaning up unused envoy resources when ingress controller is not envoy or externalListener access method is not LoadBalancer
+			deletionCounter := 0
+			ctx := context.Background()
+			envoyResourcesGVK := []schema.GroupVersionKind{
+				{
+					Version: corev1.SchemeGroupVersion.Version,
+					Group:   corev1.SchemeGroupVersion.Group,
+					Kind:    reflect.TypeOf(corev1.Service{}).Name(),
+				},
+				{
+					Version: corev1.SchemeGroupVersion.Version,
+					Group:   corev1.SchemeGroupVersion.Group,
+					Kind:    reflect.TypeOf(corev1.ConfigMap{}).Name(),
+				},
+				{
+					Version: appsv1.SchemeGroupVersion.Version,
+					Group:   appsv1.SchemeGroupVersion.Group,
+					Kind:    reflect.TypeOf(appsv1.Deployment{}).Name(),
+				},
+				{
+					Version: policyv1.SchemeGroupVersion.Version,
+					Group:   policyv1.SchemeGroupVersion.Group,
+					Kind:    reflect.TypeOf(policyv1.PodDisruptionBudget{}).Name(),
+				},
+			}
+			var envoyResources unstructured.UnstructuredList
+			for _, gvk := range envoyResourcesGVK {
+				envoyResources.SetGroupVersionKind(gvk)
 
-					if err := r.List(ctx, &envoyResources, client.InNamespace(r.KafkaCluster.GetNamespace()),
-						client.MatchingLabels(labelsForEnvoyIngressWithoutEListenerName(r.KafkaCluster.Name))); err != nil {
-						return errors.Wrap(err, "error when getting list of envoy ingress resources for deletion")
-					}
+				if err := r.List(ctx, &envoyResources, client.InNamespace(r.KafkaCluster.GetNamespace()),
+					client.MatchingLabels(labelsForEnvoyIngressWithoutEListenerName(r.KafkaCluster.Name))); err != nil {
+					return errors.Wrap(err, "error when getting list of envoy ingress resources for deletion")
+				}
 
-					for _, removeObject := range envoyResources.Items {
-						if !strings.Contains(removeObject.GetLabels()[util.ExternalListenerLabelNameKey], eListener.Name) ||
-							util.ObjectManagedByClusterRegistry(&removeObject) ||
-							!removeObject.GetDeletionTimestamp().IsZero() {
-							continue
-						}
-						if err := r.Delete(ctx, &removeObject); client.IgnoreNotFound(err) != nil {
-							return errors.Wrap(err, "error when removing envoy ingress resources")
-						}
-						log.V(1).Info(fmt.Sprintf("Deleted envoy ingress '%s' resource '%s' for externalListener '%s'", gvk.Kind, removeObject.GetName(), eListener.Name))
-						deletionCounter++
+				for _, removeObject := range envoyResources.Items {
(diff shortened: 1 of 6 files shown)


pkg/resources/envoy/envoy.go
```diff
@@ -53,41 +71,87 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
...
+					Group:   appsv1.SchemeGroupVersion.Group,
+					Kind:    reflect.TypeOf(appsv1.Deployment{}).Name(),
+				},
+				{
+					Version: policyv1.SchemeGroupVersion.Version,
+					Group:   policyv1.SchemeGroupVersion.Group,
+					Kind:    reflect.TypeOf(policyv1.PodDisruptionBudget{}).Name(),
+				},
+			}
+			var envoyResources unstructured.UnstructuredList
```
Kuvesz, line 129: Do we need to define this outside of the cycle? As far as I can tell it's only used inside.

pkg/resources/envoy/envoy.go
```diff
@@ -53,41 +71,87 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
...
+			for _, gvk := range envoyResourcesGVK {
+				envoyResources.SetGroupVersionKind(gvk)
+
+				if err := r.List(ctx, &envoyResources, client.InNamespace(r.KafkaCluster.GetNamespace()),
+					client.MatchingLabels(labelsForEnvoyIngressWithoutEListenerName(r.KafkaCluster.Name))); err != nil {
+					return errors.Wrap(err, "error happened when getting list of envoy ingress resources for deletion")
```
Kuvesz, line 135:
```suggestion
					return errors.Wrap(err, "error when getting list of envoy ingress resources for deletion")
```

pkg/resources/envoy/envoy.go
```diff
@@ -53,41 +71,87 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
...
 					}
-
-					for _, res := range externalListenerResources {
-						o := res(log, eListener, ingressConfig, name, defaultControllerName)
-						err := k8sutil.Reconcile(log, r.Client, o, r.KafkaCluster)
-						if err != nil {
-							return err
-						}
+					if err := r.Delete(ctx, &removeObject); client.IgnoreNotFound(err) != nil {
+						return errors.Wrap(err, "error happened when removing envoy ingress resources")
```
Kuvesz, line 145:
```suggestion
						return errors.Wrap(err, "error when removing envoy ingress resources")
```

pkg/resources/istioingress/istioingress.go
```diff
@@ -70,39 +83,82 @@ func New(client client.Client, cluster *v1beta1.KafkaCluster) *Reconciler {
...
+				} {
+					o := res(log, eListener, ingressConfig, name, defaultControllerName, istioRevision)
+					err := k8sutil.Reconcile(log, r.Client, o, r.KafkaCluster)
+					if err != nil {
+						return err
+					}
+				}
+			}
+		} else {
+			// Cleaning up unused istio resources when ingress controller is not istioingress or eListener access method is not LoadBalancer
```
Kuvesz, line 118:
```suggestion
			// Cleaning up unused istio resources when ingress controller is not istioingress or the externalListener access method is not LoadBalancer
```

pkg/resources/istioingress/istioingress.go
```diff
@@ -70,39 +83,82 @@ func New(client client.Client, cluster *v1beta1.KafkaCluster) *Reconciler {
...
+					Group:   istioclientv1beta1.SchemeGroupVersion.Group,
+					Kind:    reflect.TypeOf(istioclientv1beta1.Gateway{}).Name(),
+				},
+				{
+					Version: istioclientv1beta1.SchemeGroupVersion.Version,
+					Group:   istioclientv1beta1.SchemeGroupVersion.Group,
+					Kind:    reflect.TypeOf(istioclientv1beta1.VirtualService{}).Name(),
+				},
+			}
+			var istioResources unstructured.UnstructuredList
```
Kuvesz, line 138: Again, can't see this being used anywhere outside the cycle, can't we just define it inside?

pkg/resources/istioingress/istioingress.go
```diff
@@ -70,39 +83,82 @@ func New(client client.Client, cluster *v1beta1.KafkaCluster) *Reconciler {
...
-				istioRevision := istioOperatorApi.NamespacedRevision(
-					strings.ReplaceAll(r.KafkaCluster.Spec.IstioControlPlane.Name, ".", "-"),
-					r.KafkaCluster.Spec.IstioControlPlane.Namespace)
-
-				ingressConfigs, defaultControllerName, err := util.GetIngressConfigs(r.KafkaCluster.Spec, eListener)
-				if err != nil {
-					return err
+				if err := r.List(ctx, &istioResources, client.InNamespace(r.KafkaCluster.GetNamespace()),
+					client.MatchingLabels(labelsForIstioIngressWithoutEListenerName(r.KafkaCluster.Name, ""))); err != nil && !apimeta.IsNoMatchError(err) {
+					return errors.Wrap(err, "error happened when getting list of istio ingress resources for deletion")
```
Kuvesz, line 144:
```suggestion
					return errors.Wrap(err, "error when getting list of istio ingress resources for deletion")
```

pkg/resources/istioingress/istioingress.go
```diff
@@ -70,39 +83,82 @@ func New(client client.Client, cluster *v1beta1.KafkaCluster) *Reconciler {
...
-						r.gateway,
-						r.virtualService,
-					} {
-						o := res(log, eListener, ingressConfig, name, defaultControllerName, istioRevision)
-						err := k8sutil.Reconcile(log, r.Client, o, r.KafkaCluster)
-						if err != nil {
-							return err
-						}
+					if err := r.Delete(ctx, &removeObject); client.IgnoreNotFound(err) != nil {
+						return errors.Wrap(err, "error happened when removing istio ingress resources")
```
Kuvesz, line 154:
```suggestion
						return errors.Wrap(err, "error when removing istio ingress resources")
```

Kuvesz (commented):
Nothing major really from me, though I'm no expert on this.

pkg/resources/envoy/envoy.go
bartam1: I used it outside to be able to re-use it inside iteration but I think it does not matter much.

pkg/resources/envoy/envoy.go
```diff
@@ -15,22 +15,40 @@
...
 	"github.com/banzaicloud/koperator/pkg/k8sutil"
 	"github.com/banzaicloud/koperator/pkg/resources"
 	"github.com/banzaicloud/koperator/pkg/util"
 	envoyutils "github.com/banzaicloud/koperator/pkg/util/envoy"
+)
 
-	"github.com/go-logr/logr"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+const (
+	eListenerLabelNameKey = "eListenerName"
```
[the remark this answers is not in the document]
pregnor: Also wouldn't this be better in `pkg/util/util.go` like the rest of the constants?

pkg/resources/envoy/envoy.go
```diff
@@ -53,41 +67,87 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
...
-						}
+					if err := r.Delete(ctx, &removeObject); client.IgnoreNotFound(err) != nil {
+						return errors.Wrap(err, "error when removing envoy ingress resources")
 					}
+					deletionCounter++
 				}
 			}
+			if deletionCounter > 0 {
+				log.Info(fmt.Sprintf("Removed '%d' resources for envoy ingress", deletionCounter))
+			}
```
pregnor, line 148: **Question**: Wouldn't it be better to put a log message in the place of `deleteCounter++` stating `deleted envoy ingress resource %s` where `%s` is the name of the object?
This way we will see `removed 5 resources for envoy ingress`, not sure if that is more useful than seeing `removed resources for envoy ingress` vs. the more informative listing of removed resources.
I can be convinced we don't need the resource level info, just curious.

Same for the Istio version.

pregnor (approved):

pkg/resources/envoy/envoy.go
bartam1: Then it would be 5 remove message in case envoy.  I think it is a good idea to do this with lower log level like: log.V(1).Info(...) What do you think?

Kuvesz (approved):
I'm still not too keen on some of these EListener shorhands but I wouldn't block the PR on that. Otherwise all looks good to me!

pregnor (approved):

pkg/resources/istioingress/istioingress.go
```diff
@@ -70,38 +82,84 @@ func New(client client.Client, cluster *v1beta1.KafkaCluster) *Reconciler {
...
-						r.virtualService,
-					} {
-						o := res(log, eListener, ingressConfig, name, defaultControllerName, istioRevision)
-						err := k8sutil.Reconcile(log, r.Client, o, r.KafkaCluster)
-						if err != nil {
-							return err
+				}
+			}
+		} else {
+			if r.KafkaCluster.Spec.RemoveUnusedIngressResources {
```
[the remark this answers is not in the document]
bartam1: yep, Thanks!

bartam1:
In the half of the codebase eListener is used and the otherhalf externalListener

bartam1:
> I'm still not too keen on some of these EListener shorhands but I wouldn't block the PR on that. Otherwise all looks good to me!
In the half of the codebase eListener is used and the otherhalf externalListener

Kuvesz (dismissed):

Kuvesz (approved):

pregnor (approved):

PR #958, opened 2023-04-05
Fix out of range panic in resizePVC fn

bartam1:
| Q               | A      |
| --------------- | ------ |
| Bug fix?        | yes |
| New feature?    | no |
| API breaks?     | no |
| Deprecations?   | no |

### What's in this PR?

Fixes resizePVC  function because the storageConfigs could go out of range in some cases

### Why?

https://github.com/banzaicloud/koperator/blob/12af3cd3140cb02d224752c332b9a0072c41e72c/internal/alertmanager/currentalert/process.go#L278 this will be out of range panic when the storageConfig is in the brokerConfigGroup

### Checklist

- [x] Implementation tested

### DIFF
--- a/internal/alertmanager/currentalert/process.go
+++ b/internal/alertmanager/currentalert/process.go
@@ -283,7 +283,6 @@ func resizePvc(log logr.Logger, labels model.LabelSet, annotiations model.LabelS
 					} else {
 						broker.BrokerConfig.StorageConfigs[idx] = *modifiableConfig
 					}
-
 				}
 			}
 			cr.Spec.Brokers[i].BrokerConfig = broker.BrokerConfig
--- a/internal/alertmanager/currentalert/process.go
+++ b/internal/alertmanager/currentalert/process.go
@@ -271,19 +271,19 @@ func resizePvc(log logr.Logger, labels model.LabelSet, annotiations model.LabelS
 
 					modifiableConfig.PvcSpec.Resources.Requests["storage"] = size
 
-					// When the storage is in brokerConfigGroup we don't resize the storage there because in that case
-					// all of those brokers that are using this brokerConfigGroup its storage would be resized.
-					// We add the storage into the broker.BrokerConfig.StorageConfigs in this way only that broker pvc will be resized.
+					// When the storage is in a brokerConfigGroup we don't resize the storage there because in that case
+					// all of the brokers that are using this brokerConfigGroup would have their storages resized.
+					// We add the storage into the broker.BrokerConfig.StorageConfigs in this way only the PVC belonging to that specific broker will be resized.
 					if broker.BrokerConfig == nil {
 						broker.BrokerConfig = &v1beta1.BrokerConfig{}
-					} else {
-						idx := slices.IndexFunc(broker.BrokerConfig.StorageConfigs, func(c v1beta1.StorageConfig) bool { return c.MountPath == pvc.Annotations["mountPath"] })
-						if idx == -1 {
-							broker.BrokerConfig.StorageConfigs = append(broker.BrokerConfig.StorageConfigs, *modifiableConfig)
-						} else {
-							broker.BrokerConfig.StorageConfigs[idx] = *modifiableConfig
-						}
 					}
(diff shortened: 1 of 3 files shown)


internal/alertmanager/currentalert/process.go
```diff
@@ -262,20 +263,26 @@ func resizePvc(log logr.Logger, labels model.LabelSet, annotiations model.LabelS
...
 				modifiableConfig := c.DeepCopy()
 				if modifiableConfig.MountPath == pvc.Annotations["mountPath"] {
 					size := *modifiableConfig.PvcSpec.Resources.Requests.Storage()
 					size.Add(incrementBy)
 
 					modifiableConfig.PvcSpec.Resources.Requests["storage"] = size
 
+					// When the storage is in brokerConfigGroup we don't resize the storage there because in that case
+					// all of those brokers that are using this brokerConfigGroup its storage would be resized.
+					// We add the storage into the broker.BrokerConfig.StorageConfigs in this way only that broker pvc will be resized.
```
Kuvesz, line 276:
```suggestion
					// When the storage is in a brokerConfigGroup we don't resize the storage there because in that case
					// all of the brokers that are using this brokerConfigGroup would have their storages resized.
					// We add the storage into the broker.BrokerConfig.StorageConfigs in this way only the PVC belonging to that specific broker will be resized.
```

Kuvesz (commented):
Not sure if this circumvents a possible issue, where the broker would have the resized storage in it's own config while also still using the brokerConfigGroup thus also having the storage from there. Is that a possibility?

internal/alertmanager/currentalert/process.go
```diff
@@ -262,20 +263,26 @@ func resizePvc(log logr.Logger, labels model.LabelSet, annotiations model.LabelS
...
 					if broker.BrokerConfig == nil {
-						broker.BrokerConfig = &v1beta1.BrokerConfig{
-							StorageConfigs: []v1beta1.StorageConfig{*modifiableConfig},
-						}
+						broker.BrokerConfig = &v1beta1.BrokerConfig{}
```
pregnor, line 278: **Question**: in this case shouldn't this instead be creating a broker specific storage config with the new size on a new PV? 
Like we do with brokerConfig, but mountPath index == -1 case.
Or is that unfeasible in this case?

pregnor: We discussed it offline that in this case we do not want to do that, because we cannot determine what exactly to copy over from the brokerConfigGroup to the broker specific config and what to leave, the merging of the two is not simple enough so there is no todo with this.

pregnor (dismissed):

Kuvesz (approved):

bartam1:
> Not sure if this circumvents a possible issue, where the broker would have the resized storage in it's own config while also still using the brokerConfigGroup thus also having the storage from there. Is that a possibility?
brokerConfig is merged with the brokerConfigGroup (https://github.com/banzaicloud/koperator/blob/519a9c99e451bd25501c2194671bde41b1693f7d/pkg/resources/kafka/kafka.go#L245)
When there is a duplicate it removes it: https://github.com/banzaicloud/koperator/blob/5c781487fb7484cc013a0ff43edf5f4ac09ecb5e/api/v1beta1/kafkacluster_types.go#L1101

pregnor (dismissed):

mihaialexandrescu (dismissed):
lgtm

internal/alertmanager/currentalert/process.go
pregnor: Fixed in [`519a9c9`](https://github.com/banzaicloud/koperator/pull/958/commits/519a9c99e451bd25501c2194671bde41b1693f7d).

pregnor:
I pushed a chart bump, because we want to immediately release a patch version for this fix and we are short on maintainers for the review.
@mihaialexandrescu will review and I will give it an approve as well. 
It's not pretty but we want to move forward with this and have no other options. 
It's analogous to #959 where the bump was reviewed by 2 other people as well.

pregnor (approved):

mihaialexandrescu (approved):

Issue #911, opened 2022-12-15
Supporting external issuers with custom API groups for cert-manager dependency


charleenklang:
**Is your feature request related to a problem? Please describe.**
The koperator supports the upsteam cert-manager project natively and can use it to create and manage all needed certificates for the kafka brokers and users automatically. However, this functionality currently supports only to define the kind and the name of the issuer, which should be used for the certificate management but not the group.

**Describe the solution you'd like to see**
We would like to use an external issuer with custom API groups for the upstream cert-manager dependency when creating a new kafka cluster or kafka user so that the automatic creation of all needed certificates feature of the koperator can be used.

panyuenlau:
Hey @charleenklang, this feature is now supported by #966  and will be available in the next release

panyuenlau:
Hey @charleenklang, apologies - I think I might've misunderstood this issue, I believe the request is to support users to customize the `issuerRef.group` in two different cases (with cert-manager):

1. `KafkaUser`, for example:
```yaml
apiVersion: kafka.banzaicloud.io/v1alpha1
kind: KafkaUser
metadata:
  name: example-kafkauser-with-pki
  namespace: kafka
spec:
  clusterRef:
    name: kafka
  secretName: example-kafkauser-secret
  pkiBackendSpec:
    pkiBackend: "cert-manager"
    issuerRef:
      name: "ca-issuer"
      kind: "ClusterIssuer"
      group: <supported-group-by-cert-manager>
```
This is already addressed by @shubhamcoc in #966 

2. `KafkaCluster`, e.g.:
```
apiVersion: kafka.banzaicloud.io/v1beta1
kind: KafkaCluster
spec:
  listenersConfig:
    internalListeners:
      - type: "ssl"
        name: "ssl"
        containerPort: 29092
        usedForInnerBrokerCommunication: true
        # sslClientAuth defaults to be "required" for two-way SSL authentication, possible values are: "required", "requested", and "none"
        # sslClientAuth: "requested"
      - type: "ssl"
        name: "controller"
        containerPort: 29093
        usedForInnerBrokerCommunication: false
        usedForControllerCommunication: true
        # sslClientAuth defaults to be "required" for two-way SSL authentication, possible values are: "required", "requested", and "none"
        # sslClientAuth: "requested"
    sslSecrets:
      tlsSecretName: "kafka-ca-certs"
      create: true
      issuerRef:
        name: "ca-issuer"
        kind: "ClusterIssuer"
[...]

PR #967, opened 2023-05-08
Adding test cases for kafka user issuerRef group 

shubhamcoc:
## Description

Adding test cases for kafka user issuerRef group and rename of related functions
 
## Type of Change

- [x] New Feature (#911 )

## Checklist


pkg/util/pki/pki_common_test.go
```diff
@@ -154,6 +156,45 @@ func TestBrokerUserForCluster(t *testing.T) {
...
 
+func TestBrokerUserForClusterWithSSLSecret(t *testing.T) {
+	cluster := testCluster(t)
+	cluster.Spec.ListenersConfig.SSLSecrets = &v1beta1.SSLSecrets{
+		IssuerRef: &cmmeta.ObjectReference{
+			Name:  "test",
+			Kind:  "testKind",
+			Group: "testGroup",
+		},
+		PKIBackend: v1beta1.PKIBackendCertManager,
```
panyuenlau, line 167: I think we should also test the case when `PKIBackend` is empty

pkg/util/pki/pki_common_test.go
```diff
@@ -179,6 +220,47 @@ func TestControllerUserForCluster(t *testing.T) {
...
 
+func TestControllerUserForClusterWithSSLSecret(t *testing.T) {
+	cluster := testCluster(t)
+	cluster.Spec.ListenersConfig.SSLSecrets = &v1beta1.SSLSecrets{
+		IssuerRef: &cmmeta.ObjectReference{
+			Name:  "test",
+			Kind:  "testKind",
+			Group: "testGroup",
+		},
+		PKIBackend: v1beta1.PKIBackendCertManager,
```
panyuenlau, line 231: Same as above, we should add a test case when `PKIBackend` is empty

panyuenlau (commented):
LGTM in general, it'd be great add the missing test cases to the unit tests

pkg/util/pki/common.go
```diff
@@ -203,6 +203,42 @@ func BrokerUserForCluster(cluster *v1beta1.KafkaCluster, extListenerStatuses map
...
+			DNSNames:   append(GetInternalDNSNames(cluster), additionalHosts...),
+			IncludeJKS: true,
+			ClusterRef: v1alpha1.ClusterReference{
+				Name:      cluster.Name,
+				Namespace: cluster.Namespace,
+			},
+			PKIBackendSpec: &v1alpha1.PKIBackendSpec{
+				// Not checking IssuerRef for nil as it is checked in caller function kafkapki
+				IssuerRef:  sslConfig.IssuerRef,
+				PKIBackend: pkiBackend,
```
bartam1, line 236: Currently this can be cert-manager only so this modification useless.
We need to extend the k8s api validation here: https://github.com/shubhamcoc/koperator/blob/2c2621944be96aa162024aab581807d4b3f56967/api/v1beta1/kafkacluster_types.go#L491
It need to be "k8s-csr,cert-manager"
Do we really want to this?
If yes then we have to test that case when pkiBackend is k8s-csr.
Currently when the sslConfig.Create is false (https://github.com/banzaicloud/koperator/blob/84c46436b75029b2d4dffa09b80080fb85a0eea3/pkg/pki/certmanagerpki/certmanager_pki.go#L111) then  userProvidedPKI function will be executed.
In this case a new Cert-manager issuer will be created ( https://github.com/banzaicloud/koperator/blob/84c46436b75029b2d4dffa09b80080fb85a0eea3/pkg/pki/certmanagerpki/certmanager_pki.go#L237) which is unnecessary when pkibackend is k8s-csr.

bartam1 (changes requested):

pkg/util/pki/common.go
bartam1: I can do the test after the fixes.

shubhamcoc: I did the changes, but since we are changing in the api pkg, we might need to update it in go.mod to use local api.

bartam1: We need to create a separate PR for the API change and we need a new tag for the API when it is merged to the master branch. This new tag should be used in this PR .

pregnor: You can use Go workspaces for local development before putting up the final PRs for the different modules.

pregnor:
Pushed the [api/v0.27.0](https://github.com/banzaicloud/koperator/releases/tag/api%2Fv0.27.0) tag that contains #970.

pkg/util/pki/common.go
shubhamcoc: updated the go.mod file

pkg/pki/certmanagerpki/certmanager_pki.go
```diff
@@ -121,9 +121,9 @@ func userProvidedIssuerPKI(cluster *v1beta1.KafkaCluster, extListenerStatuses ma
 	// No need to generate self-signed certs and issuers because the issuer is provided by user
 	return []runtime.Object{
 		// Broker "user"
-		pkicommon.BrokerUserForCluster(cluster, extListenerStatuses),
+		pkicommon.BrokerUserForClusterWithPKIBackendSpec(cluster, extListenerStatuses),
```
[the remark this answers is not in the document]
panyuenlau: The two functions that you are modifying here are to create `KafkaUser` CRs for the broker and the operator (these are fake "users" used within the Koperator to ensure its functionality when the Kafka cluster is configured to enable ssl).

When you create a `KafkaUser` CR that is used by an actual Kafka client, the Koperator would first check if the `Certificate` for all the `KafkaUser` (including the those two "fake" `KafkaUser`) exist, and [it only calls `clusterCertificateForUser()` when the corresponding `Certificate` are not present](https://github.com/banzaicloud/koperator/blob/master/pkg/pki/certmanagerpki/certmanager_user.go#L50-L52C46), therefore those `Certificate` for the "fake" `KafkaUser` are not going to be reconciled 

Therefore, the `clusterCertificateForUser()` is not going to help us take care of the `issuerRef` information

pkg/util/pki/common.go
panyuenlau: If my understanding is correct (see [here](https://github.com/banzaicloud/koperator/issues/911#issuecomment-1546228732)), believe the original request is to ask for support to customize the `group` field with `cert-manager`, instead of asking support for `k8s-csr` - which is good to have tho but I don't think supporting `k8s-csr` is what this PR is aiming for

shubhamcoc: Yes @panyuenlau you are right, but if the issuerRef is present with create option as true, do we create kafka-server or kafka-client certificates??

panyuenlau:
> I looked into deeper today and this will be not good. To support the K8S-CSR will be a bigger change and we should not do this in this PR. For k8s-csr support we need to do:
> 
> * Refactor a bit the current SSLSecrets's properties. We will have k8s-csr so it will more complex. (TODO comment: https://github.com/banzaicloud/koperator/blob/84c46436b75029b2d4dffa09b80080fb85a0eea3/api/v1beta1/kafkacluster_types.go#L495
>   )
> * (We should rename the SSLSecret property to SSLConfiguration but this would be breaking change. Probably we should create separate property for cert manager configuration like: SSLConfiguration.CertManager and SSLConfiguration.K8SCSR )
> * We need to introduce the SSLSecret.signerName property for k8s-csr (IssuerRef is only for cert-manager)
> * We need to put the kafkaUser generation here: https://github.com/banzaicloud/koperator/blob/84c46436b75029b2d4dffa09b80080fb85a0eea3/pkg/pki/k8scsrpki/k8scsr_pki.go#L27
>    as for cert-manager (https://github.com/shubhamcoc/koperator/blob/3a6979a6cedf666e05cfcf8be6c629b52dd6b3c1/pkg/pki/certmanagerpki/certmanager_pki.go#L95). This is necessary because when k8s-csr is selected then that reconcilePKI will be called from here: https://github.com/banzaicloud/koperator/blob/84c46436b75029b2d4dffa09b80080fb85a0eea3/pkg/resources/kafka/kafka.go#L231
> * We need to create test cases for k8s-csr use-case.
> * We need to create documentation about this change into the API and into https://github.com/banzaicloud/koperator-docs
> * We need to add KafkaCluster samples for using k8s-csr into https://github.com/banzaicloud/koperator/tree/master/config/samples
> 
> We should support the usage of the Kubernetes CertificateSigningRequests when creating server certificates but we should do this with more caution and planning.
> 
> Dear @shubhamcoc!
> 
> The community would be happy for a PR to support k8s-csr. In case when you creating this PR we will help you.
> 
> Regarding the current PR please:
> 
> * please revert the API to v0.26.0
> * rename userProvidedPKIBackend -> userProvidedIssuerforPKICertManager
> * remove BrokerUserForClusterWithPKIBackendSpec and ControllerUserForClusterWithPKIBackendSpec
> 
> back to your question: "I realized, these changes are not required, as clusterCertificateForUser() func will take care if the issuer is provided in cluster.Spec.ListenersConfig.SSLSecrets.IssuerRef. Can @panyuenlau, @pregnor confirm it?" Answer: It is true they are not required. Here we can see (
> 
> https://github.com/banzaicloud/koperator/blob/84c46436b75029b2d4dffa09b80080fb85a0eea3/pkg/pki/certmanagerpki/certmanager_user.go#L191
> 
> ) koperator uses the issuerRef from the kafkacluster.spec.sslsecrets for the brokerUser (
> https://github.com/banzaicloud/koperator/blob/84c46436b75029b2d4dffa09b80080fb85a0eea3/pkg/util/pki/common.go#L192
> 
> ) because the user.Spec.PKIBackendSpec == nil in this case.

I don't think this PR is aiming to support k8s-csr tho, see my comment [here](https://github.com/banzaicloud/koperator/pull/967#discussion_r1192788758)

shubhamcoc:
@bartam1 I have modified the code as per your comments. Kindly review it.

panyuenlau (approved):

bartam1 (approved):

bartam1 (approved):
LGTM! Thank you!

pregnor:
Mergable is stuck a bit, I'm trying to get it unstuck so I could merge the PR.

PR #973, opened 2023-05-16
Using gomock to mock sigs.k8s.io/controller-runtime/pkg/client If

bartam1:
## Description

We want to use gomock everywhere in the Koperator repository and generate mocks through Makefile.
The sigs.k8s.io/controller-runtime/pkg/client Client interface was generated with mockery and it used testify mock.
When interfaces changes because updates we can re-generate mock easily with the "make mock-generate" command.

## Type of Change

- [x] Refactor

## Checklist


pregnor (approved):
LGTM, didn't examine the mock client line by line, but in general seems sane.

mihaialexandrescu (approved):

mihaialexandrescu:
I wasn't sure if there needs to be a call to the `Finish` method of each new `Controller` we create.

I saw [this](https://github.com/golang/mock/blob/main/gomock/controller.go#L51) saying (just like most examples I find online about gomock):

> Each test should create a new Controller and invoke Finish via defer.

And gives these usage examples : 

```
//   func TestFoo(t *testing.T) {
//     ctrl := gomock.NewController(t)
//     defer ctrl.Finish()
//     // ..
//   }
//
//   func TestBar(t *testing.T) {
//     t.Run("Sub-Test-1", st) {
//       ctrl := gomock.NewController(st)
//       defer ctrl.Finish()
//       // ..
//     })
//     t.Run("Sub-Test-2", st) {
//       ctrl := gomock.NewController(st)
//       defer ctrl.Finish()
//       // ..
//     })
//   })
```

And then the `Finish` method of a `Controller` [indicates](https://github.com/golang/mock/blob/main/gomock/controller.go#L222) : 

> It should be invoked for each Controller.

**But** then the same method later added : 

> // New in go1.14+, if you are passing a *testing.T into NewController function you no
> // longer need to call ctrl.Finish() in your test methods.

So I think we're fine without deferring `Finish`.

Issue #828, opened 2022-06-29
Koperator panics when CruiseControl responds with a progress/async response


amuraru:
**Describe the bug**

CruiseControl has [`webserver.request.maxBlockTimeMs`](https://github.com/linkedin/cruise-control/blob/f32dd14a32391018f4f3b469ee1d20b36023b81d/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/config/constants/WebServerConfig.java#L93) configuration that controls the threshold above which a request is converted to async and a handler returned.
koperator is not able to handle this and panics:
```
{"level":"info","ts":"2022-06-29T05:24:46.382Z","logger":"controller.CruiseControl","msg":"got response for request","reconciler group":"kafka.banzaicloud.io","reconciler kind":"KafkaCluster","name":"pipeline-kafka","namespace":"ns-team-aep-pipeline-kafka-1-int","url":"http://pipeline-kafka-cruisecontrol-svc.ns-team-aep-pipeline-kafka-1-int.svc.cluster.local:8090/kafkacruisecontrol/state?json=true&substates=ANALYZER%2CANOMALY_DETECTOR%2CEXECUTOR%2CMONITOR&verbose=true","status":202}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x2e8 pc=0x16dc126]

goroutine 1368 [running]:
github.com/banzaicloud/koperator/pkg/scale.(*cruiseControlScaler).Status(0xc002a49bf0)
	/workspace/pkg/scale/scale.go:75 +0x186
github.com/banzaicloud/koperator/controllers.(*CruiseControlTaskReconciler).Reconcile(0xc000304b58, {0x2143e20, 0xc002a48c60}, {{{0xc00068f940?, 0x1d65200?}, {0xc000973ac0?, 0x30?}}})
	/workspace/controllers/cruisecontroltask_controller.go:91 +0x28a
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile(0xc00056a000, {0x2143e20, 0xc002a48ba0}, {{{0xc00068f940?, 0x1d65200?}, {0xc000973ac0?, 0x4041f4?}}})
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.0/pkg/internal/controller/controller.go:114 +0x27e
[...]

PR #974, opened 2023-05-16
Handle the case when the request to CC /state endpoint returns the async task instead of the result

aguzovatii:
## Description

This PR attempts to fix https://github.com/banzaicloud/koperator/issues/828.

Currently marked this PR as draft until the change in the API module introduced with the TBA PR is merged.

When the request on CC /state endpoint takes longer than webserver.request.maxBlockTimeMs, then CC will convert the request into an async task and will respond with the taskID instead of the result. The Koperator should handle this case by waiting for CC to complete the task.

The approach is the following:
1. When getting the state of Cruise Control we can have 2 possible scenarios
    1. Cruise Control returns the actual state, in this case (as before) we just return the state and continue the reconciliation
    2. Cruise Control converts the `Status` request into an async task and returns just the `taskID`, then:
        1. We create a new `CruiseControlOperation` to track the progress of the Status request and requeue the reconciliation for later
        2. Next time the reconciliation is started, we notice that there is already a `Status` `CruiseControlOperation` in progress and just request the latest status of this task. If not ready, then we requeue again. This step is repeated until the CC task is completed

Notes:
- that we will have no more than 1 `Status` `CruiseControlOperation` in progress for a `kafkaClusterReference`.
- I added a default timeout of 5 minutes, meaning that if we get the `Status` of CC in more than 5 minutes, then we will consider this outdated and retry

## Type of Change

- [x] Bug Fix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


controllers/cruisecontroloperation_controller.go
```diff
@@ -468,6 +479,121 @@ func (r *CruiseControlOperationReconciler) updateCurrentTasks(ctx context.Contex
...
+		}
+
+		if res.Status == nil {
+			return scale.CruiseControlStatus{}, errors.New("could not get Cruise Control status")
+		}
+
+		return *res.Status, nil
+	}
+
+	res, err := r.scaler.Status(ctx)
```
bartam1, line 536: I think when statusOperation.CurrentTaskState() == banzaiv1beta1.CruiseControlTaskCompletedWithError or the "Cruise Control status operation took too long to finish" then we should run a new  r.scaler.Status(ctx) function

controllers/cruisecontroloperation_controller.go
```diff
@@ -468,6 +479,121 @@ func (r *CruiseControlOperationReconciler) updateCurrentTasks(ctx context.Contex
...
+		if err != nil {
+			return scale.CruiseControlStatus{}, errors.WrapIfWithDetails(err, "could not update the state of Status CruiseControlOperation", "name", statusOperation.GetName(), "namespace", statusOperation.GetNamespace())
+		}
+
+		if statusOperation.CurrentTask().Finished != nil &&
+			statusOperation.CurrentTask().Finished.Time.Sub(statusOperation.CurrentTask().Started.Time) > defaultCruiseControlStatusOperationMaxDuration {
+			return scale.CruiseControlStatus{}, errors.New("the Cruise Control status operation took too long to finish")
+		}
+
+		if res.Status == nil {
```
bartam1, line 529: this will be true when statusOperation.CurrentTaskState() == banzaiv1beta1.CruiseControlTaskCompletedWithError or the task is still active and we dont have result.

I think we should only return here when the task is still active or running so the statusOperation.IsCurrentTaskRunning() so the next time we can check it again.

controllers/cruisecontroloperation_controller.go
```diff
@@ -144,7 +153,7 @@ func (r *CruiseControlOperationReconciler) Reconcile(ctx context.Context, reques
 	}
 
 	// Checking Cruise Control health
-	status, err := r.scaler.Status(ctx)
+	status, err := r.getStatus(ctx, log, kafkaCluster, kafkaClusterRef, ccOperationListClusterWide)
```
bartam1, line 156: we dont need to pass the kafkaClusterRef because it is the reference of the kafkaCluster what is also passed as argument (https://github.com/banzaicloud/koperator/blob/782996952d7af9886b7acfd67ea9f386f116aa6f/controllers/cruisecontroloperation_controller.go#L130)

pkg/scale/scale.go
```diff
@@ -103,51 +103,136 @@ type cruiseControlScaler struct {
...
+	// if the execution takes too much time, then Cruise Control converts the request into async
+	// and returns the taskID
+	if resp.Result == nil {
+		return StatusTaskResult{
+			TaskResult: &Result{
+				TaskID:             resp.TaskID,
+				StartedAt:          resp.Date,
+				ResponseStatusCode: resp.StatusCode,
+				RequestURL:         resp.RequestURL,
+				State:              v1beta1.CruiseControlTaskActive,
```
bartam1, line 124: At this point the State should be v1beta1.CruiseControlTaskInExecution because CC is doing the operation in the background

controllers/cruisecontroloperation_controller.go
```diff
@@ -468,6 +479,121 @@ func (r *CruiseControlOperationReconciler) updateCurrentTasks(ctx context.Contex
...
+	ccOperationListClusterWide banzaiv1alpha1.CruiseControlOperationList,
+) (scale.CruiseControlStatus, error) {
+	var statusOperation *banzaiv1alpha1.CruiseControlOperation
+	for i := range ccOperationListClusterWide.Items {
+		ccOperation := &ccOperationListClusterWide.Items[i]
+		// ignoring the error here to continue processing the operations,
+		// even if the user does not provide a KafkaClusterRef label on the CCOperation then the ref will be an empty object (not nil) and the filter will skip it.
+		ref, _ := kafkaClusterReference(ccOperation)
+		if ref.Name == kafkaClusterRef.Name && ref.Namespace == kafkaClusterRef.Namespace && ccOperation.Status.CurrentTask != nil &&
+			ccOperation.Status.CurrentTask.Operation == banzaiv1alpha1.OperationStatus && ccOperation.IsCurrentTaskRunning() {
```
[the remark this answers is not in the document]
aguzovatii: Makes sense, changed in https://github.com/banzaicloud/koperator/pull/974/commits/2c47d7dbbae5a00806acb152d172b3ad4ada2b42

controllers/cruisecontroloperation_controller.go
aguzovatii: Changed in https://github.com/banzaicloud/koperator/pull/974/commits/2c47d7dbbae5a00806acb152d172b3ad4ada2b42

pkg/scale/scale.go
```diff
@@ -103,51 +103,136 @@ type cruiseControlScaler struct {
...
+
+	status := convert(resp.Result)
+
+	return StatusTaskResult{
+		TaskResult: &Result{
+			TaskID:             resp.TaskID,
+			StartedAt:          resp.Date,
+			ResponseStatusCode: resp.StatusCode,
+			RequestURL:         resp.RequestURL,
+			State:              v1beta1.CruiseControlTaskActive,
```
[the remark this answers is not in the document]
aguzovatii: Changed in https://github.com/banzaicloud/koperator/pull/974/commits/2c47d7dbbae5a00806acb152d172b3ad4ada2b42

pkg/scale/scale.go
aguzovatii: Changed in https://github.com/banzaicloud/koperator/pull/974/commits/2c47d7dbbae5a00806acb152d172b3ad4ada2b42

controllers/cruisecontroloperation_controller.go
```diff
@@ -468,6 +479,121 @@ func (r *CruiseControlOperationReconciler) updateCurrentTasks(ctx context.Contex
...
+		if err := updateResult(log, res.TaskResult, statusOperation, false); err != nil {
+			return scale.CruiseControlStatus{}, errors.WrapIfWithDetails(err, "could not update the state of Status CruiseControlOperation", "name", statusOperation.GetName(), "namespace", statusOperation.GetNamespace())
+		}
+
+		err = r.Status().Update(ctx, statusOperation)
+		if err != nil {
+			return scale.CruiseControlStatus{}, errors.WrapIfWithDetails(err, "could not update the state of Status CruiseControlOperation", "name", statusOperation.GetName(), "namespace", statusOperation.GetNamespace())
+		}
+
+		if statusOperation.CurrentTask().Finished != nil &&
```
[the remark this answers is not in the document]
aguzovatii: > Why we return and dont run a new r.scaler.Status(ctx) function in this case? This status operation cannot be used anymore.

We return an error instead of retrying `r.scaler.Status(ctx)` because we rely on the fact that the reconciliation will be executed again later. Yes, current status operation cannot be used anymore, but notice that every time the `getStatus` returns an error, then the reconciliation will be requeued for later (`10 seconds`):
https://github.com/banzaicloud/koperator/blob/2c47d7dbbae5a00806acb152d172b3ad4ada2b42/controllers/cruisecontroloperation_controller.go#L155-L160

And during the next reconciliation `r.scaler.Status(ctx)` will be called again.

The approach that I tried to use was to fail in case of any error, and rely on the fact that the reconciliation will be executed again later.

But I can change it of course, if you think it will be easier to just call `r.scaler.Status(ctx)` again.

aguzovatii:
Hey @bartam1, thank you for your review!
I addressed a few of the easier suggestions

I will be in vacation for a week, so I will not be responsive during this time.
The rest of suggestions I will be able to address starting with 29th of May.

controllers/cruisecontroloperation_controller.go
```diff
@@ -97,6 +100,12 @@ func (r *CruiseControlOperationReconciler) Reconcile(ctx context.Context, reques
 		return reconciled()
 	}
 
+	// Skip reconciliation for Cruise Control Status operation
+	if currentCCOperation.CurrentTaskOperation() == banzaiv1alpha1.OperationStatus {
```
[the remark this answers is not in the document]
aguzovatii: Not sure why did I need to add this, will double check and remove it if it doesn't make sense

controllers/cruisecontroloperation_controller.go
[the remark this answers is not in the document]
bartam1: I think it is ok to leave it as it is from my side. This is an edge case anyway and rarely happens so waiting 10sec not a problem.

controllers/cruisecontroloperation_controller.go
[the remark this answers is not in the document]
aguzovatii: I verified and can confirm that everything works as expected without this section, so it's safe to remove it.

Basically the only reason why I added it is to avoid logging an error:

```
Koperator does not support this operation...
```
as I thought it can be misleading, instead I considered that it sounds better to have
```
skipping reconciliation for Cruise Control Status operation
```

In conclusion we can remove it if you think that it doesn't bring value, let me know what you think about this

bartam1:
New api tag has been created: https://github.com/banzaicloud/koperator/tree/api/v0.28.1

aguzovatii:
Thank you, @bartam1!
I upgraded `github.com/banzaicloud/koperator/api` to `v0.28.1` and marked this pr as ready for review, let me know if there are other things to address

panyuenlau (approved):
The implementation LGTM.

Just one question (I might be nit-picking here):

I found the naming of this `OperationStatus CruiseControlTaskOperation = "status"` in #975 is a bit misleading: the `CruiseControlTaskOperation` type was created to represent the actual CC operation. So when I saw `OperationStatus CruiseControlTaskOperation = "status"`, I would assume there is an actual CC operation called "status", but I checked the CC wiki and source code and I couldn't really find such an operation.

But this could be due to my limited knowledge on cruise control

aguzovatii:
Hi @panyuenlau, there is the CC GET `state` endpoint that is corresponding with the newly added `OperationStatus CruiseControlTaskOperation = "status"`, check [CC REST API docs](https://github.com/linkedin/cruise-control/wiki/REST-APIs#query-the-state-of-cruise-control).

I used the `status` name instead of `state` just to be consistent with how it was used in Koperator until now (check `CruiseControlScaler`):
https://github.com/banzaicloud/koperator/blob/d00b18cb2eeeb85fd5f4e1075646100d7f7f2931/pkg/scale/types.go#L28

I don't know the reason why it's `status` in Koperator, but we can change everywhere from `status` to `state` to not create confusion in the future if you think it's better.

panyuenlau:
@aguzovatii Thank you for the explanation behind this! 

I think it is fine in terms of this PR. We will refactor the naming in a separate PR if our team decide to go that way.

panyuenlau:
@aguzovatii Thank you for the explanation behind this! 

I think it is fine in terms of this PR. We will refactor the naming in a separate PR if our team decide to go that way.

bartam1 (dismissed):

controllers/tests/mocks/scale.go
```diff
@@ -295,6 +295,21 @@ func (mr *MockCruiseControlScalerMockRecorder) Status(ctx interface{}) *gomock.C
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "Status", reflect.TypeOf((*MockCruiseControlScaler)(nil).Status), ctx)
 }
 
+// StatusTask mocks base method.
```
amuraru, line 298: re-run `make mock-generate` to generate this file

aguzovatii: done, generated the file in 193a716

panyuenlau (approved):

PR #977, opened 2023-05-19
Added Koperator installation end to end test

pregnor:
## Description

Adding basic end to end tests for Koperator installation according to the docs in the `tests/e2e` package.
Also setting some baseline for e2e test reusable components.

## Type of Change

- ~Bug Fix~
- ~New Feature~
- ~Breaking Change~
- ~Refactor~
- ~Documentation~
- [X] Other (please describe): test

## Checklist


panyuenlau (commented):
Very clean implementation 👍🏼 

Quick question: since the E2E test will be pre-commit, shouldn't we use the latest CRD in the workspace and the Koperator image built with the the source code in the workspace?  It's fine to install the latest tagged version for now and we can worry about this part later when we configure GHA to run the test

pregnor:
> Very clean implementation 👍🏼
> 
> Quick question: since the E2E test will be pre-commit, shouldn't we use the latest CRD in the workspace and the Koperator image built with the the source code in the workspace? It's fine to install the latest tagged version for now and we can worry about this part later when we configure GHA to run the test

Yeah, I suppose in case of Koperator we should use the master CRDs, i went into blind copy paste mode when I used the latest tag.

pregnor:
CI is failing on licensei, could be Go workspaces related.

pregnor:
> Very clean implementation 👍🏼
> 
> Quick question: since the E2E test will be pre-commit, shouldn't we use the latest CRD in the workspace and the Koperator image built with the the source code in the workspace? It's fine to install the latest tagged version for now and we can worry about this part later when we configure GHA to run the test

Fixed in [3f98e58](https://github.com/banzaicloud/koperator/pull/977/commits/3f98e5865cd2eefaa3872e86ed1c6f60b646447d)

bartam1:
Currently when the e2e is in the go workspace the make test fails:
`o vet ./...
# sigs.k8s.io/controller-runtime/pkg/cache
../../../../pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/cache/informer_cache.go:144:9: cannot use i.Informer (variable of type "k8s.io/client-go/tools/cache".SharedIndexInformer) as Informer value in return statement: "k8s.io/client-go/tools/cache".SharedIndexInformer does not implement Informer (wrong type for method AddEventHandler)
                have AddEventHandler("k8s.io/client-go/tools/cache".ResourceEventHandler) ("k8s.io/client-go/tools/cache".ResourceEventHandlerRegistration, error)`

bartam1:
Im thinking to rename those files which do not the true test like install_test.go
e.g: k8s_test.go -> k8s_subtest.go. Maybe it would be better if we would create a different directory for those files.
I think when there are more tests it will be hard to recognize the real test from the sub tests.

tests/e2e/koperator_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+			Expect(err).NotTo(HaveOccurred())
+		}
+
+		rawKoperatorCRDs := bytes.Split(rawKoperatorCRDsManifest, []byte("\n---\n"))
+		for rawKoperatorCRDIndex, rawKoperatorCRD := range rawKoperatorCRDs {
+			rawKoperatorCRDs[rawKoperatorCRDIndex] = bytes.Trim(rawKoperatorCRD, "\n-")
+		}
+
+		allowedCRDByteCount := 262144
```
Kuvesz, line 81: Question: Is this an exact number that is required to be this or is it just sort of something that should be fine?

Kuvesz (approved):
Looks good to me, got a couple of questions.

tests/e2e/cert_manager_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	. "github.com/onsi/ginkgo/v2"
+)
+
+// requireApplyingCertManagerCRDs deploys the cert-manager CRDs and checks
+// their existence afterwards.
+func requireApplyingCertManagerCRDs(kubectlOptions *k8s.KubectlOptions, certManagerVersion Version) {
+	It("Applying cert-manager CRDs", func() {
```
bartam1, line 25: Capitalization is not necessary after It, By, When... keywords because the test output will be "When Applying cert-manager"

tests/e2e/install_test.go
```diff
@@ -0,0 +1,38 @@
...
+package e2e
+
+import (
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = When("Installing Koperator", func() {
+	var kubeconfigPath string
```
bartam1, line 24: I think we can use global variable for kubeconfigPath and kubecontextName. In the BeforeSuite there is a check and value (https://github.com/banzaicloud/koperator/blob/3f98e5865cd2eefaa3872e86ed1c6f60b646447d/tests/e2e/koperator_suite_test.go#L39)

bartam1:
I can see some places using "Koperator" and sometimes using "koperator". I think we should use same capitalization everywhere.  e.g:  https://github.com/banzaicloud/koperator/blob/3f98e5865cd2eefaa3872e86ed1c6f60b646447d/tests/e2e/install_test.go#L23 vs https://github.com/banzaicloud/koperator/blob/3f98e5865cd2eefaa3872e86ed1c6f60b646447d/tests/e2e/koperator_test.go#L128

tests/e2e/helm_test.go
```diff
@@ -0,0 +1,167 @@
...
+	Namespace   string            `json:"namespace" yaml:"namespace"`
+	Revision    string            `json:"revision" yaml:"revision"`
+	UpdatedTime string            `json:"updated" yaml:"updated"` // Note: not parsable implicitly.
+	Status      HelmReleaseStatus `json:"status" yaml:"status"`
+	Chart       string            `json:"chart" yaml:"chart"`
+	AppVersion  string            `json:"app_version" yaml:"app_version"`
+}
+
+// installHelmChart deploys a Helm chart to the specified kubectl context and
+// namespace using the specified infos, extra arguments can be any of the helm
```
bartam1, line 52: In the description I can see "extra arguments can be any of the helm CLI install flag arguments" but the arguments list does not contain the extra arguments for CLI install flag.

tests/e2e/koperator_test.go
```diff
@@ -0,0 +1,146 @@
...
+) {
+	It("Installing koperator Helm chart", func() {
+		switch koperatorVersion {
+		case LocalVersion:
+			installHelmChartIfDoesNotExist(kubectlOptions, "", "../../charts/kafka-operator", "", "kafka-operator", nil)
+		default:
+			installHelmChartIfDoesNotExist(
+				kubectlOptions,
+				"https://kubernetes-charts.banzaicloud.com",
+				"kafka-operator",
```
bartam1, line 136: I think we should move out the namespaces for the different components into global variable

bartam1:
I think we need to completely re-structure the current implementation. When there are multiple tests. 
My idea is to create a collector function for install:
```
install
 -InstallKoperator 
    -requireInstallingKoperator
 -installDependecies
    - requireInstallingCertManager
    - requireInstallingZookeeperOperator
    - requireInstallingPrometheusOperator
```
1. Create separate dirs and test suites for every complex test case like this: (future e2e tests)
simplekafkacluster suite:
```
It(simplekafkacluster check)
 - install
 - simplekafkaclustercheck
 - uninstall
```
 install suite:
```
It(install check)
 - install
 - uninstall
```
2. Create one suite for all e2e tests:
e2e_suite
```
beforeSuite
 - create kind cluster

It(simplekafkacluster check)
 - install
 - simplekafkaclustercheck
 - uninstall
It(install check)
 - install
 - uninstall
 ...
```

tests/e2e/helm_test.go
```diff
@@ -0,0 +1,167 @@
...
+// using the specified kubectl context and namespace.
+func listHelmReleases(kubectlOptions *k8s.KubectlOptions) []*HelmRelease {
+	By("Listing Helm releases")
+	output, err := helm.RunHelmCommandAndGetOutputE(
+		GinkgoT(),
+		&helm.Options{
+			KubectlOptions: kubectlOptions,
+		},
+		"list",
+		"--output", "yaml",
```
bartam1, line 139: We can go with json and later json.Unmarshal([]byte(output), &releases)
yaml unmarshaller converts first the output to json and after into struct

tests/e2e/install_test.go
```diff
@@ -0,0 +1,38 @@
...
+		var err error
+		kubeconfigPath, kubecontextName, err = currentEnvK8sContext()
+
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	requireInstallingCertManager(k8s.NewKubectlOptions(kubecontextName, kubeconfigPath, "cert-manager"), "v1.11.0")
+	requireInstallingZookeeperOperator(k8s.NewKubectlOptions(kubecontextName, kubeconfigPath, "zookeeper"), "0.2.14")
+	requireInstallingPrometheusOperator(k8s.NewKubectlOptions(kubecontextName, kubeconfigPath, "prometheus"), "42.0.1")
+	requireInstallingKoperator(k8s.NewKubectlOptions(kubecontextName, kubeconfigPath, "kafka"), LocalVersion)
```
bartam1, line 37: I think we should move the namespaces for the different components into global variable

Makefile
```diff
@@ -98,6 +98,12 @@ test: generate fmt vet manifests bin/setup-envtest
 		-timeout 1h
 	cd properties && go test -coverprofile cover.out -cover -failfast -v -covermode=count ./pkg/... ./internal/...
 
+# Run e2e tests
+test-e2e:
+	go test github.com/banzaicloud/koperator/tests/e2e \
+		-v \
+		-timeout 5m
+
```
bartam1, line 106: Maybe we can add more verbose output for ginkgo:
--ginkgo.show-node-events \
		--ginkgo.trace \
		--ginkgo.v

tests/e2e/cert_manager_test.go
```diff
@@ -0,0 +1,73 @@
...
+package e2e
+
+import (
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	. "github.com/onsi/ginkgo/v2"
+)
+
+// requireApplyingCertManagerCRDs deploys the cert-manager CRDs and checks
+// their existence afterwards.
+func requireApplyingCertManagerCRDs(kubectlOptions *k8s.KubectlOptions, certManagerVersion Version) {
```
bartam1, line 24: Im thinking on to move out to global the kubectlOptions

tests/e2e/cert_manager_test.go
```diff
@@ -0,0 +1,73 @@
...
+			"issuers.cert-manager.io",
+			"orders.acme.cert-manager.io",
+		)
+	})
+}
+
+// requireInstallingCertManager deploys cert-manager CRDs and Helm chart and
+// checks the success of those operations.
+func requireInstallingCertManager(kubectlOptions *k8s.KubectlOptions, certManagerVersion Version) {
+	When("Installing cert-manager", func() {
```
bartam1, line 48: In this case we need to use Ordered containers so the order of the "It"s will be fixed.
When("Installing cert-manager", Ordered, func() {

tests/e2e/cert_manager_test.go
```diff
@@ -0,0 +1,73 @@
...
+			kubectlOptions,
+			"https://charts.jetstack.io",
+			"cert-manager",
+			certManagerVersion,
+			"cert-manager",
+			nil,
+		)
+
+		By("Verifying cert-manager pods")
+		requireRunningPods(kubectlOptions, "app.kubernetes.io/name", "cert-manager", "cainjector", "webhook")
```
bartam1, line 71: Maybe this is not necessary because helm checking the deployed pods health before move forward.

tests/e2e/helm_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+	if isInstalled {
+		By(fmt.Sprintf(
+			"Skipping the installation of %s, because the Helm release is already present: %+v",
+			helmReleaseName, helmRelease,
+		))
+
+		return
+	}
+
```
bartam1, line 118: Here we should add this:
```
By(
		fmt.Sprintf(
			"Installing Helm chart %s from %s with version %s by name %s",
			helmChartNameOrLocalPath, helmRepository, helmChartVersion, helmReleaseName,
		),
	)
```

tests/e2e/koperator_test.go
```diff
@@ -0,0 +1,146 @@
...
+				},
+				"../../charts/kafka-operator",
+				"dummy",
+				[]string{"templates/crds.yaml"},
+			))
+		default:
+			httpClient := new(http.Client)
+			httpClient.Timeout = 5 * time.Second
+
+			response, err := httpClient.Get("https://github.com/banzaicloud/koperator/releases/download/" + koperatorVersion + "/kafka-operator.crds.yaml")
```
bartam1, line 64: It would be better to use here this path: https://raw.githubusercontent.com/banzaicloud/koperator/$branchnameOrTag/config/base/crds/kafka.banzaicloud.io_kafkaclusters.yaml etc...
In this way we can run test for different branches

tests/e2e/k8s_test.go
```diff
@@ -0,0 +1,174 @@
...
+	for _, crdName := range crdNames {
+		crdFullNames = append(crdFullNames, "customresourcedefinition.apiextensions.k8s.io/"+crdName)
+	}
+
+	Expect(crds).To(ContainElements(crdFullNames))
+}
+
+// requireRunningPods checks whether the specified pod names are existing in the
+// namespace and have a running status.
+func requireRunningPods(kubectlOptions *k8s.KubectlOptions, matchingLabelKey string, podNames ...string) {
```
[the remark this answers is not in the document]
Kuvesz: Also this function doesn't really work in some cases, for example if I'm trying to check a kafka cluster's pods that are in the kafka namespace this will fail because the number of pods in the kafka namespace don't match the length of podNames (because of the kafka-operator pod being in that namespace). Also our kafka cluster pods don't really have any label with the pod name itself so that's another thing that won't really work as expected.
Another thing: Since we're checking a label I don't really feel like we're doing what we say we're doing in the function name and description either.

tests/e2e/cert_manager_test.go
Kuvesz: I agree, though I'd also add `ContinueOnFailure` because of this:

> When a spec in an Ordered container fails all subsequent specs are skipped. Ginkgo will then run any AfterAll node closures to clean up after the specs.
> 
> You can override this behavior by decorating an Ordered container with ContinueOnFailure.

tests/e2e/koperator_test.go
```diff
@@ -0,0 +1,146 @@
...
+			"kafkaclusters.kafka.banzaicloud.io",
+			"kafkatopics.kafka.banzaicloud.io",
+			"kafkausers.kafka.banzaicloud.io",
+		)
+	})
+}
+
+// requireInstallingKoperator deploys koperator CRDs and Helm chart and checks
+// the success of those operations.
+func requireInstallingKoperator(kubectlOptions *k8s.KubectlOptions, koperatorVersion Version) {
```
bartam1, line 115: I suggest to not use pointer type for kubectlOptions. 
I found that in some cases I needed to change the kubectlOptions.namespace value to "" to avoid an error or warning. I also had to change back the original value because of the pointer type.

tests/e2e/cert_manager_test.go
bartam1: I think we should use Fast fail at installation, kind cluster tests, kafkacluster deploy test, zookeeper cluster deploy test and at uninstallations (https://onsi.github.io/ginkgo/#mental-model-how-ginkgo-handles-failure)
Otherwise I agree to use the ContinueOnFailure in other cases.

tests/e2e/zookeeper_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+// requireInstallingZookeeperOperator deploys zookeeper-operator Helm chart and
+// checks the success of that operation.
+func requireInstallingZookeeperOperator(kubectlOptions *k8s.KubectlOptions, certManagerVersion string) {
+	When("Installing zookeeper-operator", func() {
+		requireInstallingZookeeperOperatorHelmChartIfDoesNotExist(kubectlOptions, certManagerVersion)
+	})
+}
+
+// requireDeployingCertManagerHelmChart checks the existence of the cert-manager
```
bartam1, line 31: The function description here about requireDeployingCertManagerHelmChart

tests/e2e/koperator_test.go
pregnor: This is currently an exact limitation set by kubectl.

tests/e2e/helm_test.go
```diff
@@ -0,0 +1,167 @@
...
+		helmReleaseName,
+	)
+}
+
+// installHelmChart checks whether the specified named Helm release exists in
+// the provided kubectl context and namespace, logs it if it does and returns or
+// alternatively deploys the Helm chart to the specified kubectl context and
+// namespace using the specified infos, extra arguments can be any of the helm
+// CLI install flag arguments, flag keys and values must be provided separately.
+func installHelmChartIfDoesNotExist(
```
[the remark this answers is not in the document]
pregnor: I have mixed feelings about doing this, I do understand how it would simplify development, but if we would need a non-idempotent Helm install we would need these as separate functions, but that's not the case right now, I will think about merging them.

tests/e2e/k8s_test.go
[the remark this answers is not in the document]
pregnor: While I agree we can use higher level functions to check more of a healthy state, I would definitely separate checking 
- resources we recently interacted with and 
- resources which should remain idle/functioning
to make sure in case of a failure we have a quick understanding of whether the resources failed the healthiness check which we manipulated or something unrelated broke so far unexpectedly.

The description definitely requires an update.
I take suggestions for a better function name, I didn't want to involve health/ready keywords unless we check the readiness/liveness endpoints of the pods.

tests/e2e/install_test.go
pregnor: I opted not to use a global variable for those, because we would want to pass it in as a parameter once the multiple K8s cluster support is done.
Imagine a scenario where we want to run this scenario on AKS, AWS, GKE, RHOS, we would run the same scenario with different clusters as parameters.
I didn't introduce these as parameters yet, because I didn't have time to check what would be the best to pass those parameters into such Ginkgo scenarios yet.

Also globals are anti-pattern, because they are not properly access-bound so I'm staying away from those when it's possible without significant costs.

tests/e2e/helm_test.go
pregnor: I just picked a random format I could parse into an object, I prefer YAML, because in printed format it's more readable to a human eye than JSON for debugging purposes.
For tests the performance difference is immeasurable - the test steps will take significantly more time than the helm list command.
But I can convert it to JSON.

Makefile
pregnor: Why do you think so?
In case of failure we can rerun the automated tests with more verbose output, but I found the manually added logging adequate enough, especially from the Ginkgo side, especially if we want to order the test specifications (which I'm debating in myself still which would be the best).

tests/e2e/cert_manager_test.go
pregnor: https://onsi.github.io/ginkgo/#spec-subjects-it
> Every Ginkgo spec has exactly one subject node. You can add a single spec to a suite by adding a new subject node using It(<description>, <closure>). Here's a spec to validate that we can extract the author's last name from a Book model:

 https://onsi.github.io/ginkgo/#organizing-specs-with-container-nodes
> Ginkgo allows you to hierarchically organize the specs in your suite using container nodes. Ginkgo provides three synonymous nouns for creating container nodes: Describe, Context, and When. These three are functionally identical and are provided to help the spec narrative flow. You usually Describe different capabilities of your code and explore the behavior of each capability across different Contexts.

https://onsi.github.io/ginkgo/#mental-model-how-ginkgo-traverses-the-spec-hierarchy
> When running a spec Ginkgo invokes the setup and subject nodes closures in the correct order and tracks any failed assertions.

https://onsi.github.io/ginkgo/#spec-randomization
> Ginkgo's default behavior is to only randomize the order of top-level containers -- the specs within those containers continue to run in the order in which they are specified in the test files. This is helpful when developing specs as it mitigates the cognitive overload of having specs within a container continuously change the order in which they run during a debugging session.

By default Ginkgo behavior the order of the Its will always be fixed according to the Ginkgo docs - because they are never top level containers, they are subject nodes under at least one container node. Only Whens/Contexts/Describes are parallelized.
`It` is considered a subject node while `When` is considered a container node.
See the example in my 3rd quote, I think we are analogous to that.

That's why I designed the solution so the highest level test specifications such as "external producer-consumer" test are going to be the top level containers which are going to be randomized.
1. which eliminates the need to specify ordered in lower level whens (that would be redundant/noisy, but we can opt for it if we want to) and also
2. which is supposed to be fine because with slow-fail we don't care if the failure order is InstallFailed SimpleKafkaClusterDeployFailed ExternalProducerConsumerFailed or the other way around, all will be run and logged anyway and we need to check which is the first to fix.

We can still opt for ordered run of these if we want to make sure to have these in an orderly fashion in the log, but IMO these should be run in parallel anyway on different clusters - to quicken up the execution - which will cause the order to be somewhat messed up anyway, but slowfail will help us be able to determine the smallest test specification in featureset that failed to start investigating upon.

(I do agree if we introduce the order then we need to introduce the `ContinueOnFailure` as well.)

tests/e2e/helm_test.go
bartam1: I agree with you that for debugging purposes we should use yaml.

Makefile
bartam1: I agree with you, maybe it is unecessary but what is the advantage to not run with more verbose output? 
Maybe it is better to not re-run a 15 min test again to get more verbose output when CI pipeline fail. What if the error does not happen every time?

tests/e2e/helm_test.go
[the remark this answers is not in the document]
pregnor: Fixed in cfedc13.

tests/e2e/koperator_test.go
pregnor: Fixed in fa8387b.

pregnor:
> I can see some places using "Koperator" and sometimes using "koperator". I think we should use same capitalization everywhere. e.g:
> 
> https://github.com/banzaicloud/koperator/blob/3f98e5865cd2eefaa3872e86ed1c6f60b646447d/tests/e2e/install_test.go#L23
> 
> vs
> https://github.com/banzaicloud/koperator/blob/3f98e5865cd2eefaa3872e86ed1c6f60b646447d/tests/e2e/koperator_test.go#L128

Fixed these typos in 0eeddfa.

tests/e2e/helm_test.go
pregnor: Fixed in 5bfcc05.

tests/e2e/helm_test.go
pregnor: Fixed in a8def0d.

pregnor: Fixed the fix in 1f98ba9.

tests/e2e/koperator_test.go
bartam1: Maybe it would be be safer to use not pointer type for this. What do you think?

Makefile
pregnor: Fixed in 2ca83fe.

tests/e2e/k8s_test.go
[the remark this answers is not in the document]
bartam1: Im thinking on to only check the readiness of the pod like here: https://github.com/banzaicloud/koperator/blob/f7fe06f20be0d303c28575169aaaa0ef1f0d1888/tests/e2e/kafka_prod_consum_test.go#L51
When it fails we put the kubectl describe pod output into the error

tests/e2e/install_test.go
bartam1: I think it is OK to keep it as it is. Later probably we need to add k8s health test like here: https://github.com/banzaicloud/koperator/blob/f7fe06f20be0d303c28575169aaaa0ef1f0d1888/tests/e2e/koperator_suite_test.go#L43 also into tests to check those other clusters health

tests/e2e/cert_manager_test.go
bartam1: I also checked and I missed something. We dont need to use ordered container in this case.
We need to use only here (https://github.com/banzaicloud/koperator/blob/f7fe06f20be0d303c28575169aaaa0ef1f0d1888/tests/e2e/install_test.go#L23) because to level containers here will be randomized and uninstallation will happen sooner then installation :)

pregnor: Hmmm, in that case I'm the one who misunderstood something - I thought `It`s should always be ordered inside a `When` and only top level `When`s should be randomized, but I believe your example on Koperator - and then the same should hold true for cert-manager (CRD and Helm chart) and then I would say we should do the same for all (prometheus and zookeeper included).
I will get this fixed.

tests/e2e/cert_manager_test.go
pregnor: Fixed in eb36698.

I decided to remove all these pod checks after Helm installs, we can add more sophisticated post-Helm-install checks if we want to but it's not required for now.

tests/e2e/zookeeper_test.go
pregnor: Fixed in 5916442. (I accidentally bundled it with other changes, apologies)

tests/e2e/koperator_test.go
pregnor: Fixed in ef59837.

This does make it a bit more cumbersome to use kubectlOptions, but I agree the trade-off may be desirable.

tests/e2e/cert_manager_test.go
pregnor: Fixed in 64f5aac.

TBH I don't fully understand how Ginkgo work and don't have the time to figure out so I added the Ordered decorator to each `When` to be sure, that's a safer pattern to follow for now, if we find a better solution later we can easily search & replace.

pregnor:
> I think we should do something like this structure to avoid unecessary install uninstall.
> 
> ```
> beforeSuite
>  - create kind cluster
>  - install
> afterSuite
> -uninstall
> -destroy kind cluster
> 
> justBefore
> - deploy zookeeper cluster
> justAfter:
> - remove zookeeper cluster
> - 
> Describe(simplekafkacluster check)
>   JustBefore() {
>      - deploy
>     }
>    JustAfter() {
>      - removeKafkaCluster
>  - simplekafkaclustercheck
>  
> Describe(simplekafkacluster-SSL check)
>   JustBefore() {
>      - deploy
>     }
>    JustAfter() {
>      - removeKafkaCluster
>  - simplekafkaclustercheck-SSL
> ```
> 
> Probably we want to deploy only one kind cluster for all of the tests.
> 
> "By default, Ginkgo will randomize the order in which the specs in a suite run. This is done intentionally. By randomizing specs, Ginkgo can help suss out spec pollution - accidental dependencies between specs - throughout a suite's development.
> 
> Ginkgo's default behavior is to only randomize the order of top-level containers -- the specs within those containers continue to run in the order in which they are specified in the test files. This is helpful when developing specs as it mitigates the cognitive overload of having specs within a container continuously change the order in which they run during a debugging session."
> 
> We need to use the Ordered keyword for top level containers to avoid the Specs randomization. Most of the time the order when there are different Specs on same level is important for us.

Fixed this in 64f5aac, see the related thread for more details.

tests/e2e/k8s_test.go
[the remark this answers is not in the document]
pregnor: I decided to remove this function, because it was only used in the Helm post install pod readiness check and that was removed in https://github.com/banzaicloud/koperator/commit/eb366989392cb596456685145f47b9612bc54a2a. see https://github.com/banzaicloud/koperator/pull/977#discussion_r1208007607 for more details.

tests/e2e/koperator_test.go
```diff
@@ -0,0 +1,146 @@
...
+	When("Installing koperator", func() {
+		requireApplyingKoperatorCRDs(kubectlOptions, koperatorVersion)
```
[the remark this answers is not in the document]
pregnor: So I will do 2 things:
1. before CRD install, only install CRDs when missing - using the Helm version for CRD version as well
2. before Helm install, check the CRD apiVersion to match the one in the Helm chart and fail the test if they mismatch.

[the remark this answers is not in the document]
pregnor: Fixed in c08f3d2 (with a bit of a small refactor).

pregnor:
> Im thinking on to rename those files which are not the true test like install_test.go e.g: k8s_test.go -> k8s_subtest.go.

Fixed in a1c5fe9.

> Currently when the e2e is in the go workspace the make test fails:
> 
> ```
> # sigs.k8s.io/controller-runtime/pkg/cache
> ../../../../pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/cache/informer_cache.go:144:9: cannot use i.Informer (variable of type "k8s.io/client-go/tools/cache".SharedIndexInformer) as Informer value in return statement: "k8s.io/client-go/tools/cache".SharedIndexInformer does not implement Informer (wrong type for method AddEventHandler)
>                 have AddEventHandler("k8s.io/client-go/tools/cache".ResourceEventHandler) ("k8s.io/client-go/tools/cache".ResourceEventHandlerRegistration, error)```
> 
> Should we need to  use here build tags in GoLang to separate integration and unit test code ( // +build e2e )
> ```

I think this was caused by the _test suffix, so with the renamings this should go away, but let me know if it persists and we can fix that later - I don't see what else could cause this.

pregnor:
> I think we should do something like this structure to avoid unecessary install uninstall.
> 
> ```
> beforeSuite
>  - create kind cluster
>  - install
> afterSuite
> -uninstall
> -destroy kind cluster
> 
> justBefore
> - deploy zookeeper cluster
> justAfter:
> - remove zookeeper cluster
> - 
> Describe(simplekafkacluster check)
>   JustBefore() {
>      - deploy
>     }
>    JustAfter() {
>      - removeKafkaCluster
>  - simplekafkaclustercheck
>  
> Describe(simplekafkacluster-SSL check)
>   JustBefore() {
>      - deploy
>     }
>    JustAfter() {
>      - removeKafkaCluster
>  - simplekafkaclustercheck-SSL
> ```
> 
> Probably we want to deploy only one kind cluster for all of the tests.
> 
> "By default, Ginkgo will randomize the order in which the specs in a suite run. This is done intentionally. By randomizing specs, Ginkgo can help suss out spec pollution - accidental dependencies between specs - throughout a suite's development.
> 
> Ginkgo's default behavior is to only randomize the order of top-level containers -- the specs within those containers continue to run in the order in which they are specified in the test files. This is helpful when developing specs as it mitigates the cognitive overload of having specs within a container continuously change the order in which they run during a debugging session."
> 
> We need to use the Ordered keyword for top level containers to avoid the Specs randomization. Most of the time the order when there are different Specs on same level is important for us.

Fixed in d241978 according to what we discussed.

tests/e2e/koperator_test.go
pregnor: Fixed in b4f5d00.

tests/e2e/helm.go
```diff
@@ -0,0 +1,294 @@
...
+// namespace using the specified infos, extra arguments can be any of the helm
+// CLI install flag arguments, flag keys and values must be provided separately.
+func (helmDescriptor *helmDescriptor) installHelmChart(kubectlOptions k8s.KubectlOptions) error {
+	if helmDescriptor == nil {
+		return errors.Errorf("invalid nil Helm descriptor")
+	}
+
+	kubectlOptions.Namespace = helmDescriptor.Namespace
+
+	if !helmDescriptor.IsRemote() { // Note: local chart with directory path in helmDescriptor.Repository.
```
Kuvesz, line 86: Not crucial in any way:
```suggestion
	if !helmDescriptor.IsRemote() { // Note: local chart with directory path in helmDescriptor.Repository
```

Kuvesz (dismissed):
I think this looks good! Thanks for the work!

panyuenlau (approved):
LGTM. Just one minor thing that I've noticed, it might not even worth another commit

tests/e2e/test_install.go
```diff
@@ -0,0 +1,78 @@
...
+	kubectlOptions := kubectlOptions(kubecontextName, kubeconfigPath, "")
+
+	When("Installing cert-manager", func() {
+		It("Installing cert-manager CRDs", func() {
+			certManagerCRDPath, err := certManagerHelmDescriptor.crdPath()
+			Expect(err).NotTo(HaveOccurred())
+
+			err = installCRDs(kubectlOptions, certManagerCRDPath)
+			Expect(err).NotTo(HaveOccurred())
+		})
```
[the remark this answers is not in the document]
pregnor: I think it is, I will try this.

tests/e2e/helm.go
pregnor: Are you asking to remove the sentence terminating dot?

My reasoning for adding it was the [comment sentences](https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences) Go idiom.

I can do so easily, just want to clarify what we aim for - so far I tried to uphold that idiom.

Kuvesz: Ah I see, then just ignore this, didn't know about that.

tests/e2e/k8s.go
```diff
@@ -0,0 +1,395 @@
...
+		"create",
+		fmt.Sprintf("--validate=%t", shouldBeValidated),
+		"--filename", manifestPath,
+	)
+}
+
+// createOrReplaceK8sResourcesFromManifest creates non-existent Kubernetes
+// resources or replaces existing ones from the specified manifest to the
+// provided kubectl context and namespace.
+func createOrReplaceK8sResourcesFromManifest(
```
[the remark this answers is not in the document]
pregnor: Fixed in 0318535.

tests/e2e/test_install.go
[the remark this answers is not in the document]
pregnor: Fixed in 87b788f.
Left the now unused stuff in for now because I feel like we might need it soon and don't want to reimplement, but if we finish the next batch of test cases and don't need it then we can throw those out.

panyuenlau (approved):

Kuvesz (approved):

Kuvesz (dismissed):

Kuvesz (approved):

bartam1 (approved):

Issue #964, opened 2023-04-26
Configure the kafka certificates with longer period


chrodrigues:
## Description

I have the following scenario:

1 - Producers (logstash) are connecting on Kafka cluster using SSL.

2 - When the producer certificate expires it can't connect to the cluster due to this error:

% ERROR: Local: SSL error: ssl://kafka.mydomain.com:29092/bootstrap: SSL handshake failed: ../ssl/statem/statem_clnt.c:1913: error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed:  (after 368ms in state CONNECT)

3 - All times that the certificate is rotated I have to renew this one on each producer.

## Expected Behavior
I would like Kafka to generate certificates with a longer validity period.

## Actual Behavior
Kafka is generating the certificates with 3 months of validity period

## Affected Version

3.1-IV0

## Steps to Reproduce

Wait for expiration of certificate
Wait for renewal of certificate
Client connexion issue should appears

Thank you!

panyuenlau:
Hey @chrodrigues - I am seeing two different requests from this issue:
- certificate rotation (this is a duplicate of #827)
- customized validity of the generated certificates

Please correct me if I am mistaken, otherwise I think it's better to change the issue title to something like "Allow users to customize validity of generated certificates"

panyuenlau:
Hey @fquinino, the Koperator is not yet capable of allowing users to configure the validity of the certificates, the 90 days validity is the default certificate validity from cert-manager, see the `certificate.spec.duration` in here https://cert-manager.io/docs/usage/certificate/#creating-certificate-resources

bartam1:
I can confirm what @panyuenlau said: The default expiration for all certificates is 90 days. The default time before expiration when the Certificate manager service renews the certificates is 30 days.
We will fix this soon.

chrodrigues:
> Hey @fquinino, the Koperator is not yet capable of allowing users to configure the validity of the certificates, the 90 days validity is the default certificate validity from cert-manager, see the `certificate.spec.duration` in here https://cert-manager.io/docs/usage/certificate/#creating-certificate-resources

Hi @panyuenlau, thank you for your help. It's worked for me.

panyuenlau:
@shubhamcoc is willing to pick this up

shubhamcoc:
@panyuenlau yes i can pick this up

panyuenlau:
Closing this as @shubhamcoc's commits should cover the ask here

PR #980, opened 2023-05-24
Adding certificate validity duration in Kafka user

shubhamcoc:
## Description

This MR is to add duration in Kafka user spec. It allows user to specify validity time for certificate created by cert manager.  
It helps the user to create user certificate for longer duration than the default value of cert manager. #964 

## Type of Change

- [x] New Feature

## Checklist

### DIFF
--- a/controllers/cruisecontroloperation_controller.go
+++ b/controllers/cruisecontroloperation_controller.go
@@ -267,6 +267,8 @@ func (r *CruiseControlOperationReconciler) executeOperation(ctx context.Context,
 		cruseControlTaskResult, err = r.scaler.RebalanceWithParams(ctx, ccOperationExecution.CurrentTaskParameters())
 	case banzaiv1alpha1.OperationStopExecution:
 		cruseControlTaskResult, err = r.scaler.StopExecution(ctx)
+	case banzaiv1alpha1.OperationStatus:
+		err = errors.NewWithDetails("Cruise Control operation not supported", "name", ccOperationExecution.GetName(), "namespace", ccOperationExecution.GetNamespace(), "operation", ccOperationExecution.CurrentTaskOperation(), "parameters", ccOperationExecution.CurrentTaskParameters())
 	default:
 		err = errors.NewWithDetails("Cruise Control operation not supported", "name", ccOperationExecution.GetName(), "namespace", ccOperationExecution.GetNamespace(), "operation", ccOperationExecution.CurrentTaskOperation(), "parameters", ccOperationExecution.CurrentTaskParameters())
 	}
--- a/pkg/pki/certmanagerpki/certmanager_user_test.go
+++ b/pkg/pki/certmanagerpki/certmanager_user_test.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/banzaicloud/koperator/api/v1alpha1"
 	"github.com/banzaicloud/koperator/pkg/errorfactory"
+	"github.com/banzaicloud/koperator/pkg/util"
 	certutil "github.com/banzaicloud/koperator/pkg/util/cert"
 )
 
@@ -98,6 +99,14 @@ func TestReconcileUserCertificate(t *testing.T) {
 			Group: "testGroup",
 		},
 	}
+
+	if _, err := manager.ReconcileUserCertificate(ctx, user, scheme.Scheme, clusterDomain); err != nil {
+		t.Error("Expected no error, got:", err)
+	}
+
+	// Test cert duration
+	user = newMockUser()
+	user.Spec.ExpirationSeconds = util.Int32Pointer(7200)
 
 	if _, err := manager.ReconcileUserCertificate(ctx, user, scheme.Scheme, clusterDomain); err != nil {
 		t.Error("Expected no error, got:", err)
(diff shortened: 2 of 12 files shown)


bartam1 (changes requested):
We also have to use the validation duration for the k8s CSR object:
https://github.com/banzaicloud/koperator/blob/a64633a24935763cab1714e909e669323392ffe8/pkg/pki/k8scsrpki/k8scsr_user.go#L253

shubhamcoc:
> We also have to use the validation duration for the k8s CSR object:
> 
> https://github.com/banzaicloud/koperator/blob/a64633a24935763cab1714e909e669323392ffe8/pkg/pki/k8scsrpki/k8scsr_user.go#L253

Yes one validation is required as we have 2 different combination available. One for cert-manager with 1 hour as minimum duration and another with k8s CSR object with 10 mins as minimum duration.

shubhamcoc:
> We also have to use the validation duration for the k8s CSR object:
> 
> https://github.com/banzaicloud/koperator/blob/a64633a24935763cab1714e909e669323392ffe8/pkg/pki/k8scsrpki/k8scsr_user.go#L253

panyuenlau (commented):
Can we also extend the existing tests to include the change in this PR?

https://github.com/banzaicloud/koperator/blob/ffb1fabbf3c44e6069b6dc51d13e55a82f11f986/pkg/pki/certmanagerpki/certmanager_user_test.go#L66

https://github.com/banzaicloud/koperator/blob/ffb1fabbf3c44e6069b6dc51d13e55a82f11f986/pkg/pki/k8scsrpki/k8scsr_user_test.go#L91

controllers/cruisecontroloperation_controller.go
```diff
@@ -267,6 +267,8 @@ func (r *CruiseControlOperationReconciler) executeOperation(ctx context.Context,
 		cruseControlTaskResult, err = r.scaler.RebalanceWithParams(ctx, ccOperationExecution.CurrentTaskParameters())
 	case banzaiv1alpha1.OperationStopExecution:
 		cruseControlTaskResult, err = r.scaler.StopExecution(ctx)
+	case banzaiv1alpha1.OperationStatus:
+		err = errors.NewWithDetails("Cruise Control operation not supported", "name", ccOperationExecution.GetName(), "namespace", ccOperationExecution.GetNamespace(), "operation", ccOperationExecution.CurrentTaskOperation(), "parameters", ccOperationExecution.CurrentTaskParameters())
```
panyuenlau, line 271: Note: these two lines are not related to the PR and they are copied from https://github.com/banzaicloud/koperator/pull/974 to make linter not complain the missing case in the switch statement.

shubhamcoc:
> Can we also extend the existing tests to include the change in this PR?
> 
> https://github.com/banzaicloud/koperator/blob/ffb1fabbf3c44e6069b6dc51d13e55a82f11f986/pkg/pki/certmanagerpki/certmanager_user_test.go#L66
> 
> https://github.com/banzaicloud/koperator/blob/ffb1fabbf3c44e6069b6dc51d13e55a82f11f986/pkg/pki/k8scsrpki/k8scsr_user_test.go#L91

Added it. Kindly review it

pkg/pki/certmanagerpki/certmanager_user_test.go
```diff
@@ -103,6 +104,14 @@ func TestReconcileUserCertificate(t *testing.T) {
 		t.Error("Expected no error, got:", err)
 	}
 
+	// Test cert duration
+	user = newMockUser()
+	user.Spec.ExpirationSeconds = util.Int32Pointer(7200)
+
+	if _, err := manager.ReconcileUserCertificate(ctx, user, scheme.Scheme, clusterDomain); err != nil {
+		t.Error("Expected no error, got:", err)
+	}
```
panyuenlau, line 113: Instead of only verifying err, can we also verify the actual cert duration (IMO this is the most important part to verify here)?

Something like this as a reference:
```suggestion
	userCert, err := manager.ReconcileUserCertificate(ctx, user, scheme.Scheme, clusterDomain)
	if err != nil {
		t.Error("Expected no error, got:", err)
	}
	
	decodedCert, err := certutil.DecodeCertificate(userCert.Certificate)
	if err != nil {
		t.Error("Expected no error, got:", err)
	}
	
	if user.Spec.ExpirationSeconds != cert.NotAfter.Sub(cert.NotBefore).Seconds() {
	     t.Error("Expected cert duration is not the same as the actual cert duration, expected: %d, got: %d", user.Spec.ExpirationSeconds, cert.NotAfter.Sub(cert.NotBefore).Seconds())
	}
```

shubhamcoc: Hi @panyuenlau, it seems if we compare the values, it is not working because time related fields are not populated properly in test env, NotAfter and NotBefore both have zero value of time.Time (i.e. 0001-01-01 00:00:00 +0000 UTC) and the difference is coming as zero. I think we have to check for no err for now. What do you think?

shubhamcoc: Reverted the change.

panyuenlau (approved):

bartam1 (approved):

Issue #983, opened 2023-05-25
Kafka User deletion failed


shubhamcoc:
## Description
If the Kafka User is created with ACLs,  the deletion of the Kafka user is failing.

Banzaicloud koperator logs says "failed to finalize kafkauser" with error EOF.
```
{"level":"info","ts":"2023-05-24T11:20:53.218Z","msg":"Reconciling KafkaUser","controller":"KafkaUser","controllerGroup":"kafka.banzaicloud.io","controllerKind":"KafkaUser","KafkaUser":{"name":"example-kafkauser-with-pki","namespace":"default"},"namespace":"default","name":"example-kafkauser-with-pki","reconcileID":"d1c7e23c-cf67-47f3-8675-c8c6b6579a68"}
{"level":"info","ts":"2023-05-24T11:20:53.223Z","msg":"Kafka user is marked for deletion, revoking certificates","controller":"KafkaUser","controllerGroup":"kafka.banzaicloud.io","controllerKind":"KafkaUser","KafkaUser":{"name":"example-kafkauser-with-pki","namespace":"default"},"namespace":"default","name":"example-kafkauser-with-pki","reconcileID":"d1c7e23c-cf67-47f3-8675-c8c6b6579a68"}
{"level":"info","ts":"2023-05-24T11:20:53.223Z","msg":"Deleting user ACLs from kafka","controller":"KafkaUser","controllerGroup":"kafka.banzaicloud.io","controllerKind":"KafkaUser","KafkaUser":{"name":"example-kafkauser-with-pki","namespace":"default"},"namespace":"default","name":"example-kafkauser-with-pki","reconcileID":"d1c7e23c-cf67-47f3-8675-c8c6b6579a68"}
{"level":"info","ts":"2023-05-24T11:20:53.223Z","msg":"Creating kafka client connection","controller":"KafkaUser","controllerGroup":"kafka.banzaicloud.io","controllerKind":"KafkaUser","KafkaUser":{"name":"example-kafkauser-with-pki","namespace":"default"},"namespace":"default","name":"example-kafkauser-with-pki","reconcileID":"d1c7e23c-cf67-47f3-8675-c8c6b6579a68"}
[...]

PR #984, opened 2023-05-25
Fixing Delete acl functionality

shubhamcoc:
## Description

Fixing Delete ACL functionality as it was not passing correct arguments. Just passing the dn in sarama.AclFilter struct leading to take default values (i.e. Unknown) for other arguments as a result it was failing in the validation at kafka and resulting in connection close with EOF error.

```
private void normalizeAndValidate() {
        if (version() == 0) {
            for (DeleteAclsRequestData.DeleteAclsFilter filter : data.filters()) {
                PatternType patternType = PatternType.fromCode(filter.patternTypeFilter());

                // On older brokers, no pattern types existed except LITERAL (effectively). So even though ANY is not
                // directly supported on those brokers, we can get the same effect as ANY by setting the pattern type
                // to LITERAL. Note that the wildcard `*` is considered `LITERAL` for compatibility reasons.
                if (patternType == PatternType.ANY)
                    filter.setPatternTypeFilter(PatternType.LITERAL.code());
                else if (patternType != PatternType.LITERAL)
                    throw new UnsupportedVersionException("Version 0 does not support pattern type " +
                            patternType + " (only LITERAL and ANY are supported)");
            }
        }

        final boolean unknown = data.filters().stream().anyMatch(filter ->
                filter.patternTypeFilter() == PatternType.UNKNOWN.code()
                        || filter.resourceTypeFilter() == ResourceType.UNKNOWN.code()
                        || filter.operation() == AclOperation.UNKNOWN.code()
                        || filter.permissionType() == AclPermissionType.UNKNOWN.code()
        );

        if (unknown) {
            throw new IllegalArgumentException("Filters contain UNKNOWN elements, filters: " + data.filters());
        }
    }
```

linked issue #983
## Type of Change

- [x] Bug Fix

## Checklist


shubhamcoc:
Hi @panyuenlau, @bartam1, Kindly review the patch.

controllers/kafkauser_controller.go
```diff
@@ -332,8 +332,10 @@ func (r *KafkaUserReconciler) checkFinalizers(ctx context.Context, cluster *v1be
 	var err error
 	if util.StringSliceContains(instance.GetFinalizers(), userFinalizer) {
 		if len(instance.Spec.TopicGrants) > 0 {
-			if err = r.finalizeKafkaUserACLs(reqLogger, cluster, user); err != nil {
-				return requeueWithError(reqLogger, "failed to finalize kafkauser", err)
+			for _, topicGrant := range instance.Spec.TopicGrants {
+				if err = r.finalizeKafkaUserACLs(reqLogger, cluster, user, topicGrant.PatternType); err != nil {
```
panyuenlau, line 336: Passing `topicGrant.PatternType` directly could be problematic due to these few lines in the `CreateUserACLs` function: https://github.com/banzaicloud/koperator/blob/master/pkg/kafkaclient/users.go#L45-L47

If the `topicGrant.PatternType` is empty, then when it gets to:
```golang
aclPatternType := AclPatternTypeMapping(patternType)
```
the `aclPatternType` will be `sarama.AclPatternUnknown`

And at the end when the delete ACL request gets to the Kafka cluster, the request will fail in the validation phase, for example: 

```
[2023-06-08 19:10:59,462] ERROR Exception while processing request from 10.244.4.61:29092-127.0.0.6:44155-237 (kafka.network.Processor)
org.apache.kafka.common.errors.InvalidRequestException: Error getting request for apiKey: DELETE_ACLS, apiVersion: 1, connectionId: 10.244.4.61:29092-127.0.0.6:44155-237, listenerName: ListenerName(INTERNAL), principal: User:CN=kafka-kafka-operator
Caused by: java.lang.IllegalArgumentException: Filters contain UNKNOWN elements, filters: [DeleteAclsFilter(resourceTypeFilter=1, resourceNameFilter=null, patternTypeFilter=0, principalFilter='CN=kafka-user', hostFilter=null, operation=6, permissionType=1)]
```

Therefore, we need the [same if statement](https://github.com/banzaicloud/koperator/blob/master/pkg/kafkaclient/users.go#L45-L47) to make sure the `topicGrant.PatternType` is not empty

panyuenlau (changes requested):
The direction of this is correct, but we are missing some details here

pkg/kafkaclient/users.go
```diff
@@ -68,9 +68,14 @@ func (k *kafkaClient) ListUserACLs() ([]sarama.ResourceAcls, error) {
...
 	matches, err := k.admin.DeleteACL(sarama.AclFilter{
-		Principal: &dn,
+		Principal:                 &dn,
+		ResourcePatternTypeFilter: aclPatternType,
+		Operation:                 sarama.AclOperationDelete,
+		ResourceType:              sarama.AclResourceTopic,
+		PermissionType:            sarama.AclPermissionAny,
 	}, false)
 	if err != nil {
 		return
```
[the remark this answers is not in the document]
shubhamcoc: > while going through the PR, I've found this minor bug here

I think its not a bug, it was a "named parameter" as a result it will return the err, but I changed it to more generic way of writing for better readability.

controllers/kafkauser_controller.go
shubhamcoc: Nice catch. Thank you for pointing out. Added the suggested line in DeleteUserACLs function instead of adding here, as it is better to take care in the lower function in case it is used somewhere else in future.

pkg/kafkaclient/users_test.go
```diff
@@ -81,16 +81,25 @@ func TestCreateUserACLs(t *testing.T) {
...
-	if err := client.DeleteUserACLs("with-error"); err == nil {
-		t.Error("Expected error, got nil")
+	for _, patternType := range validPatternTypes {
+		if err := client.DeleteUserACLs("test-user", patternType); err != nil {
+			t.Error("Expected no error, got:", err)
+		}
+
+		if err := client.DeleteUserACLs("with-error", patternType); err == nil {
+			t.Error("Expected error, got nil")
+		}
```
panyuenlau, line 98: Can we also add a test case when `patternType` is empty to catch the case that we just caught in the last commit?

panyuenlau (approved):

pregnor (approved):
LGTM

PR #988, opened 2023-05-30
Add kafka cluster create test with our simplekafkacluster example and rename zookeeper cluster

Kuvesz:
## Description

This PR adds a test case for creating a kafka cluster, to be more specific the `simplekafkacluster` example that we have. It also updates the zookeeper cluster naming to `zookeeper-server` from just `zookeeper`.

## Type of Change
- [x] Other (test)

## Checklist


bartam1:
I think we should do the renaming in different PR.

tests/e2e/k8s_test.go
```diff
@@ -119,6 +119,17 @@ func applyK8sResourceFromTemplate(kubectlOptions *k8s.KubectlOptions, templateFi
 	applyK8sResourceManifestFromString(kubectlOptions, manifest.String())
 }
 
+// checkExistenceOfK8sResource queries a Resource by it's kind, namespace and name and
+// returns the output of stderr
+func checkExistenceOfK8sResource(
```
bartam1, line 124: We already have similar function: https://github.com/banzaicloud/koperator/blob/b44b042f6ade0335d8617fb3b2f7e5230f9cc6c5/tests/e2e/k8s_test.go#LL278C18-L278C18

Kuvesz:
> I think we should do the zookeeper cluster renaming in different PR.

I tend to agree, I was just a bit lazy because this PR would then have to be based on that one, which means another branch added to this testing stuff, I'd rather not bother if possible, though it would be nicer.

tests/e2e/k8s_test.go
```diff
@@ -127,6 +138,7 @@ func createK8sResourcesFromManifest(kubectlOptions *k8s.KubectlOptions, manifest
 		GinkgoT(),
 		kubectlOptions,
 		"create",
+		"-n", kubectlOptions.Namespace,
```
bartam1, line 141: This should not be necessary. The -n namespace should be included automatically in the query.

tests/e2e/kafka_cluster_test.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// createKafkaClusterIfDoesNotExist creates a kafka cluster if
+// there isn't a preexisting one
+func createKafkaClusterIfDoesNotExist(kubectlOptions *k8s.KubectlOptions, path string) {
+	By("Checking existing kafkaClusters")
+	err := checkExistenceOfK8sResource(kubectlOptions, kafkaKind, kafkaClusterName)
+
+	if err == nil {
+		fmt.Printf("Kafka cluster %s already exists", kafkaClusterName)
```
bartam1, line 38: I think in the test we cannot write out things with fmt.Printf. This message will be hidden.

tests/e2e/k8s_test.go
Kuvesz: Just checked, not sure why I added that but I seemed to remember a kafka cluster wrongly deploying into the default namespace but it seems correct now, may have been something else, will remove.

tests/e2e/kafka_cluster_test.go
Kuvesz: It can:
```
STEP: Checking the existence of resource kafka @ 05/31/23 11:03:50.937
when Creating a simple kafka cluster Creating a Kafka cluster 2023-05-31T11:03:50+02:00 logger.go:66: Running command kubectl with args [--namespace kafka get kafkaclusters.kafka.banzaicloud.io kafka]
when Creating a simple kafka cluster Creating a Kafka cluster 2023-05-31T11:03:51+02:00 logger.go:66: NAME    CLUSTER STATE    CLUSTER ALERT COUNT   LAST SUCCESSFUL UPGRADE   UPGRADE ERROR COUNT   AGE
when Creating a simple kafka cluster Creating a Kafka cluster 2023-05-31T11:03:51+02:00 logger.go:66: kafka   ClusterRunning   0                                               0                     3m9s
Kafka cluster kafka already exists
```

tests/e2e/kafka_cluster_test.go
```diff
@@ -0,0 +1,57 @@
...
+		createK8sResourcesFromManifest(kubectlOptions, path, false)
+	}
+
+	return
+}
+
+// requireCreatingKafkaCluster creates a Kafka cluster and
+// checks the success of that operation.
+func requireCreatingKafkaCluster(kubectlOptions *k8s.KubectlOptions, path string) {
+	It("Creating a Kafka cluster", func() {
```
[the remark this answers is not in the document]
Kuvesz: done!

tests/e2e/kafka_cluster_test.go
Kuvesz: Though using By looks nicer, I'll change it later.

tests/e2e/kafka_cluster_test.go
```diff
@@ -0,0 +1,57 @@
...
+		createKafkaClusterIfDoesNotExist(kubectlOptions, path)
+		By("Verifying Kafka cluster and it's pods")
+		waitK8sResourceCondition(kubectlOptions, kafkaKind, fmt.Sprintf("jsonpath={.status.state}=%s", string(v1beta1.KafkaClusterRunning)), "300s", "", kafkaClusterName)
+		// TODO: Can't really get this to work right now, might have to create a new function or update this one a bit
+		//requireRunningPods(kubectlOptions, "kafka_cr", "kafka-0", "kafka-1", "kafka-2")
```
bartam1, line 64: Here we should do a wait test: 
kubectl --namespace kafka wait pods --for=condition=Ready --timeout=5s --selector=kafka_cr=almafa

tests/e2e/kafka_cluster_test.go
```diff
@@ -0,0 +1,57 @@
...
+// there isn't a preexisting one
+func createKafkaClusterIfDoesNotExist(kubectlOptions *k8s.KubectlOptions, path string) {
+	By("Checking existing kafkaClusters")
+	err := checkExistenceOfK8sResource(kubectlOptions, kafkaKind, kafkaClusterName)
+
+	if err == nil {
+		fmt.Printf("Kafka cluster %s already exists", kafkaClusterName)
+	} else {
+		By("Deploying a KafkaCluster")
+		createK8sResourcesFromManifest(kubectlOptions, path, false)
```
bartam1, line 41: Here we should use sophisticated resource deployment to make version compatible the simplekafkacluster.
Because our koperator deployment test can run in different koperator versions (https://github.com/banzaicloud/koperator/blob/d9b806a3008a1fee2c30fc26eb3535a70e7ce541/tests/e2e/koperator_test.go#L124)
It would be good to make capable to use the corresponding kafkacluster sample file.
I wrote a method for this usecase what can be used: https://github.com/banzaicloud/koperator/blob/d9b806a3008a1fee2c30fc26eb3535a70e7ce541/tests/e2e/koperator_test.go#L71

tests/e2e/k8s_test.go
```diff
@@ -119,6 +119,17 @@ func applyK8sResourceFromTemplate(kubectlOptions *k8s.KubectlOptions, templateFi
...
 }
 
+// checkExistenceOfK8sResource queries a Resource by it's kind, namespace and name and
+// returns the output of stderr
+func checkExistenceOfK8sResource(
+	kubectlOptions *k8s.KubectlOptions,
+	resourceKind string,
+	resourceName string,
+) error {
+	By(fmt.Sprintf("Checking the existence of resource %s", resourceName))
```
bartam1, line 129: We should extend the log message with  kind name and namespace

config/samples/zookeeper_sample.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: zookeeper.pravega.io/v1beta1
```
bartam1, line 1: We should not add zookeeperCluster CR into koperator sample files.
I suggest to use template files for tests like here: https://github.com/banzaicloud/koperator/blob/8ee5d6dad71b558585a780e77f1aa824f64d2d5b/tests/e2e/kafka_prod_consum_test.go#L44

tests/e2e/kafka_cluster_test.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	return
+}
+
+// requireCreatingKafkaCluster creates a Kafka cluster and
+// checks the success of that operation.
+func requireCreatingKafkaCluster(kubectlOptions *k8s.KubectlOptions, koperatorVersion string, sampleFile string) {
+	When("Creating a Kafka cluster", func() {
+		It("Deploying a Kafka cluster", func() {
```
bartam1, line 49: If we want to follow Patrik's way (https://github.com/banzaicloud/koperator/blob/ef6db9416dc459e992587ad259c00da1ac2fa130/tests/e2e/koperator_test.go#L173) we should move the "It("Deploying a Kafka cluster", func()"
into requireCreateKafkaClusterIfDoesNotExist (https://github.com/banzaicloud/koperator/blob/ef6db9416dc459e992587ad259c00da1ac2fa130/tests/e2e/kafka_cluster_test.go#L31)

tests/e2e/kafka_cluster_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+func requireKafkaClusterReady(kubectlOptions *k8s.KubectlOptions) {
+	It("Verifying Kafka cluster health", func() {
+		By("Verifying the Kafka cluster resource")
+		waitK8sResourceCondition(kubectlOptions, kafkaKind, fmt.Sprintf("jsonpath={.status.state}=%s", string(v1beta1.KafkaClusterRunning)), kafkaClusterCreateTimeout, "", kafkaClusterName)
+		By("Verifying the CruiseControl pod")
+		Eventually(context.Background(), func() bool {
+			resources := getK8sResources(kubectlOptions, []string{"pod"}, "kafka_cr="+kafkaClusterName+",app=cruisecontrol", kafkaClusterName+"-cruisecontrol")
+			if len(resources) > 1 {
+				waitK8sResourceCondition(kubectlOptions, "pod", "condition=Ready", cruiseControlPodReadinessTimeout, "kafka_cr="+kafkaClusterName+",app=cruisecontrol", kafkaClusterName+"-cruisecontrol")
```
bartam1, line 64: Here we should not have to specify the name because selector is enough (waitK8sResourceCondition(kubectlOptions, "pod", "condition=Ready", cruiseControlPodReadinessTimeout, "kafka_cr="+kafkaClusterName+",app=cruisecontrol", "")
Same applies here: https://github.com/banzaicloud/koperator/blob/ef6db9416dc459e992587ad259c00da1ac2fa130/tests/e2e/kafka_cluster_test.go#L62

tests/e2e/kafka_cluster_test.go
```diff
@@ -0,0 +1,72 @@
...
+				return true
+			}
+			return false
+		}, kafkaClusterResourceCleanupTimeout, 3*time.Second).Should(BeTrue())
```
bartam1, line 68: I think we should use a timeout with different name.

tests/e2e/zookeeper_test.go
```diff
@@ -92,9 +118,9 @@ func requireRemoveZookeeperOperatorCRDs(kubectlOptions *k8s.KubectlOptions) {
 func requireZookeeperClusterReady(kubectlOptions *k8s.KubectlOptions) {
 	It("Verifying Zookeeper cluster health", func() {
 		By("Verifying the Zookeeper cluster resource")
-		waitK8sResourceCondition(kubectlOptions, zookeeperCRDs()[0], "jsonpath={.status.readyReplicas}=1", "240s", "", zookeeperClusterName)
+		waitK8sResourceCondition(kubectlOptions, zookeeperCRDs()[0], "jsonpath={.status.readyReplicas}=1", zookeeperClusterCreateTimeout, "", zookeeperClusterName)
```
bartam1, line 121: I think it is better to use here: https://github.com/banzaicloud/koperator/blob/ef6db9416dc459e992587ad259c00da1ac2fa130/tests/e2e/const.go#L37
I dont like this zookeeperCRDs()[0] :)

tests/e2e/zookeeper_test.go
```diff
@@ -92,9 +118,9 @@ func requireRemoveZookeeperOperatorCRDs(kubectlOptions *k8s.KubectlOptions) {
...
 		By("Verifying the Zookeeper cluster's pods")
-		waitK8sResourceCondition(kubectlOptions, "pod", "condition=Ready", "60s", "app="+zookeeperClusterName, zookeeperClusterName)
+		waitK8sResourceCondition(kubectlOptions, "pod", "condition=Ready", defaultPodReadinessWaitTime, "app="+zookeeperClusterName, zookeeperClusterName)
```
bartam1, line 123: We should only use the selector here. In case when multiple zookeeper server is present we can check all of readiness.
We should not use selector and name altogether for waitK8sResourceCondition and getK8sResources

Kuvesz: I mean, why not? Here if we have multiple zookeeper clusters do we really care about all of them or just the one that we are trying to deploy right now? I feel it should be okay to only check the one we are deploying instead of all of them, but I may be misunderstanding something.

bartam1: The selector "app="+zookeeperClusterName  selects one zookeeper cluster. I think this check is not working right now but I have not tested. It seems that you want to check a pod readiness and select the pod with selector and with a zookeeperClusterName which is wrong because it is not a pod name.
The only reason for that is work because I wrote this function in this way: https://github.com/banzaicloud/koperator/blob/ef6db9416dc459e992587ad259c00da1ac2fa130/tests/e2e/utils.go#L53 :)

bartam1: When you using names then you need to specify the resource exact name so the selector is kinda useless.
kubectl does not support both (error: name cannot be provided when a selector is specified)

tests/e2e/zookeeper_test.go
```diff
@@ -25,11 +25,43 @@ import (
 	. "github.com/onsi/gomega"
 )
 
+// createZookeeperClusterIfDoesNotExist creates a zookeeper cluster if
+// there isn't a preexisting one
+func createZookeeperClusterIfDoesNotExist(kubectlOptions *k8s.KubectlOptions, zkClusterReplicaCount int) {
```
bartam1, line 30: I think we should move the zookeeperCluster related tests in separate file like you did for kafkaCluster

tests/e2e/kafka_cluster_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/gruntwork-io/terratest/modules/k8s"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// createKafkaClusterIfDoesNotExist creates a kafka cluster if
+// there isn't a preexisting one
+func createKafkaClusterIfDoesNotExist(kubectlOptions *k8s.KubectlOptions, koperatorVersion string, sampleFile string) {
+	It("Deploying a Kafka cluster", func() {
```
panyuenlau, line 32:
```suggestion
	It("Deploying a KafkaCluster", func() {
```

tests/e2e/kafka_cluster_test.go
```diff
@@ -0,0 +1,70 @@
...
+		By("Checking existing kafkaClusters")
```
panyuenlau, line 33:
```suggestion
		By("Checking existing KafkaClusters")
```

tests/e2e/kafka_cluster_test.go
```diff
@@ -0,0 +1,70 @@
...
+		err := checkExistenceOfK8sResource(kubectlOptions, kafkaKind, kafkaClusterName)
+
+		if err == nil {
+			By(fmt.Sprintf("Kafka cluster %s already exists\n", kafkaClusterName))
```
panyuenlau, line 37:
```suggestion
			By(fmt.Sprintf("KafkaCluster %s already exists\n", kafkaClusterName))
```

tests/e2e/kafka_cluster_test.go
```diff
@@ -0,0 +1,70 @@
...
+func requireCreatingKafkaCluster(kubectlOptions *k8s.KubectlOptions, koperatorVersion string, sampleFile string) {
+	When("Creating a Kafka cluster", func() {
+		createKafkaClusterIfDoesNotExist(kubectlOptions, koperatorVersion, sampleFile)
+		requireKafkaClusterReady(kubectlOptions)
+	})
+}
+
+func requireKafkaClusterReady(kubectlOptions *k8s.KubectlOptions) {
+	It("Verifying Kafka cluster health", func() {
+		By("Verifying the Kafka cluster resource")
```
panyuenlau, line 56:
```suggestion
		By("Verifying the KafkaCluster state")
```

tests/e2e/kafka_cluster_test.go
```diff
@@ -0,0 +1,70 @@
...
+		Eventually(context.Background(), func() bool {
+			resources := getK8sResources(kubectlOptions, []string{"pod"}, "kafka_cr="+kafkaClusterName+",app=cruisecontrol", "")
+			if len(resources) > 1 {
+				waitK8sResourceCondition(kubectlOptions, "pod", "condition=Ready", cruiseControlPodReadinessTimeout, "kafka_cr="+kafkaClusterName+",app=cruisecontrol", "")
+				return true
+			}
+			return false
+		}, kafkaClusterResourceReadinessTimeout, 3*time.Second).Should(BeTrue())
+		By("Verifying all Kafka pods")
+		waitK8sResourceCondition(kubectlOptions, "pod", "condition=Ready", defaultPodReadinessWaitTime, "kafka_cr="+kafkaClusterName, "")
```
panyuenlau, line 68:
```suggestion
		waitK8sResourceCondition(kubectlOptions, "pod", "condition=Ready", defaultPodReadinessWaitTime, v1beta1.KafkaCRLabelKey+"="+kafkaClusterName, "")
```

tests/e2e/kafka_cluster_test.go
```diff
@@ -0,0 +1,70 @@
...
+	})
+}
+
+func requireKafkaClusterReady(kubectlOptions *k8s.KubectlOptions) {
+	It("Verifying Kafka cluster health", func() {
+		By("Verifying the Kafka cluster resource")
+		waitK8sResourceCondition(kubectlOptions, kafkaKind, fmt.Sprintf("jsonpath={.status.state}=%s", string(v1beta1.KafkaClusterRunning)), kafkaClusterCreateTimeout, "", kafkaClusterName)
+		By("Verifying the CruiseControl pod")
+		Eventually(context.Background(), func() bool {
+			resources := getK8sResources(kubectlOptions, []string{"pod"}, "kafka_cr="+kafkaClusterName+",app=cruisecontrol", "")
```
panyuenlau, line 60: I'd use this reserved const https://github.com/banzaicloud/koperator/blob/master/api/v1beta1/kafkacluster_types.go#L54 rather than a hardcoded string here

```suggestion
			resources := getK8sResources(kubectlOptions, []string{"pod"}, v1beta1.KafkaCRLabelKey+"="+kafkaClusterName+",app=cruisecontrol", "")
```

tests/e2e/kafka_cluster_test.go
```diff
@@ -0,0 +1,70 @@
...
+			if len(resources) > 1 {
+				waitK8sResourceCondition(kubectlOptions, "pod", "condition=Ready", cruiseControlPodReadinessTimeout, "kafka_cr="+kafkaClusterName+",app=cruisecontrol", "")
```
panyuenlau, line 62:
```suggestion
				waitK8sResourceCondition(kubectlOptions, "pod", "condition=Ready", cruiseControlPodReadinessTimeout, v1beta1.KafkaCRLabelKey+"="+kafkaClusterName+",app=cruisecontrol", "")
```

tests/e2e/koperator_test.go
```diff
@@ -39,45 +39,42 @@ import (
 // The full path of the manifest also can be specified.
 // It supports different versions that can be specified with the koperatorVersion parameter.
 func requireApplyingKoperatorSampleResource(kubectlOptions *k8s.KubectlOptions, koperatorVersion Version, sampleFile string) {
-	It("Applying Koperator sample resource", func() {
-		By(fmt.Sprintf("Retrieving Koperator sample resource: '%s' with version: '%s' ", sampleFile, koperatorVersion))
+	By(fmt.Sprintf("Retrieving Koperator sample resource: '%s' with version: '%s' ", sampleFile, koperatorVersion))
```
panyuenlau, line 42:
```suggestion
	By(fmt.Sprintf("Retrieving Koperator sample resource: '%s' with version: '%s'", sampleFile, koperatorVersion))
```

tests/e2e/zookeeper_cluster_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+// createZookeeperClusterIfDoesNotExist creates a zookeeper cluster if
+// there isn't a preexisting one
+func createZookeeperClusterIfDoesNotExist(kubectlOptions *k8s.KubectlOptions, zkClusterReplicaCount int) {
+	It("Deploying a Zookeeper cluster", func() {
+		By("Checking existing ZookeeperClusters")
+		err := checkExistenceOfK8sResource(kubectlOptions, zookeeperKind, zookeeperClusterName)
+
+		if err == nil {
+			By(fmt.Sprintf("Zookeeper cluster %s already exists\n", zookeeperClusterName))
```
panyuenlau, line 36:
```suggestion
			By(fmt.Sprintf("ZookeeperCluster %s already exists\n", zookeeperClusterName))
```

tests/e2e/zookeeper_cluster_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/gruntwork-io/terratest/modules/k8s"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// createZookeeperClusterIfDoesNotExist creates a zookeeper cluster if
+// there isn't a preexisting one
+func createZookeeperClusterIfDoesNotExist(kubectlOptions *k8s.KubectlOptions, zkClusterReplicaCount int) {
+	It("Deploying a Zookeeper cluster", func() {
```
panyuenlau, line 31:
```suggestion
	It("Deploying a ZookeeperCluster", func() {
```

panyuenlau (approved):
LGTM in general. 

I've found that we are using a few words interchangeably here:
-  "Kafka cluster" vs "KafkaCluster" vs "kafka" 
- "Zookeeper cluster" vs "ZooKeeper"

> note: same pattern in some of our source code as well, e.g. "Kafka topic" vs "KafkaTopic"

I've found this to be very confusing in certain scenarios, IMO, "KafkaCluster", "ZooKeeper", "KafkaTopic", etc should all refer to the corresponding K8s custom resource while "Kafka cluster", "Zookeeper cluster", "Kafka topic" should all refer to the actual resources/objects/things that those corresponding K8s custom resource are created to represent.

That's where some of my comments come from and I hope this makes sense

tests/e2e/koperator_suite_test.go
```diff
@@ -55,8 +55,9 @@ var _ = BeforeSuite(func() {
 
 var _ = When("Testing e2e test altogether", Ordered, func() {
 	//testInstall()
-	//testProduceConsumeInternal()
-	testUninstallZookeeperCluster()
+	testInstallZookeeperCluster(1)
+	testInstallKafkaCluster("local", "simplekafkacluster.yaml")
```
Kuvesz, line 59: The SSL cluster install would basically be the same, just with the filename change to `simplekafkacluster_ssl.yaml`

panyuenlau (approved):
looking sharp

tests/e2e/koperator.go
```diff
@@ -0,0 +1,81 @@
...
+	"emperror.dev/errors"
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// requireApplyingKoperatorSampleResource deploys the specified sample resource (config/samples).
+// The full path of the manifest also can be specified.
+// It supports different versions that can be specified with the koperatorVersion parameter.
+func requireApplyingKoperatorSampleResource(kubectlOptions k8s.KubectlOptions, koperatorVersion Version, sampleFile string) error {
```
bartam1, line 34: I would only use the require function name prefix when the function contains "Expect" in any level.

tests/e2e/koperator.go
```diff
@@ -0,0 +1,81 @@
...
+
+		rawKoperatorSampleResource, err = io.ReadAll(response.Body)
+		if err != nil {
+			return err
+		}
+	}
+
+	By(fmt.Sprintf("Applying K8s manifest %s", sampleFile))
+	k8s.KubectlApplyFromString(GinkgoT(), &kubectlOptions, string(rawKoperatorSampleResource))
+	return nil
```
bartam1, line 80: We should return the k8s.KubectlApplyFromString(GinkgoT(), &kubectlOptions, string(rawKoperatorSampleResource)) error

tests/e2e/test_install_cluster.go
```diff
@@ -0,0 +1,51 @@
...
+			kubectlOptions, err = kubectlOptionsForCurrentContext()
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		kubectlOptions.Namespace = zookeeperOperatorHelmDescriptor.Namespace
+		requireCreatingZookeeperCluster(kubectlOptions, replicaCount)
+	})
+}
+
+func testInstallKafkaCluster(kafkaClusterVersion string, kafkaClusterSample string) bool {
```
bartam1, line 38: kafkaClusterVersion -> koperatorVersion

tests/e2e/test_install_cluster.go
```diff
@@ -0,0 +1,51 @@
...
+		var kubectlOptions k8s.KubectlOptions
+		var err error
+
+		It("Acquiring K8s config and context", func() {
+			kubectlOptions, err = kubectlOptionsForCurrentContext()
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		kubectlOptions.Namespace = koperatorLocalHelmDescriptor.Namespace
+		requireCreatingKafkaCluster(kubectlOptions, kafkaClusterVersion, kafkaClusterSample)
```
bartam1, line 49: kafkaClusterVersion -> koperatorVersion

tests/e2e/test_install_cluster.go
```diff
@@ -0,0 +1,51 @@
...
+
+package e2e
+
+import (
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func testInstallZookeeperCluster(replicaCount int) bool {
```
bartam1, line 23: I think we can use replicaCount as a const variable

bartam1 (changes requested):

tests/e2e/koperator.go
```diff
@@ -0,0 +1,81 @@
...
+
+		rawKoperatorSampleResource, err = os.ReadFile(sampleFile)
+		if err != nil {
+			return err
+		}
+	default:
+		httpClient := new(http.Client)
+		httpClient.Timeout = 5 * time.Second
+
+		Expect(sampleFileSplit).Should(HaveLen(1))
```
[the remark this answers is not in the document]
Kuvesz: forgot to get rid of it, thanks!

tests/e2e/koperator.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+// requireApplyingKoperatorSampleResource deploys the specified sample resource (config/samples).
+// The full path of the manifest also can be specified.
+// It supports different versions that can be specified with the koperatorVersion parameter.
+func requireApplyingKoperatorSampleResource(kubectlOptions k8s.KubectlOptions, koperatorVersion Version, sampleFile string) error {
+	By(fmt.Sprintf("Retrieving Koperator sample resource: '%s' with version: '%s' ", sampleFile, koperatorVersion))
+
+	sampleFileSplit := strings.Split(sampleFile, "/")
+	if len(sampleFileSplit) == 0 {
```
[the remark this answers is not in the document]
Kuvesz: I personally disagree with this as I'd think for a user it's easier to understand that a file path should not have been empty rather than a variable shouldn't be empty but if you insist I'll change it.

tests/e2e/koperator.go
Kuvesz: That function does not return an error, instead it has a require.NoError() call in it so it will error out regardless, but I can't return that error here.

pkg/resources/kafka/configmap_test.go
```diff
@@ -45,7 +45,7 @@ func TestGetMountPathsFromBrokerConfigMap(t *testing.T) {
 listeners=INTERNAL://:29092,CONTROLLER://:29093
 log.dirs=/kafka-logs3/kafka,/kafka-logs/kafka,/kafka-logs2/kafka,/kafka-logs4/kafka
 metric.reporters=com.linkedin.kafka.cruisecontrol.metricsreporter.CruiseControlMetricsReporter\noffsets.topic.replication.factor=2
-zookeeper.connect=zookeeper-server-client.zookeeper:2181/
+zookeeper.connect=zookeeper-client.zookeeper:2181/
```
pregnor: **Required**: I guess this was unintentional and we **MUST** keep this in the long run to use the new value.

tests/e2e/helm.go
```diff
@@ -311,6 +311,12 @@ func listHelmReleases(kubectlOptions k8s.KubectlOptions) ([]*HelmRelease, error)
 		"list",
 		"--output", "json",
 	)
+
+	fmt.Println("OUTPUT: " + output)
+	if strings.Contains(output, "[]") || strings.Contains(output, "WARNING") {
+		return nil, nil
+	}
+
```
pregnor, line 319: **Required**:  We **MUST** remove the debug output.

tests/e2e/k8s.go
```diff
@@ -51,6 +51,17 @@ func applyK8sResourceManifest(kubectlOptions k8s.KubectlOptions, manifestPath st
 	k8s.KubectlApply(GinkgoT(), &kubectlOptions, manifestPath)
 }
 
+// checkExistenceOfK8sResource queries a Resource by it's kind, namespace and name and
+// returns the output of stderr
+func checkExistenceOfK8sResource(
+	kubectlOptions k8s.KubectlOptions,
+	resourceKind string,
+	resourceName string,
+) error {
```
pregnor, line 60: **Recommended**: This **SHOULD** be a bool function `isExistingK8sResource()` that is expected to be true wherever we want to assert that.
Such a function would be usable to assert not existing stuff as well.
I thought we already have a similar function in this file, but haven't checked.

tests/e2e/koperator_suite_test.go
pregnor: **Required**: IMO we **MUST** add both variants here.

tests/e2e/test_install_cluster.go
```diff
@@ -0,0 +1,51 @@
...
+
+		It("Acquiring K8s config and context", func() {
+			kubectlOptions, err = kubectlOptionsForCurrentContext()
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		kubectlOptions.Namespace = zookeeperOperatorHelmDescriptor.Namespace
+		requireCreatingZookeeperCluster(kubectlOptions, zookeeperReplicaCount)
+	})
+}
```
pregnor, line 36: **Note**: I will think about refactoring those because I'm not entirely sold on having these as separate test cases.

A test case should be something which run on its own does a full thing we want to test.

testInstall == installs the deployments
testClusterDeployment == -||- + deploys ZK and Kafka clusters
testProduceConsumeInternal == -||- + executes produce/consume with internal clients.

These can live as test steps == logical units for a bounded operation such as the requires.

This is something that becomes clear along the way, it's just some impression I started to have.

You can leave it this way for the PR but this needs immediate refactoring afterwards if we agree on something.

tests/e2e/koperator.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"emperror.dev/errors"
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	. "github.com/onsi/ginkgo/v2"
+)
+
+// applyKoperatorSampleResource deploys the specified sample resource (config/samples).
+// The full path of the manifest also can be specified.
+// It supports different versions that can be specified with the koperatorVersion parameter.
+func applyKoperatorSampleResource(kubectlOptions k8s.KubectlOptions, koperatorVersion Version, sampleFile string) error {
```
pregnor, line 33: **Question**: which part of this function is not covered by `k8sResourcesFromManifestPaths()`?

We do the split for CRDs anyway and we apply local files for local version while URL resources for remote version.

IMO we should be able to adapt that to this use case.

I feel like this is a heavily specialized function I would avoid if possible, sorry I don't have time to deeply compare now.

pregnor (commented):
LGTM in general, left a couple comments.

pkg/resources/kafka/configmap_test.go
Kuvesz: Yeah, this ws unintentional as this was already set to the zookeeper-server value by default.

tests/e2e/k8s.go
Kuvesz: What you mean is getK8SResources I'd guess, we've had hat discussian before with Márton, we decided against using that but I'm not sure why, I don't really care either way, I'll use that.

tests/e2e/koperator.go
Kuvesz: Sure it can be adapted I think but at the same time this looks a lot nicer in my opinion as this doesn't require full paths to be given. I'll change it anyway.

tests/e2e/koperator.go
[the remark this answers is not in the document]
bartam1: I would use the "sampleFile should not be empty" because sampleFile is not a path it can contain the full path. But it also can be just filename

tests/e2e/koperator.go
bartam1: You have right. We need to use here that version which returns the error: k8s.KubectlApplyFromStringE(GinkgoT(), &kubectlOptions, manifest)

tests/e2e/k8s.go
bartam1: I think it is good to have this simplified function to check a resource existance. But we need the bool version

Kuvesz: Yeah, I lso tried getK8SResource and it would be a lot uglier to use that, so I'll keep using this.

pregnor (approved):
LGTM, one little thingie, it can be done separately

tests/e2e/k8s.go
```diff
@@ -51,6 +51,21 @@ func applyK8sResourceManifest(kubectlOptions k8s.KubectlOptions, manifestPath st
...
+func isExistingK8SResource(
+	kubectlOptions k8s.KubectlOptions,
+	resourceKind string,
+	resourceName string,
+) bool {
+	By(fmt.Sprintf("Checking the existence of resource %s in namespace %s (kind: %s)", resourceName, kubectlOptions.Namespace, resourceKind))
+	err := k8s.RunKubectlE(GinkgoT(), &kubectlOptions, "get", resourceKind, resourceName)
+	if err != nil {
+		return false
+	}
```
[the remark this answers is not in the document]
Kuvesz: did it real quick

pregnor (approved):

bartam1 (approved):
LGTM!

PR #991, opened 2023-06-06
Added Koperator e2e test - external consumer producer

bartam1:
## Description

Adding basic end to end tests for Koperator.
Included tests:
- external kafka consumer producer
- external kafka consumer producer SSL

## Type of Change

- [x] Other (test)

## Checklist


tests/e2e/kafka_prod_consum_test.go
```diff
@@ -146,3 +148,128 @@ func requireInternalProducingConsumingMessage(kubectlOptions *k8s.KubectlOptions
...
+			externalAddresses = getExternalListenerAddresses(kubectlOptions, "", kafkaClusterName)
+		}
+		Expect(externalAddresses).ShouldNot(BeEmpty())
+
+		message := time.Now().String()
+		requireProducingSyncMessages(externalAddresses, topicName, message)
+
+		consumedMessages := requireConsumingMessages(externalAddresses, topicName)
+		// This check may fail when the topic is re-created and has some undeleted content.
+		Expect(consumedMessages).Should(HaveLen(1))
```
Kuvesz, line 181: Do we need this check in this case? We might end up reusing stuff in the end. Instead we could just compare with the last message from what we consumed here I think.

tests/e2e/kafka_prod_consum_test.go
```diff
@@ -146,3 +148,128 @@ func requireInternalProducingConsumingMessage(kubectlOptions *k8s.KubectlOptions
...
+		// in those cases when externalListener added dynamically.
+		// Currently we dont know much about our testing environment.
+		requireAvailableExternalKafkaAddress(kubectlOptions, "", kafkaClusterName)
+		requireExternalProducingConsumingMessage(kubectlOptions, testTopicName)
+		requireDeleteKafkaTopic(kubectlOptions, testTopicName)
+	})
+}
+
+// requireExternalProducingConsumingMessage gets the Kafka cluster external addresses from the kafkaCluster CR status
+// when externalAddresses is not specified and producing, consuming messages and make comparison between them.
```
Kuvesz, line 168:
```suggestion
// when externalAddresses is not specified. It also produces and consumes messages and makes a comparison between them.
```

Kuvesz (commented):
Looks good really, couple of small nitpicks

tests/e2e/kafka_prod_consum_test.go
panyuenlau: Will we actually run into this case that " the topic is re-created and has some undeleted content" in the test environment? If not, I think it's better to remove that comment to avoid confusions

panyuenlau (approved):

tests/e2e/kafka_prod_consum_test.go
bartam1: Ok I will remove it.

bartam1: fixed: [0e03f84](https://github.com/banzaicloud/koperator/pull/991/commits/0e03f843cc92a38634d4adb4b1c22330b51cd011)

tests/e2e/kafka_client.go
pregnor, line 1: **Question**: Shouldn't this file be merged with `kafka.go`?
What distinction would we make between the two?
And what would be the trade-off of it? (I don't see any point in separating kafka-related operations into multiple files.)

tests/e2e/koperator_suite_test.go
```diff
@@ -55,8 +55,9 @@ var _ = BeforeSuite(func() {
...
-	//testProduceConsumeInternal()
-	testUninstallZookeeperCluster()
-	testUninstallKafkaCluster()
-	testUninstall()
+	testProduceConsumeExternal("kafka-controller")
+	// testProduceConsumeInternal()
+	// testUninstallZookeeperCluster()
+	// testUninstallKafkaCluster()
+  // testUninstall()
 })
```
pregnor, line 63: **Question**: What is the reason for commenting out the internal producer/consumer test?
My impression is that `testProduceConsumeInternal` and `testProduceConsumeExternal` are leaf level tests - they contain everything else from the rest of the cases and bits of extra stuff except for the actual internal and external producer/consumer tests which are unique to those two - so I would think to have those 2 enabled when we want to run `make test-e2e`.

tests/e2e/produce_consume.go
```diff
@@ -91,3 +93,89 @@ func requireInternalProducingConsumingMessage(kubectlOptions k8s.KubectlOptions,
...
+			"",
+			kafkaClusterName,
+			kubectlArgGoTemplateExternalListenersName,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("getting external listeners name: %w", err)
+		}
+		if len(externalListenerNames) == 0 {
+			return nil, fmt.Errorf("external listener %w", ErrorNotFound)
+		}
```
pregnor, line 153: **Note**: This could be a function on its own to list external listener names, sounds like something we might want to reuse in the future, it's a refactor for later.

pregnor (dismissed):

tests/e2e/kafka_client.go
bartam1: My idea was to separate the kafka client specific operations into separate file. Like we did with the kcat.

bartam1: I think it can be moved into kafka.go.

bartam1: Maybe we should create a separate file for creating and deleting custom resources. Currently what we have in kafka.go is creating kafkaTopic and deleting kafkaTopic CR

tests/e2e/koperator_suite_test.go
bartam1: After David PR we can remove all of the commenting out for other test cases. We need the deploying kafkacluster and zookeepercluster to make them complete. Until that we cannot do complex tests. I commented out because I could test only this way by deploying kafkacluster and zookeeper by hand and run e2e test

tests/e2e/produce_consume.go
bartam1: I agree. Im going to refactor it into separate function

tests/e2e/kafka_client.go
pregnor: I can live with any of the approaches you mentioned. 
It makes sense to have a kafka_client.go with this reasoning.

panyuenlau (approved):
looking sharp

pregnor (approved):

Kuvesz (approved):
Looks good!

PR #997, opened 2023-06-14
Allow users to use well-known ports for client connections to both ingress controllers

panyuenlau:
## Description

Current implementation: when users set this `externalListener[x].anyCastPort` value, Koperator uses this directly as the corresponding container port for the envoy controller containers and target port for the associated services. 

### envoy controller
When users use  a well-known port (with port number < 1024), Koperator can handle this for `envoy` controller since the `envoy` container is run as `root`, for example:
```
kubectl exec envoy-external-kafka-6495bf4fd7-9n2b8 -c envoy -n kafka -- whoami                                                                                                                          
root
```
> But this still could be problematic if the users want to use a customized envoy image but do not wish to run it as `root`.

### istioingress controller
When users use a well-known port, Koperator creates a `IstioMeshGateway` object with using that well-known port directly as the `targetPort` for one of the service ports, for example:
```yaml
    - name: tcp-all-broker
      port: 443 # anyCastPort is set to 443
      protocol: TCP
      targetPort: 443
```
And the resulting istioingress controller container fails to use this well-known port and therefore the kafka brokers can't be reached by external clients.

## What is in this PR?

- Introduce a new optional field, `ingressControllerTargetPort` under `ExternalListenerConfig` to allow users to specify the container port that they would like to use in the ingress controller. It also extends the validation webhook implementation to prevent users from using well-known port for the `istioingress` controller's deployment.

  This way, users can use any `anyCastPort` values for client access while using either of the ingress controllers
- Add checks to the KafkaCluster validation webhook (both create and update) to prevent invalid user configurations on external listeners

## Type of Change

- [x] New Feature

## Checklist


pkg/resources/istioingress/meshgateway.go
```diff
@@ -113,7 +113,7 @@ func generateExternalPorts(kc *v1beta1.KafkaCluster, brokerIds []int,
 		Name:       fmt.Sprintf(kafkautils.AllBrokerServiceTemplate, "tcp"),
 		Protocol:   string(corev1.ProtocolTCP),
 		Port:       externalListenerConfig.GetAnyCastPort(),
-		TargetPort: &istioOperatorApi.IntOrString{IntOrString: intstr.FromInt(int(externalListenerConfig.GetAnyCastPort()))},
+		TargetPort: &istioOperatorApi.IntOrString{IntOrString: intstr.FromInt(int(externalListenerConfig.GetIngressControllerTargetPort()))},
```
panyuenlau, line 116: FYI - looks like named targetPort can't be used here, otherwise istio-operator will complain about not having an exact value for this field. But I could me missing something

Kuvesz (dismissed):
LGTM really

pkg/resources/istioingress/meshgateway.go
pregnor: That's very sad.

pkg/webhooks/kafkacluster_validator_test.go
```diff
@@ -690,3 +693,83 @@ func TestCheckExternalListenerStartingPort(t *testing.T) {
...
+						},
+						{
+							CommonListenerSpec:          v1beta1.CommonListenerSpec{Name: "test-external2"},
+							IngressControllerTargetPort: util.Int32Pointer(int32(v1beta1.MaxWellKnownPort + rand.Intn(65536-v1beta1.MaxWellKnownPort))),
+						},
+					},
+				},
+			},
+			expected: nil,
+		},
```
pregnor, line 736: **Question**: I may be missing something here, but wouldn't this test create 2 different external listeners trying to listen on the same containerport (1024+65536-1024==65536)?
Shouldn't we guard against that in the webhook as well?

pkg/webhooks/kafkacluster_validator_test.go
```diff
@@ -690,3 +693,83 @@ func TestCheckExternalListenerStartingPort(t *testing.T) {
...
+						"ExternalListener '%s' would use a well-known port number '%d'(<%d) as the ingress controller container port, and this is not permitted",
+						"test-external1", int32(1), v1beta1.MaxWellKnownPort)),
+				field.Invalid(field.NewPath("spec").Child("listenersConfig").Child("externalListeners").Index(1).
+					Child("ingressControllerTargetPort"), int32(1023),
+					invalidExternalListenerTargetPortErrMsg+": "+fmt.Sprintf(
+						"ExternalListener '%s' would use a well-known port number '%d'(<%d) as the ingress controller container port, and this is not permitted",
+						"test-external2", int32(1023), v1beta1.MaxWellKnownPort)),
+			),
+		},
+	}
```
pregnor, line 767: **Note**: I used to do error message unit testing exactly like this, but when the implementation changes this is going to require a similar update here - which sounds easy, but if it is not done by the same person and immediately after the logic change, can be tricky - so I switched to `errors.Errorf("")`-ing the error message I steal from the first test run output (where this fails on a dummy message) and just replace the test-case-local variables in the static string with formatting and pass them as arguments which is a bit easier to read as non-author and also to update. I can even make a case for using the static string, but that's up to personal preference, I wouldn't comment that, such a difference is easy to track down inside the test case.

I'm not advocating for or against any of these approaches, just wanted to share my experience.

pregnor (approved):
LGTM, take the other PR's comment into account here

pkg/webhooks/kafkacluster_validator_test.go
panyuenlau: This part is not needed anymore with the latest api tag

pkg/webhooks/kafkacluster_validator_test.go
panyuenlau: Thx, I also found that the current way of matching the test outputs to the actual outputs is very cumbersome, will try to use the easy way tmr

panyuenlau: hey @pregnor - I tried with something `append(field.ErrorList{}, errors.Errorf("test"))`, but realized `field.ErrorList{}` is expecting `[]*Error` (with the its own [`Error` type](https://pkg.go.dev/k8s.io/apimachinery@v0.26.4/pkg/util/validation/field#Error)), and the type seems to be incompatible with `errors.Errorf("")`

panyuenlau:
The changes in this validation function is not really required for the original purpose of this PR, it is more for 
something that we missed to prevent user error - https://github.com/banzaicloud/koperator/pull/997/files#diff-7161be6912ad21f597382d9e2dcef3eb028c868b8206ac2d1fafa9f2cde317f5R222-R249

pkg/webhooks/kafkacluster_validator_test.go
pregnor: Sad, thanks for trying.

pregnor (commented):
Other than my question on one of the cases, LGTM.

pkg/webhooks/kafkacluster_validator_test.go
```diff
@@ -681,6 +684,36 @@ func TestCheckExternalListenerStartingPort(t *testing.T) {
...
+						"externalStartingPort + Broker ID) that collide with either the envoy admin port ('%d'), the envoy health-check port ('%d'), "+
+						"or the ingressControllerTargetPort ('%d') for brokers %v",
+						"test-external1", int32(8081), int32(8080), int32(29092), []int32{0})),
+				field.Invalid(field.NewPath("spec").Child("listenersConfig").Child("externalListeners").Index(1).Child("externalStartingPort"), int32(8070),
+					invalidExternalListenerStartingPortErrMsg+": "+fmt.Sprintf("ExternalListener '%s' would generate external access port numbers ("+
+						"externalStartingPort + Broker ID) that collide with either the envoy admin port ('%d'), the envoy health-check port ('%d'), "+
+						"or the ingressControllerTargetPort ('%d') for brokers %v",
+						"test-external2", int32(8081), int32(8080), int32(29092), []int32{11})),
+			),
+		},
```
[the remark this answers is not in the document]
panyuenlau: `8070` is the `externalStartingPort` value, and one of the generated container port numbers is "8070 + 11" (externalStartingPort + brokerID), and this collides with the default envoy health-check port (8081)

pregnor (approved):

Kuvesz (approved):
Looks good to me, like the attention to small details like the healthcheck port. Thanks!

PR #998, opened 2023-06-14
Add ingressControllerTargetPort to KafkaCluster API

panyuenlau:
## Description

Add the API change needed for https://github.com/banzaicloud/koperator/pull/997

## Type of Change

- [x] New Feature

## Checklist

### DIFF
--- a/api/v1beta1/kafkacluster_types.go
+++ b/api/v1beta1/kafkacluster_types.go
@@ -41,6 +41,10 @@ const (
 	DefaultServiceAccountName = "default"
 	// DefaultAnyCastPort kafka anycast port that can be used by clients for metadata queries
 	DefaultAnyCastPort = 29092
+	// DefaultIngressControllerTargetPort is the default container port for the ingress controller
+	DefaultIngressControllerTargetPort = 29092
+	// MaxWellKnownPort marks the largest well known port value in Unix-based systems
+	MaxWellKnownPort = 1024
 	// DefaultEnvoyHealthCheckPort envoy health check port
 	DefaultEnvoyHealthCheckPort = 8080
 	// DefaultEnvoyAdminPort envoy admin port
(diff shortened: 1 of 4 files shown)


Kuvesz (dismissed):
LGTM

api/v1beta1/kafkacluster_types.go
```diff
@@ -544,7 +557,15 @@ type ExternalListenerConfig struct {
...
 	// configuring AnyCastPort allows kafka cluster access without specifying the exact broker
+	// If not defined, 29092 will be used for external clients to reach the kafka cluster
 	AnyCastPort *int32 `json:"anyCastPort,omitempty"`
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=65535
+	// +optional
+	// IngressControllerTargetPort defines the container port that the ingress controller uses for handling external traffic.
+	// If defined, IngressControllerTargetPort should be >= 1024 when using IstioIngress as the ingress controller.
+	// If not defined, 29092 will be used as the default IngressControllerTargetPort value.
+	IngressControllerTargetPort *int32 `json:"ingressControllerTargetPort,omitempty"`
```
pregnor, line 568: **Question**: Wouldn't it be easier to set the validation minimum port to 1025 (as`MaxWellKnownPort` + 1) and that would work for all ingress cases.
Or do we want to explicitly support non-IstioIngress cases to use ports from the [0;1024] range?

pregnor (dismissed):
LGTM

api/v1beta1/kafkacluster_types.go
pregnor: Especially considering the envoy case only working because of the root setting, I would say we can get by without <=1024 ports IMO.

panyuenlau: Done in https://github.com/banzaicloud/koperator/pull/998/commits/7e365c5d0274157a0028cb70db2db43afedeefc1

Kuvesz (approved):
Even better :))

api/v1beta1/kafkacluster_types.go
```diff
@@ -544,7 +555,14 @@ type ExternalListenerConfig struct {
 	// If accessMethod is Nodeport and externalStartingPort is set to 0 then the broker IDs are not added and the Nodeport port numbers will be chosen automatically by the K8s Service controller
 	ExternalStartingPort int32 `json:"externalStartingPort"`
 	// configuring AnyCastPort allows kafka cluster access without specifying the exact broker
+	// If not defined, 29092 will be used for external clients to reach the kafka cluster
 	AnyCastPort *int32 `json:"anyCastPort,omitempty"`
+	// +kubebuilder:validation:Minimum=1024
```
pregnor, line 560: **Question**: sorry about being this pedantic, but is 1024 safe to use already? My impression was 1025 would be the first but could be wrong, haven't looked into it.

pregnor (approved):

api/v1beta1/kafkacluster_types.go
panyuenlau: Well known ports are [0,1023] (can't really find a agreed reference by the networking industry due to my limited knowledge), but a reference from wikipedia: https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers

And I've tested with "istioingress" ingress controller, 1024 works fine as the container port within the container

pregnor: Sounds good, thanks.

PR #1001, opened 2023-06-22
Allow concurrent broker restarts within same broker rack

ctrlaltluc:
## Description

Cluster restarts (during rolling upgrades) are time consuming for big clusters.

Given that clients might be using Kafka `broker.rack` config to spread replicas across N AZs, we can speed up the cluster restart by allowing concurrent broker restarts within the same AZ. Allowing only same AZ concurrent restarts makes sure that we always have at most 1/Nth of the replicas of any topic-partition offline, not more, while the other (N-1)/N replicas are fine.
For example, using a common 3AZ setup, this would ensure that at most 1/3rd of the replicas are offline, while the other 2/3rds are in-sync.

### Key changes
- add configuration parameter `RollingUpgradeConfig.ConcurrentBrokerRestartsPerRackCount` which is by default 1
- allow concurrent broker restarts only if there are no failures or restarts in other AZs and if there are no more failures or restarts in the same AZ than specified by `RollingUpgradeConfig.ConcurrentBrokerRestartsPerRackCount` and `RollingUpgradeConfig.FailureThreshold`
- do not change the semantic of `RollingUpgradeConfig.FailureThreshold` by allowing (same as before) multiple failures even if not in the same AZ
- if `broker.rack` is not properly configured for all brokers, consider each broker as being in a separate AZ, meaning we err on the side of caution and restart at most 1 broker at once
- concurrent restarts are triggered only if Cruise Control has the `RackAwareDistributionGoal` ready and not recently violated (either as fixable or unfixable)

### Notes
- in order to enable multiple concurrent restarts, we must set both `RollingUpgradeConfig.ConcurrentBrokerRestartsPerRackCount` and `RollingUpgradeConfig.FailureThreshold` to a value above 1

## Type of Change
- [x] New Feature

## Checklist


pkg/resources/kafka/configmap_test.go
```diff
@@ -615,7 +615,6 @@ zookeeper.connect=example.zk:2181/`,
 
 		t.Run(test.testName, func(t *testing.T) {
 			mockClient := mocks.NewMockClient(mockCtrl)
-			mockClient.EXPECT().Get(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil)
```
ctrlaltluc, line 618: This might not be needed and only a remainder after cherry picking from our Adobe fork. After the API module includes required changes, I will recheck and fix any issues while validating.

pkg/resources/kafka/kafka.go
```diff
@@ -940,6 +992,53 @@ func (r *Reconciler) handleRollingUpgrade(log logr.Logger, desiredPod, currentPo
...
+	}
+	return false
+}
+
+func (r *Reconciler) getConcurrentBrokerRestartsAllowed() int {
+	if r.KafkaCluster.Spec.RollingUpgradeConfig.ConcurrentBrokerRestartsAllowed > 1 {
+		return r.KafkaCluster.Spec.RollingUpgradeConfig.ConcurrentBrokerRestartsAllowed
+	}
+	return defaultConcurrentBrokerRestartsAllowed
+}
```
panyuenlau, line 1012: Can we add this getter function to https://github.com/banzaicloud/koperator/pull/1002 as a method of `RollingUpgradeConfig` instead of having it to be a method of the `Reconciler`?

pkg/resources/kafka/kafka.go
```diff
@@ -940,6 +992,53 @@ func (r *Reconciler) handleRollingUpgrade(log logr.Logger, desiredPod, currentPo
...
+	for _, terminatingOrPendingPod := range terminatingOrPendingPods {
+		if currentPodAz != r.getBrokerAz(&terminatingOrPendingPod) {
+			return true
+		}
+	}
+	return false
+}
+
+// Checks for match between pod labels and TaintedBrokersSelector
+func (r *Reconciler) isPodTainted(log logr.Logger, pod *corev1.Pod) bool {
```
panyuenlau, line 1027: Is this function needed? I don't see where in the implementation it is being used

api/v1beta1/kafkacluster_types.go
```diff
@@ -137,6 +137,16 @@ type RollingUpgradeConfig struct {
...
+
+	// ConcurrentBrokerRestartsAllowed controls how many brokers can be restarted in parallel during a rolling upgrade. If
+	// it is set to a value greater than 1, the operator will restart up to that amount of brokers in parallel, if the
+	// brokers are within the same rack (as specified by "broker.rack" in broker read-only configs). Since using Kafka broker
+	// racks spreads out the replicas, we know that restarting multiple brokers in the same rack will not cause more than
+	// 1/Nth of the replicas of a topic-partition to be unavailable at the same time, where N is the number of racks used.
+	// This is a safe way to speed up the rolling upgrade. Note that for the rack distribution explained above, Cruise Control
+	// requires `com.linkedin.kafka.cruisecontrol.analyzer.goals.RackAwareDistributionGoal` to be configured.
+	// +optional
+	ConcurrentBrokerRestartsAllowed int `json:"concurrentBrokerRestartsAllowed,omitempty"`
```
panyuenlau, line 149: Suggestion: My first impression of seeing "ConcurrentBrokerRestartsAllowed" (before the comment and actual implementation) was that I thought this is a variable to turn on/off the feature

```suggestion
	ConcurrentBrokerRestartsCount int `json:"concurrentBrokerRestartsCount,omitempty"`
```

But it could be just me, and if so feel free to ignore this

panyuenlau (commented):
Thx for the contribution! Took a quick look at the implementation and it looks good. Will come back for the tests when I get some more time tmr

bartam1:
Hello @ctrlaltluc !
Thank you for the contribution. We are reviewing it.
I can see a problematic case when the rack awareness is only enabled after some time. In this case it can happen that there are some topics without rack awareness. Restarting those brokers which contains those topics can be problem.

What if we dont do restart based on racks rather we impelement a check which maps broker-topic-partitions on the kafka cluster and checks that the actual broker contains or not such a partition which is not appear on another healthy broker (also check is the partition is in sync) . If there is no such partition then we can restart the broker. It would be more advanced if before the broker restart there would be a new leader election so the consuming producing would be totally seamless (no need to wait for the new leader election by controller). After the rolling upgrade check that everything is in sync and then we should execute the kafka-preferred-replica-election command.

pkg/resources/kafka/kafka_test.go
```diff
@@ -950,3 +953,353 @@ func TestGetServerPasswordKeysAndUsers(t *testing.T) { //nolint funlen
...
+			},
+			outOfSyncReplicas: []int32{101},
+			errorExpected:     false,
+		},
+	}
+
+	mockCtrl := gomock.NewController(t)
+
+	for _, test := range testCases {
+		mockClient := mocks.NewMockClient(mockCtrl)
```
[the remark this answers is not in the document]
bartam1: In our source codes we are using the mockgen to generate mocks in a dynamic way (gomock).
https://github.com/banzaicloud/koperator/blob/25c4e9741a8c3a7ffe87376b2c1b1106039371f0/Makefile#L267

bartam1 (changes requested):

bartam1:
When I was reviewing the code I saw that if there is pod with terminated container during rolling upgrade then it could happen that the rolling upgrade stops because for example of the low failureThresold (because of the pod with terminated container) and that pod with terminated container will not get restarted until we reach it in the slice.

Im thinking on to create a separate for loop before rollingUpgrade which would restart those pods that has terminated container or it would be better to take consideration when the brokers are being reodered (https://github.com/banzaicloud/koperator/blob/25c4e9741a8c3a7ffe87376b2c1b1106039371f0/pkg/resources/kafka/kafka.go#L1324)

pkg/resources/kafka/kafka.go
```diff
@@ -874,18 +917,23 @@ func (r *Reconciler) handleRollingUpgrade(log logr.Logger, desiredPod, currentPo
...
 			}
-			for _, pod := range podList.Items {
-				pod := pod
-				if k8sutil.IsMarkedForDeletion(pod.ObjectMeta) {
-					return errorfactory.New(errorfactory.ReconcileRollingUpgrade{}, errors.New("pod is still terminating"), "rolling upgrade in progress")
-				}
-				if k8sutil.IsPodContainsPendingContainer(&pod) {
-					return errorfactory.New(errorfactory.ReconcileRollingUpgrade{}, errors.New("pod is still creating"), "rolling upgrade in progress")
-				}
+			if len(podList.Items) < len(r.KafkaCluster.Spec.Brokers) {
```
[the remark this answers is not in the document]
ctrlaltluc: That is a good question! 
I was thinking about the following situation:
- Kafka cluster has 6 brokers in 3 AZs: 101, 102, 201, 202, 301, 302
- failure threshold is 1, concurrent broker restarts is 1 (so the defaults)
- broker 101 was deleted and terminated, it no longer exists
- reconcile runs with current pod 102
- `r.Client.List` returns pods 201, 202, 301, 302 (101 is not returned as it is not created back yet)
- none of pods 201, 202, 301, 302 is either terminating or pending

In this case:
- the actual behavior is that 102 is deleted
- the expected behavior is that we must wait for 101 to be re-created before 102, as only checking for terminating/pending pods is not enough to guarantee there is no broker missing

This seemed to me like an edge case, in the context that all pods are being reconciled in a loop, with 1 pod reconciliation handling either delete (see `reconcileKafkaPod` 2nd switch case, which delegates to `handleRollingUpgrade` which only performs delete), either create (see `reconcileKafkaPod` 1st switch case).
So I enforced any evaluation of pod status or replica availability to happen only if the pod list result has the same amount of pods as the spec.

Does that make sense?

pkg/resources/kafka/kafka.go
ctrlaltluc: Good catch!
It is a remainder during cherry-pick from our fork. We have some other features we did not yet make upstream PRs for. Will address it after the previous PR (for API) is merged.

api/v1beta1/kafkacluster_types.go
ctrlaltluc: Good point, Marton also made it in the other PR, which I addressed by renaming to `ConcurrentBrokerRestartCountPerRack`.

ctrlaltluc:
> Hello @ctrlaltluc ! Thank you for the contribution. We are reviewing it. I can see a problematic case when the rack awareness is only enabled after some time. In this case it can happen that there are some topics without rack awareness. Restarting multiple brokers in same rack which contains all of those topics partitions can be problem. (com.linkedin.kafka.cruisecontrol.analyzer.goals.RackAwareGoal is turned off or violated)
> 
> What if we dont do restart based on racks rather we impelement a check which maps broker-topic-partitions on the kafka cluster and checks that the actual broker contains a partition which is not appeared on another healthy broker (also check is the partition in sync) . If there is no such partition then we can restart the broker. It would be more advanced if before the broker restart there would be a new leader election (for that partition which is on the actual broker what we want to restart) so the consuming producing would be totally seamless (no need to wait for the new leader election by controller). We can introduce a new field which would specify how many in sync replicas should be remain when rollingUpgrade happens. (In this way if another broker goes down unintentionally under rolling upgrade then cluster can be also helathy). After the rolling upgrade we should checks that everything is in sync and when yes then we should execute the kafka-preferred-replica-election command to achieve that state where every leader is the prefered one.
> 
> The https://github.com/linkedin/cruise-control/wiki/REST-APIs#demote-a-list-of-brokers-from-the-kafka-cluster can also be used before we restart a broker. (Actually it also can be used before remove_broker in this case the remove_broker operation could be finished sooner because there were be no leader on the removable broker which is problematic under heavy load).
> 
> Execute a demotebroker operation in rolling upgrade. When the broker is demoted do the restart and go to the next broker. If we do this it is possible that scennario when the rest of the healthy broker getting too much load.

Hi Marton @bartam1! I would like to reach a conclusion here before addressing the other PR comments.
Please note that, as I mentioned in the CRD field description, enabling concurrent restarts requires CC `com.linkedin.kafka.cruisecontrol.analyzer.goals.RackAwareDistributionGoal` to be enabled so we ensure we have uniform replica distribution across AZs. The reason for this is not only to ensure more than 1 replica per AZ, but also to balance load in other AZs, while some replicas in a certain AZ are down.

Regarding the suggestion for keeping a map of topic-partitions to brokers, I believe this would not be easy to implement as to ensure consistency. Topic-partitions might change between queries and we must keep that map consistent with actual cluster metadata. I would keep this option as a last resort.

Regarding the suggestion for using demote, I agree that would be safe, but that would also make the restart much slower and might not be necessary. I first want to understand if it is necessary.

Regarding the problem you raised, I'll try to summarize to make sure I understand that what you mention is not covered and needs addressing. Please correct me if I am wrong.

Problematic use case:
- Kafka clusters with multiple brokers, `broker.rack` disabled
- client sets proper `broker.rack` read-only config for all brokers
- Kafka starts enforcing rack distribution
- before this finishes, client sets concurrent broker restarts
Expected behavior: brokers are not yet restarted concurrently
Actual behavior: brokers are restarted concurrently and might match all of a topic's replicas as rack distribution is not yet finalized by Kafka and CC

Is this what you mean is not covered?

If this is accurate, then I would argue that:
- prerequisite for enabling CC `com.linkedin.kafka.cruisecontrol.analyzer.goals.RackAwareDistributionGoal` was not met
- if the above is enabled and CC starts a replica rebalance, the rolling restart will be prevented anyway as long as the CC task is running

Wdyt?

bartam1:
> if the above is enabled and CC starts a replica rebalance, the rolling restart will be prevented anyway as long as the CC task is running

Hello @ctrlaltluc 
Yes, the problematic use-case is that one what you wrote.
I'm not sure regarding this:
- if the above is enabled and CC starts a replica rebalance, the rolling restart will be prevented anyway as long as the CC task is running
I think the rolling restart will be not prevented in that case and we also cannot be sure that the RackAwareDistributionGoal self healing operation is started at all.

This is an edge case. 

If we want to be  sure we can check two things before multiple broker restart:
1. Is the RackAwareDistributionGoal is set in the CC configuration
2. Is there RackAwareDistributionGoal violation (CC state/AnomalyDetectorState/recentGoalViolations)
What do you think?

Regarding the demoting: It happens instantly when the target broker is in sync. I tested this with a continuously producer-consumer.  There were no any communication interruption under demoting. Otherwise when there is an active producer to the leader and there is a rolling upgrade (without demotion) there will be interruption (until the new leader is elected). I'm not sure about that is it worth or not the effort to implement what I wrote regarding mapping and demotion. I don't have much experience with production Kafka clusters. If it is worth it can be done.  I saw that as a possible improvement and I was curious about your opinion.

pkg/resources/kafka/kafka.go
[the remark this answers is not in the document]
bartam1: It makes sense. Thank you for the detailed explanation.
Can you add please a short comment into the code why is it necessary. Thank you!

ctrlaltluc:
> I was curious about your opinion.

Thanks @bartam1! Your point makes perfect sense.

I will add these additional checks before the decision to restart more than 1 broker:
- RackAwareDistributionGoal is set in the CC configuration
- RackAwareDistributionGoal is not violated

I will have a look into demoting too, as it would make it smoother for clients (no interruptions), makes sense. Will come back on this with either code changes or a comment.

bartam1:
> > I was curious about your opinion.
> 
> Thanks @bartam1! Your point makes perfect sense.
> 
> I will add these additional checks before the decision to restart more than 1 broker:
> 
>     * RackAwareDistributionGoal is set in the CC configuration
> 
>     * RackAwareDistributionGoal is not violated
> 
> 
> I will have a look into demoting too, as it would make it smoother for clients (no interruptions), makes sense. Will come back on this with either code changes or a comment.

Thank you for doing!

ctrlaltluc (commented):
@bartam1 @panyuenlau I have addressed all comments on the PR, including the check for `RackAwareDistributionGoal`. Please have another look, thanks in advance for your time!

@bartam1 I have not implemented demote brokers, as it is not necessary. By default, Kafka brokers have [controlled.shutdown.enable](https://kafka.apache.org/documentation/#brokerconfigs_controlled.shutdown.enable) default true. This means that any graceful Kafka shutdown (i.e. anything excluding SIGKILL) will [ensure that leaders are moved](https://docs.confluent.io/platform/current/kafka/post-deployment.html#graceful-shutdown).
**LE**: I also tested demote myself, and the behavior is the same as normal broker restarts (leaders are moved). However, demote has some downsides: 1/ it takes a long time (~10 min per broker for brokers with ~4500 replicas) as demote moves the demoted broker as last in any set of in-sync replicas and this takes time, even though there is no data movement, 2/ we later need to add the demoted broker back, as it would otherwise remain in recently demoted brokers, which complicates the code without added benefit.

Please note that the tests will still be failing, until https://github.com/banzaicloud/koperator/pull/1002 is merged and a new API module release is done.

pkg/resources/kafka/kafka.go
```diff
@@ -102,9 +113,42 @@ func New(client client.Client, directClient client.Reader, cluster *v1beta1.Kafk
...
+	// if incomplete broker AZ information, consider all brokers as being in different AZs
+	if len(brokerAzMap) != len(cluster.Spec.Brokers) {
+		for _, broker := range cluster.Spec.Brokers {
+			brokerAzMap[broker.Id] = fmt.Sprintf("%d", broker.Id)
+		}
+	}
+	return brokerAzMap
+}
+
+func getBrokerRack(readOnlyConfig string) string {
```
panyuenlau, line 141: Question: It looks like the latest implementation doesn't require this func, can we remove it since it's not being used?

pkg/resources/kafka/kafka.go
```diff
@@ -102,9 +113,42 @@ func New(client client.Client, directClient client.Reader, cluster *v1beta1.Kafk
 			DirectClient: directClient,
 			KafkaCluster: cluster,
 		},
-		kafkaClientProvider: kafkaClientProvider,
+		kafkaClientProvider:        kafkaClientProvider,
+		CruiseControlScalerFactory: scale.ScaleFactoryFn(),
 	}
 }
+
+func getBrokerAzMap(cluster *v1beta1.KafkaCluster) map[int32]string {
```
panyuenlau, line 121: Suggestion: It'd be appreciated if we can add some quick unit tests for func like this one, but since the caller `handleRollingUpgrade` is tested, I am not going to ask for any more changes.

panyuenlau (approved):
Apologies - was zoned out for doing my own implementation. Thx for the great work! Only left a couple of quick comments (more like nits)

pkg/resources/kafka/kafka.go
ctrlaltluc: Oh, good catch! Forgot about it. Will address in another commit, thanks for paying attention!

pkg/resources/kafka/kafka.go
ctrlaltluc: Let's do this properly. I added unit tests for it in cae78f54b77050afa23466899d1ee7a38316dee3, please have a look. Thanks!

P.S. I saw you were doing heavy lifting with kraft, so no worries and thanks for the review. :)

panyuenlau:
FYI - The new API tag https://github.com/banzaicloud/koperator/tree/api/v0.28.7 should be available soon

ctrlaltluc:
@panyuenlau @bartam1 the PR is now final and ready for review. CI fails because of the flaky nodeports test (tried once to retrigger it with an empty commit, gave up), otherwise all tests are passing. Thanks for your patience!

bartam1:
I think this can have (https://github.com/banzaicloud/koperator/pull/1023)  effect for this PR
Maybe we should not do multiple broker restarts when the brokers are controllers.
When multiple brokers are restarting it has bigger chance to not be enough controller to make quorum.

ctrlaltluc:
> I think this can have (#1023) effect for this PR Maybe we should not do multiple broker restarts when the brokers are controllers. When multiple brokers (controllers) are restarting it has bigger chance to not be enough controller to make quorum.

@bartam1 had a look on #1023 and indeed, if KRaft is enabled and the controllers are not spread across AZs (racks), then there are edge cases where multiple restarts within the same rack/AZ might be an issue.

I see 2 options to move forward:

1. Merge this PR as it is now, then when #1023 PR is merged, we will have the `spec.kraft` available and we can add another check: concurrent restarts > 1, terminating or pending pods > 1 and exists any controller in the same AZ/rack among terminating or pending pods, then do not restart another pod (to ensure at most 1 KRaft controller is restarted)
2. Wait until #1023 is merged, then add the check above in this PR.

@bartam1 @panyuenlau lmk your thoughts.

bartam1:
> > I think this can have (#1023) effect for this PR Maybe we should not do multiple broker restarts when the brokers are controllers. When multiple brokers (controllers) are restarting it has bigger chance to not be enough controller to make quorum.
> 
> @bartam1 had a look on #1023 and indeed, if KRaft is enabled and the controllers are not spread across AZs (racks), then there are edge cases where multiple restarts within the same rack/AZ might be an issue.
> 
> I see 2 options to move forward:
> 
>     1. Merge this PR as it is now, then when [Initial KRaft support #1023](https://github.com/banzaicloud/koperator/pull/1023) PR is merged, we will have the `spec.kraft` available and we can add another check: concurrent restarts > 1, terminating or pending pods > 1 and exists any controller in the same AZ/rack among terminating or pending pods, then do not restart another pod (to ensure at most 1 KRaft controller is restarted)
> 
>     2. Wait until [Initial KRaft support #1023](https://github.com/banzaicloud/koperator/pull/1023) is merged, then add the check above in this PR.
> 
> 
> @bartam1 @panyuenlau lmk your thoughts.

I agree. We should merge this PR. We can add modifications later if needed.

pkg/scale/scale.go
```diff
@@ -137,6 +137,7 @@ func (cc *cruiseControlScaler) Status(ctx context.Context) (StatusTaskResult, er
 			State:              v1beta1.CruiseControlTaskCompleted,
 		},
 		Status: &status,
+		Result: resp.Result,
```
bartam1, line 140: We should rename the Result -> State

bartam1 (changes requested):

pkg/resources/kafka/kafka.go
```diff
@@ -24,8 +24,10 @@ import (
 	"strings"
 
 	"emperror.dev/errors"
+	types2 "github.com/banzaicloud/go-cruise-control/pkg/types"
```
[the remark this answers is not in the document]
ctrlaltluc: Addressed in adac7932a9524bfd57b4949693c7c0e558c1a247.

pkg/resources/kafka/kafka.go
```diff
@@ -989,6 +999,27 @@ func (r *Reconciler) handleRollingUpgrade(log logr.Logger, desiredPod, currentPo
 	return nil
 }
 
+func (r *Reconciler) isCruiseControlRackAwareDistributionGoalViolated() error {
```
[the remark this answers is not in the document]
ctrlaltluc: Good point! I first had it return a boolean, but then iterated on it and forgot to rename. Addressed in adac7932a9524bfd57b4949693c7c0e558c1a247.

pkg/scale/scale.go
ctrlaltluc: Addressed in adac7932a9524bfd57b4949693c7c0e558c1a247.

bartam1 (approved):
LGTM!
Thank you for implementing it with consideration of the edge cases.

panyuenlau (approved):
Thx for the consideration for the potentional impact on KRaft support @bartam1 @ctrlaltluc, 
will keep this new feature in mind when we get there during the KRaft implementation.

ctrlaltluc:
Thanks for your thorough reviews as well! Looking forward to getting this merged.

PR #1011, opened 2023-07-13
add kafkauser csr approval support

LuciferInLove:
## Description

Add `CertificateSigningRequests` approval support for `KafkaUser` resources.

It's used when `pkiBackend` is `k8s-csr`, and `signerName` contains `clusterissuers.cert-manager.io`. So, it can be used in parallel with the csr-operator. 

## Type of Change
- [x] New Feature

## Checklist


pkg/pki/k8scsrpki/k8scsr.go
```diff
@@ -22,8 +22,9 @@ import (
 )
 
 const (
-	DependingCsrAnnotation     string = "banzaicloud.io/csr"
-	IncludeFullChainAnnotation string = "csr.banzaicloud.io/fullchain"
+	DependingCsrAnnotation      string = "banzaicloud.io/csr"
+	IncludeFullChainAnnotation  string = "csr.banzaicloud.io/fullchain"
+	CertManagerSignerNamePrefix string = "clusterissuers.cert-manager.io"
```
panyuenlau, line 27: good job on using a const var here instead of hardcoded value in a local function for better visibility of the code while preventing copy&paste hard-coded value errors in the future.

And I am really surprised that upstream cert-manager doesn't have const like this that we can use, they seem to be using the hardcoded string `clusterissuers.cert-manager.io` everywhere across their implementation and tests

pkg/pki/k8scsrpki/k8scsr_user.go
```diff
@@ -309,3 +320,27 @@ func isKafkaUserCertificateReady(secret *corev1.Secret, includeJKS bool) bool {
...
 	return true
 }
+
+// Approve approves certificate signing requests
+func (c *k8sCSR) Approve(ctx context.Context, signingReq *certsigningreqv1.CertificateSigningRequest) error {
+	cond := certsigningreqv1.CertificateSigningRequestCondition{
+		Type:    certsigningreqv1.CertificateApproved,
+		Status:  corev1.ConditionTrue,
+		Reason:  approveReason,
+		Message: "CSR has been approved by signer",
```
panyuenlau, line 330: I am a bit confused about this message - Koperator is not the signer here, right? My understanding was that:
1. Koperator approves CSR
2. The actual signer (in this case, cert-manager) detects CSR gets approved by the approver, it then signs the CSR

pkg/pki/k8scsrpki/k8scsr_user.go
```diff
@@ -158,8 +162,15 @@ func (c *k8sCSR) ReconcileUserCertificate(
...
-		return nil, errorfactory.New(errorfactory.FatalReconcileError{}, errors.New(notApprovedErrMsg),
-			notFoundApprovedCsrErrMsg, "csrName", signingReq.GetName())
+		if strings.Split(signingReq.Spec.SignerName, "/")[0] == CertManagerSignerNamePrefix {
+			err = c.Approve(ctx, signingReq)
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			return nil, errorfactory.New(errorfactory.FatalReconcileError{}, errors.New(notApprovedErrMsg),
+				notFoundApprovedCsrErrMsg, "csrName", signingReq.GetName())
```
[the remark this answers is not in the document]
LuciferInLove: Could you please suggest the message's text?

pkg/pki/k8scsrpki/k8scsr_user.go
LuciferInLove: Is it OK to replace this with "CSR has been approved by kafka-operator"?

panyuenlau: I think we should either use `"Koperator"` or just `"the operator"` instead of `"kafka-operator"` since this is a deprecated name of the Koperator

LuciferInLove: Ok, thanks. `"Koperator"` is the best option because some other tools can also theoretically approve CSRs.

panyuenlau (changes requested):
Implementation lgtm.

But please add these two markers to the end of https://github.com/banzaicloud/koperator/blob/master/controllers/kafkauser_controller.go#L162:
```
// +kubebuilder:rbac:groups=certificates.k8s.io,resources=certificatesigningrequests/approval,verbs=update
// +kubebuilder:rbac:groups=certificates.k8s.io,resources=signers,verbs=approve
```

Then run `make manifests` - this will update both the helm chart (it will be the same as what you currently have), and https://github.com/banzaicloud/koperator/blob/master/config/base/rbac/role.yaml (what you are missing from the PR).

Asking for this because we don't want to update the RBAC manually

pkg/pki/k8scsrpki/k8scsr_user.go
```diff
@@ -38,12 +39,15 @@ import (
...
 	"k8s.io/apimachinery/pkg/types"
+	csrclient "k8s.io/client-go/kubernetes/typed/certificates/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 const (
 	notApprovedErrMsg         = "instance is not approved"
-	notFoundApprovedCsrErrMsg = "could not find approved csr"
+	notFoundApprovedCsrErrMsg = "could not find approved csr, or the operator is not capable of approving the csr"
```
panyuenlau, line 49: This is meant to be an "and", right? (according to the implementation where this error message is used)
```suggestion
	notFoundApprovedCsrErrMsg = "could not find approved csr and the operator is not capable of approving the csr"
```

LuciferInLove: I apologize, I wrote this mechanically. Thanks for noticing.

panyuenlau (approved):
LGTM, thanks for contributing!

Kuvesz (approved):
LGTM, though it would be nice to have unit tests for new functions later on.

PR #1013, opened 2023-07-13
add function to delete namespaces during uninstall e2e test

mihaialexandrescu:
## Description

The `uninstall` e2e test does not currently remove the namespaces created by the install test/steps (via the Namespace field of helmDescriptors and the "--create-namespace" `helm` command flag).  

This PR brings a function that will do that.  
This action should be idempotent due to the call to `deleteK8sResourceNoErrNotFound()`.

This will be useful in the context of reusing test clusters and being able to check that an uninstall didn't leave unwanted objects behind (like we did with these namespaces).

I also moved the `requireRemoveKoperatorCRDs()` function definition a bit lower in the file to maintain some consistency in the order we define functions in the `tests/e2e/uninstall.go` file.

The small change in the `tests/e2e/helm.go` file ("delete" word) is just correcting a typo that has been harmless so far.

## Type of Change

- [x] Other (please describe) : complete a test feature

## Checklist

### DIFF
--- a/tests/e2e/uninstall.go
+++ b/tests/e2e/uninstall.go
@@ -22,22 +22,12 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-// requireRemoveKoperatorCRDs deletes the koperator CRDs
-func requireRemoveKoperatorCRDs(kubectlOptions k8s.KubectlOptions) {
-	It("Removing koperator CRDs", func() {
-		for _, crd := range koperatorCRDs() {
-			err := deleteK8sResourceNoErrNotFound(kubectlOptions, defaultDeletionTimeout, crdKind, crd)
-			Expect(err).ShouldNot(HaveOccurred())
-
-		}
-	})
-}
-
 // requireUninstallingKoperator uninstall koperator Helm chart and removes Koperator's CRDs.
 func requireUninstallingKoperator(kubectlOptions k8s.KubectlOptions) {
 	When("Uninstalling Koperator", func() {
 		requireUninstallingKoperatorHelmChart(kubectlOptions)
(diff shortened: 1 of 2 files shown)


pregnor (dismissed):
LGTM

tests/e2e/uninstall.go
```diff
@@ -205,3 +208,15 @@ func requireRemoveCertManagerCRDs(kubectlOptions k8s.KubectlOptions) {
...
 	})
 }
+
+// requireRemoveNamespace deletes the indicated namespace object
+func requireRemoveNamespace(kubectlOptions k8s.KubectlOptions, namespace string) {
+	It(fmt.Sprintf("Removing namespace %s", namespace), func() {
+		err := deleteK8sResourceNoErrNotFound(kubectlOptions, defaultDeletionTimeout, "namespace", namespace, "--wait")
+		Expect(err).ShouldNot(HaveOccurred())
+
+		remainingResources, err := getK8sResources(kubectlOptions, []string{"namespaces"}, "", "", kubectlArgGoTemplateName)
```
bartam1, line 218: This check is unnecessary because the  deleteK8sResourceNoErrNotFound should fail if the namespace has not been deleted during the specified  defaultDeletionTimeout.

bartam1 (commented):
LGTM! Only one comment.

tests/e2e/uninstall.go
mihaialexandrescu: It was there out of an abundance of caution but I see what you mean.

Fixed in 161a74729ecfb36787f3c5d541b58bd932b93f09.

pregnor (approved):

panyuenlau (approved):
LGTM

bartam1 (approved):

PR #1018, opened 2023-07-20
add snapshot and compare cluster to e2e tests

mihaialexandrescu:
## Description

This PR adds functions to create a snapshot of the cluster and compare snapshots to determine if the cluster has been cleaned up correctly with regard to the resources we're looking for.

[PartialMetadataObject](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#PartialObjectMetadata) made the most sense as the core struct to hold snapshot items info because it's a standard K8s object and it holds all the info we could need to uniquely identify an object and more (for future use cases).

I deviated from the initially proposed design of having this as part of a combination of BeforeEach and AfterEach of testInstall() and testUninstall() because I felt this (separate functions outside of install and uninstall) fits better with the structure we have now (which itself steadily deviated from the initial plan). This also lends itself better to, in time, matching cluster states all along the way (with a simple stack-like struct of snapshots), I reckon.

To make review easier and to outline said structure this is a visualisation of tests after this PR. The highlighted parts are new and I did not include the middle part where we install and uninstall a zookeeperCluster and kafkaCluster CRs):

ginkgo-test-install-uninstall-Page-2

The main matching function used for comparing snapshots is [Gomega's ConsistOf](https://onsi.github.io/gomega/#consistofelement-interface) - which roughly translates to “the lists have the same elements and overall length but the elements don’t have to be in the same order”.

Example output:

```
// Example output with diff both ways:

  the missing elements were
      <[]e2e.localComparisonPartialObjectMetadataType | len:1, cap:1>: [
          {
              GVK: {Group: "", Version: "v1", Kind: "Namespace"},
              Namespace: "",
              Name: "my-namespace",
          },
      ]
  the extra elements were
      <[]e2e.localComparisonPartialObjectMetadataType | len:2, cap:2>: [
          {
              GVK: {Group: "", Version: "v1", Kind: "Secret"},
              Namespace: "default",
              Name: "my-service-account-token-8wpls",
          },
          {
              GVK: {
                  Group: "",
                  Version: "v1",
                  Kind: "ServiceAccount",
              },
              Namespace: "default",
              Name: "my-service-account",
          },
      ]
```

I am increasing the timeout to `20m` because the run time of all current tests enabled as of this PR is around 900s (15m) and I've experienced the occasional timeout as well as sometimes running in 800'ish seconds. In time, we might need to look into ways of substantially reducing the run time of these tests.

Some other reference durations for combinations of test functions : 

```
Duration for 2 snapshots and comparison on an empty k8s cluster: 17 seconds.

Previous plus testInstall() and testUninstall() starting from an empty k8s cluster: 300 seconds (290-330).

Previous plus testInstallZookeeperCluster() + testInstallKafkaCluster(no-ssl) and testUninstallKafkaCluster() + testUninstallZookeeperCluster() starting from an empty k8s cluster: 700 seconds.
```

Tested on a PKE cluster.

## Type of Change

- [x] Other (please describe): add component to e2e tests

## Checklist

### DIFF
--- a/tests/e2e/test_snapshot.go
+++ b/tests/e2e/test_snapshot.go
@@ -18,7 +18,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
-	"time"
 
 	"github.com/gruntwork-io/terratest/modules/k8s"
 	. "github.com/onsi/ginkgo/v2"
@@ -29,6 +28,37 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
+type clusterSnapshot struct {
+	resources []metav1.PartialObjectMetadata
+}
+
+func (s *clusterSnapshot) Resources() []metav1.PartialObjectMetadata {
+	return s.resources
+}
+
+// ResourcesAsComparisonType returns a slice of a helper type that makes comparisons easier
+func (s *clusterSnapshot) ResourcesAsComparisonType() []localComparisonPartialObjectMetadataType {
+	var localList []localComparisonPartialObjectMetadataType
+	for _, r := range s.resources {
+		localList = append(localList, localComparisonPartialObjectMetadataType{
+			GVK:       r.GroupVersionKind(),
+			Namespace: r.GetNamespace(),
+			Name:      r.GetName(),
+		})
+	}
+	return localList
+}
+
+// localComparisonPartialObjectMetadataType holds a version of the minimal information required
+// to compare k8s.io/apimachinery/pkg/apis/meta/v1.PartialObjectMetadata instances
+type localComparisonPartialObjectMetadataType struct {
+	GVK       schema.GroupVersionKind
+	Namespace string
+	Name      string
+}
+
+// snapshotCluster takes a clusterSnapshot of a K8s cluster and
+// stores it into the snapshotCluster instance referenced as input
 func snapshotCluster(snapshottedInfo *clusterSnapshot) bool {
 	return When("Get cluster resources state", Ordered, func() {
 		var kubectlOptions k8s.KubectlOptions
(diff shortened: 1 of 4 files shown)


tests/e2e/test_snapshot.go
```diff
@@ -0,0 +1,194 @@
...
+		})
+
+		var clusterResourceNames []string
+		var namespacedResourceNames []string
+
+		When("Get api-resources names", func() {
+			It("Get cluster-scoped api-resources names", func() {
+				clusterResourceNames, err = listK8sResourceKinds(kubectlOptions, "", "--namespaced=false")
+				Expect(err).NotTo(HaveOccurred())
+				clusterResourceNames = pruneUnnecessaryClusterResourceNames(clusterResourceNames)
```
panyuenlau, line 80: Is it possible and problematic that this `clusterResourceNames` is returned `nil` (it looks like we are using this info to query the resources) ? If so, we might want to do a `Expect(clusterResourceNames).NotTo(BeNil())` after this line

tests/e2e/test_snapshot.go
```diff
@@ -0,0 +1,194 @@
...
+			})
+			It("Get namespaced api-resources names", func() {
+				namespacedResourceNames, err = listK8sResourceKinds(kubectlOptions, "", "--namespaced=true")
+				Expect(err).NotTo(HaveOccurred())
+				namespacedResourceNames = pruneUnnecessaryNamespacedResourceNames(namespacedResourceNames)
```
panyuenlau, line 85: Same as above, we might want to do a `Expect(namespacedResourceNames).NotTo(BeNil())` after this line, if needed

panyuenlau (approved):
The diagram is 🔥 (impressive)!

---

LGTM, just a couple of small comments/questions.

tests/e2e/test_snapshot.go
mihaialexandrescu: As far as pure coding logic is concerned, we could check for nil.

This line `clusterResourceNames = pruneUnnecessaryClusterResourceNames(clusterResourceNames)` makes it so that in fact, when we exit this `It` section, it is never actually nil (because at worst, that `prune...` function returns an empty list that's created during that function).
Also, by the very nature of the request `kubectl api-resources --verbs=list -o name --namespaced=false" doesn't return "nothing" from a working cluster (which we already check for at the beginning of the suite by listing pods and then when acquiring kubeconfig).

If you still feel like it is necessary for the atomic code to be correct, I am ok to add the `nil` check.

pregnor:
Parametrization is up next in schedule to reduce the runtime.

tests/e2e/test_snapshot.go
mihaialexandrescu: The same comment here https://github.com/banzaicloud/koperator/pull/1018#discussion_r1269389676  applies with the modification that `kubectl api-resources ... ... --namespaced=true` .

tests/e2e/test_snapshot.go
pregnor: I would be glad but this can be done in a separate PR.

I understand it's not a necessity due to integration aspects but from a unit standpoint would be better to have it here.

pregnor (dismissed):
For now this is fine, but when we parallelize the test scenarios I would like to move the snapshot functions to beforeEach/afterEach to be able to not forget running those for the new scenarios written.

CC: @banzaicloud/supertubes 

Super great clean implementation, thank you, well done @mihaialexandrescu .

tests/e2e/test_snapshot.go
mihaialexandrescu: Done in e791dd6cdbeef047f00f5801f280b2fcb8cafc1b.

tests/e2e/test_snapshot.go
mihaialexandrescu: Done in e791dd6cdbeef047f00f5801f280b2fcb8cafc1b.

tests/e2e/test_snapshot.go
panyuenlau: > Done in [e791dd6](https://github.com/banzaicloud/koperator/commit/e791dd6cdbeef047f00f5801f280b2fcb8cafc1b).

My typing is too slow :( - didn't get to see the latest commit before I sent out my last reply

PR #1020, opened 2023-07-21
feat(e2e): parallel testing with different kubeconfigs

bartam1:
## Description
New test, testCase structure added
Tests capable to run parallel with different kubeconfigs
 

## Type of Change

- [x] New Feature

## Checklist


tests/e2e/test_alltestcases.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/onsi/gomega/format"
+)
+
+var alltestCase = tests.TestCase{
+	TestDuration: 20 * time.Minute,
+	TestName:     "ALL_TESTCASE",
+	TestFn:       allTestCase,
+}
+
+// TODO (marbarta): kubectlOptions should be passed for the subtests
```
panyuenlau, line 31: Looks like this comment can be removed now

tests/e2e/pkg/tests/tests.go
```diff
@@ -0,0 +1,568 @@
...
+	"golang.org/x/exp/maps"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+//	[]testCase
+//		|
+// Classifier ==> TestPool <-- []Test <-- K8sCluster <-- ClusterInfo
+// 		^						 ^
+//		|						 |
+// 	K8sClusterPool 			   testCase
```
panyuenlau, line 42: This multi-line comment might be a bit out-of-format

tests/e2e/pkg/tests/mocks.go
```diff
@@ -0,0 +1,264 @@
...
+import (
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	. "github.com/onsi/ginkgo/v2"
+
+	//. "github.com/onsi/gomega"
+	"time"
+)
+
+// MockTestsMinimal:
+// 3 different provider 2 different version 3 different K8s cluster with 2 tests
```
panyuenlau, line 26: Suggestion: it'd be great if we can rephrase this sentence with proper grammar so people that look at this comment would be able to understand what this func is about without the need to look into the function
```suggestion
// MockTestsMinimal () returns a Classifier that has 3 different K8s clusters using 2 different K8s versions from 3 different providers. The Classifier contains 2 tests.
```

tests/e2e/pkg/tests/mocks.go
```diff
@@ -0,0 +1,264 @@
...
+			"1.24",
+			"provider3",
+			"clusterID3",
+		),
+	)
+	return NewClassifier(k8sClusterPool, mockTest1, mockTest2)
+}
+
+// MockTestsProvider:
+// 3 different provider 3 different K8s cluster with 2 tests
```
panyuenlau, line 59: Same as above, it'd be great if we can rephrase this to make it more understandable

tests/e2e/pkg/tests/mocks.go
```diff
@@ -0,0 +1,264 @@
...
+			"1.24",
+			"provider3",
+			"clusterID3",
+		),
+	)
+	return NewClassifier(k8sClusterPool, mockTest1, mockTest2)
+}
+
+// MockTestsProviderMoreTestsThenProvider:
+// 2 different provider 2 different K8s cluster with 3 tests
```
panyuenlau, line 92: Same as above, it'd be great if we can rephrase this

tests/e2e/pkg/tests/mocks.go
```diff
@@ -0,0 +1,264 @@
...
+			"1.24",
+			"provider2",
+			"clusterID2",
+		),
+	)
+	return NewClassifier(k8sClusterPool, mockTest1, mockTest2, mockTest3)
+}
+
+// MockTestsVersionOne:
+// no different version 2 different K8s cluster with 2 tests
```
panyuenlau, line 118: Same as above, it'd be great if we can rephrase this

tests/e2e/pkg/tests/mocks.go
```diff
@@ -0,0 +1,264 @@
...
+			"1.24",
+			"provider2",
+			"clusterID2",
+		),
+	)
+	return NewClassifier(k8sClusterPool, mockTest1, mockTest2)
+}
+
+// MockTestsVersion:
+// 2 different version 3 different K8s cluster with 2 tests
```
panyuenlau, line 144: Same as above, it'd be great if we can rephrase this

tests/e2e/pkg/tests/mocks.go
```diff
@@ -0,0 +1,264 @@
...
+			"1.26",
+			"provider3",
+			"clusterID4",
+		),
+	)
+	return NewClassifier(k8sClusterPool, mockTest1, mockTest2)
+}
+
+// MockTestsVersion:
+// 2 different version 2 different version 4 K8s cluster with 2 tests
```
panyuenlau, line 177: Same as above, it'd be great if we can rephrase this

tests/e2e/helm.go
```diff
@@ -164,7 +164,7 @@ func (helmDescriptor *helmDescriptor) installHelmChart(kubectlOptions k8s.Kubect
 		fixedArguments := []string{
 			"--create-namespace",
 			"--atomic",
-			"--debug",
+			//"--debug",
```
panyuenlau, line 167: Question: Is this comment intended? Or do we actually want to remove this debug flag?

panyuenlau (commented):
Didn't get to go through every single line just yet but went through most of them - looking sharp in general. 

Also left some comments that are mostly about formatting

tests/e2e/pkg/tests/tests.go
bartam1: Interesting. For me in VS studio it looks good.

tests/e2e/pkg/tests/mocks.go
bartam1: I created  those mock tests with descriptions for my presentation.
I think we should keep them with better description for testing purposes

tests/e2e/helm.go
bartam1: This generates lot of logs. Same belongs for cluster resource snapshot.
I think this should be investigated that are these usable for anything because these causes the output  truncate on github

bartam1: I put back the debug because the investigatnion is not part of this PR.

tests/e2e/pkg/tests/mocks.go
panyuenlau: Hmmm alrighty, looks like I am too early to the review party!

tests/e2e/test_alltestcases.go
```diff
@@ -0,0 +1,48 @@
...
+var alltestCase = tests.TestCase{
+	SpecsCount: 45,
+	Duration:   20 * time.Minute,
+	Name:       "ALL_TESTCASE",
+	TestFn:     allTestCase,
+}
+
+func allTestCase(kubectlOptions k8s.KubectlOptions) {
+	var snapshottedInfo = &clusterSnapshot{}
+	snapshotCluster(snapshottedInfo)
```
mihaialexandrescu, line 33: I think `snapshotCluster()` and `snapshotClusterAndCompare()` need to have `kubectlOptions` injected as parameter (and have the corresponding step removed from within them), too.

.github/actions/kind-create/action.yaml
```diff
@@ -38,10 +42,12 @@ runs:
       kubectl cluster-info
       kubectl get pods -n kube-system
       echo "current-context:" $(kubectl config current-context)
-      echo "environment-kubeconfig:" ${KUBECONFIG}
-      mkdir tests/e2e/platforms/kind/${{ inputs.kind_k8s_version }}
-      kubeconfig_path=$(pwd)/tests/e2e/platforms/kind/${{ inputs.kind_k8s_version }}/kind.kconf
+      kubeconfig_dir_path=$(pwd)/${{ inputs.kubeconfig_dir_path }}
+      mkdir -p $kubeconfig_dir_path
+      kubeconfig_filename=${{ inputs.kind_cluster_name }}-${{ inputs.kind_k8s_version }}.yaml
```
Kuvesz, line 47: Optional: We only use this variable on the next line, I don't think we really need it, just merge it's value into the next line?

.github/workflows/e2e-test.yaml
```diff
@@ -19,14 +22,41 @@ jobs:
...
-        id: setup-kind
+      - name: Setup Kind cluster1
+        id: setup-kind1
         uses: ./.github/actions/kind-create
+        with:
+          kind_cluster_name: kind1
+          kind_k8s_version: v1.24.13
+          kubeconfig_dir_path: tests/e2e/kubeconfigs
+
+      # TODO: ZookeeperCluster create time out always on the second kind cluster
```
Kuvesz, line 33: Question: Is it an actual timeout issue or can Zookeeper not start for some reason and we retry until we reach a timeout?

.github/workflows/e2e-test.yaml
```diff
@@ -19,14 +22,41 @@ jobs:
...
+           KUBECONFIG_DIR: kubeconfigs
+           MAX_TIMEOUT: "1h"
         run: |
           go work init
           go work use -r .
           make test-e2e
+
+      - name: Publish Test Results
+        uses: EnricoMi/publish-unit-test-result-action/composite@master
+        if: always()
```
Kuvesz, line 54: Question: Do we need this if?

tests/e2e/const.go
```diff
@@ -46,16 +46,17 @@ const (
...
-	defaultUserCreationWaitTime            = 10 * time.Second
-	kafkaClusterCreateTimeout              = 600 * time.Second
+	cruiseControlPodReadinessTimeout     = 50 * time.Second
+	kafkaClusterResourceReadinessTimeout = 60 * time.Second
+	defaultDeletionTimeout               = 20 * time.Second
+	defaultPodReadinessWaitTime          = 10 * time.Second
+	defaultTopicCreationWaitTime         = 10 * time.Second
+	defaultUserCreationWaitTime          = 10 * time.Second
+
+	kafkaClusterCreateTimeout              = 600 * time.Second // Increased from 600 to 700 for multiple kind
```
Kuvesz, line 56: I think you left an invalid comment here accidentally.

tests/e2e/const.go
```diff
@@ -46,16 +46,17 @@ const (
...
 	kafkaClusterResourceCleanupTimeout     = 120 * time.Second
 	kcatDeleetionTimeout                   = 40 * time.Second
-	zookeeperClusterCreateTimeout          = 4 * time.Minute
+	zookeeperClusterCreateTimeout          = 7 * time.Minute // Increased from 4 to 7 for multiple kind
```
Kuvesz, line 59: Optional: Not sure if we need this comment here, otherwise we could add one for each line explaining why we decided on that exact amount.

tests/e2e/koperator_suite_test.go
```diff
@@ -17,54 +17,188 @@
...
+		testSuiteDuration = testPool.GetTestSuiteDurationParallel()
+	}
+
+	maxTimeout, err := time.ParseDuration(viper.GetString(config.Tests.MaxTimeout))
+	if err != nil {
+		return fmt.Errorf("could not parse MaxTimeout into time.Duration: %w", err)
+	}
+	// Protection against too long test suites
+	if testSuiteDuration > maxTimeout {
+		return fmt.Errorf("tests estimated duration: '%s' bigger then maxTimeout: '%s'", testSuiteDuration.String(), maxTimeout.String())
```
Kuvesz, line 105:
```suggestion
		return fmt.Errorf("tests estimated duration: '%s' longer than maxTimeout: '%s'", testSuiteDuration.String(), maxTimeout.String())
```

tests/e2e/koperator_suite_test.go
```diff
@@ -17,54 +17,188 @@
...
+	}
+
+	// Calculated timeout can be overran with the specified time length
+	allowedOverrun, err := time.ParseDuration(viper.GetString(config.Tests.AllowedOverrunDuration))
+	if err != nil {
+		return fmt.Errorf("could not parse AllowedOverrunDuration into time.Duration: %w", err)
+	}
+	// Set TestSuite timeout based on the generated tests
+	suiteConfig.Timeout = testSuiteDuration + allowedOverrun
```
Kuvesz, line 114: Question: If we test that we don't exceed the max duration a couple of lines up from here why don't we test with this timeout value instead? If we have an overrun set we could accidentally exceed the max timeout set by the user which can cause confusion I believe.

tests/e2e/pkg/common/common.go
```diff
@@ -0,0 +1,129 @@
...
+	"emperror.dev/errors"
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	"golang.org/x/exp/maps"
+	"gopkg.in/yaml.v2"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/clientcmd/api"
+)
+
+// currentKubernetesContext returns the currently set Kubernetes context based
```
Kuvesz, line 30:
```suggestion
// CurrentEnvK8sContext returns the currently set Kubernetes context based
```

tests/e2e/pkg/common/common.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	return k8s.KubectlOptions{
+		ConfigPath:  kubeconfigPath,
+		ContextName: kubecontextName,
+		Namespace:   "",
+	}, nil
+}
+
+func GetDefaultKubeContext(kubeconfigPath string) (string, error) {
```
Kuvesz, line 66: A comment is missing for this function.

Kuvesz: Also do we not also have basically the same function in k8s.go?

tests/e2e/pkg/common/common.go
```diff
@@ -0,0 +1,129 @@
...
+			err,
+			"kubeconfig current-context is not string",
+			"current-context", structuredKubeconfig["current-context"],
+		)
+	}
+
+	return kubecontext, nil
+}
+
+// GetRawConfig creates a raw clientcmd api config
```
Kuvesz, line 94: Optional/Nitpick really: I feel a get function isn't really supposed to create things, just get them, I'd rename it to something like `CreateRawConfigFromKubeconfigPath` maybe?

tests/e2e/pkg/common/common.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	rules.ExplicitPath = kubeconfigPath
+
+	clientConfig := clientcmd.
+		NewNonInteractiveDeferredLoadingClientConfig(rules, nil)
+
+	return clientConfig.RawConfig()
+}
+
+func GetKubeContexts(kubeconfigPath string) ([]string, error) {
```
Kuvesz, line 108: A comment is missing for this function.

tests/e2e/pkg/common/common.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return maps.Keys(configs.Contexts), nil
+}
+
+// GetConfig returns kubernetes config based on the current environment.
+// If fpath is provided, loads configuration from that file. Otherwise,
+// GetConfig uses default strategy to load configuration from $KUBECONFIG,
+// .kube/config, or just returns in-cluster config.
```
Kuvesz, line 119: It seems to me this description is for another function?

tests/e2e/pkg/tests/mocks.go
```diff
@@ -0,0 +1,261 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tests
+
+import (
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	. "github.com/onsi/ginkgo/v2"
+
+	//. "github.com/onsi/gomega"
```
Kuvesz, line 21: If we don't need this import it should be removed imo.

tests/e2e/pkg/tests/mocks.go
Kuvesz: I generally agree that we should work on these if we want to show them as examples.

tests/e2e/pkg/tests/tests.go
```diff
@@ -0,0 +1,584 @@
...
+		// Describe name has to be unique and consistent between processes for proper parallelization.
+		Describe(fmt.Sprintf("%d:", describeID), decorators.Label(fmt.Sprintf("K8sID:%s", clusterID)), Ordered, func() {
+			for _, test := range tests {
+				test.Run()
+			}
+		})
+	}
+}
+
+func (tests TestPool) Sort() {
```
Kuvesz, line 103: Function description missing.

tests/e2e/pkg/tests/tests.go
```diff
@@ -0,0 +1,584 @@
...
+
+	for i := range tests {
+		if !tests[i].equal(other[i]) {
+			return false
+		}
+	}
+	return true
+}
+
+func (tests TestPool) PoolInfo() string {
```
Kuvesz, line 63: Function description missing.

tests/e2e/pkg/tests/tests.go
```diff
@@ -0,0 +1,584 @@
...
+	for _, test := range tests {
+		specsCount += test.testCase.SpecsCount
+	}
+	return specsCount
+}
+
+// GetParallelTotal returns the maximum number of the possible parallelization
+func (tests TestPool) GetParallelTotal() int {
+	return len(maps.Keys(tests.getSortedTestsByClusterID()))
+}
```
Kuvesz, line 190: Question: Isn't this just the number of clusters? Do we need a function for this?

tests/e2e/pkg/tests/tests_test.go
Kuvesz, line 1: Is this WIP? A lot of unit tests seem to be missing from this still.

Kuvesz (changes requested):
Left some comments, overall doesn't look too bad, it's just very confusing to me in some ways, I really was hoping a simpler solution is possible to achieve this.

Otherwise I would only ask that we try to refrain from these very long PRs (I know it's not always possible, this is just a reminder) when we can, I'm pretty sure I missed a lot of stuff in there because I can't pay attention long enough to review something like this properly.

.github/workflows/e2e-test.yaml
bartam1: Here you can see the problem: https://github.com/banzaicloud/koperator/runs/15416787902
It would be nice to get more information why is this fail (like pod status)

.github/workflows/e2e-test.yaml
bartam1: I dont know. I copy pasted from the documentation.

tests/e2e/const.go
bartam1: I added these for testing purposes. Because the 2 kind cluster failed because zookeeper cluster. The more time did not help.

tests/e2e/koperator_suite_test.go
bartam1: You have right.

My idea was that the netto test execution should not be bigger then the max timeout.
But I use for timeout the testSuiteDuration + allowedOverrun which can be bigger then the maxTimeout.

tests/e2e/pkg/tests/tests.go
bartam1: This Sort should can be unexported.

tests/e2e/pkg/common/common.go
bartam1: We have. I moved out because of the packages into a common package.
This should be used in the k8s.go also

tests/e2e/pkg/common/common.go
bartam1: I need to check the common go because I forgot to do that.

tests/e2e/pkg/tests/mocks.go
bartam1: fixed: [2b85e9f](https://github.com/banzaicloud/koperator/pull/1020/commits/2b85e9f4affdc52e1b2980983dcb990fa193bf2f)

tests/e2e/pkg/tests/mocks.go
bartam1: fixed: [2b85e9f](https://github.com/banzaicloud/koperator/pull/1020/commits/2b85e9f4affdc52e1b2980983dcb990fa193bf2f)

tests/e2e/pkg/tests/mocks.go
bartam1: fixed: [2b85e9f](https://github.com/banzaicloud/koperator/pull/1020/commits/2b85e9f4affdc52e1b2980983dcb990fa193bf2f)

tests/e2e/pkg/tests/mocks.go
bartam1: fixed: [2b85e9f](https://github.com/banzaicloud/koperator/pull/1020/commits/2b85e9f4affdc52e1b2980983dcb990fa193bf2f)

tests/e2e/pkg/tests/mocks.go
bartam1: fixed: [2b85e9f](https://github.com/banzaicloud/koperator/pull/1020/commits/2b85e9f4affdc52e1b2980983dcb990fa193bf2f)

tests/e2e/pkg/tests/mocks.go
bartam1: fixed: [2b85e9f](https://github.com/banzaicloud/koperator/pull/1020/commits/2b85e9f4affdc52e1b2980983dcb990fa193bf2f)

tests/e2e/test_alltestcases.go
bartam1: fixed: [b678b17](https://github.com/banzaicloud/koperator/pull/1020/commits/b678b1756446e619a22cd9024b8298a78c3ae292)

.github/actions/kind-create/action.yaml
bartam1: It is used here also: https://github.com/banzaicloud/koperator/blob/b678b1756446e619a22cd9024b8298a78c3ae292/.github/actions/kind-create/action.yaml#L48

tests/e2e/const.go
bartam1: I think it is ok to leave it now to know what is the original value for 1 kind cluster

tests/e2e/const.go
bartam1: fixed: [400c0ff](https://github.com/banzaicloud/koperator/pull/1020/commits/400c0ff9d666b916831c6613d2fc80175563f46a)

tests/e2e/koperator_suite_test.go
bartam1: fixed: [400c0ff](https://github.com/banzaicloud/koperator/pull/1020/commits/400c0ff9d666b916831c6613d2fc80175563f46a)

tests/e2e/koperator_suite_test.go
bartam1: fixed: [400c0ff](https://github.com/banzaicloud/koperator/pull/1020/commits/400c0ff9d666b916831c6613d2fc80175563f46a)

tests/e2e/pkg/common/common.go
bartam1: Im not sure about this.
The parameter name contains the rest of the information.

tests/e2e/pkg/tests/tests.go
bartam1: fixed: [4c0b935](https://github.com/banzaicloud/koperator/pull/1020/commits/4c0b935457d32e7f364886804092ebd9a8d376e6)

tests/e2e/pkg/common/common.go
bartam1: fixed: [4c0b935](https://github.com/banzaicloud/koperator/pull/1020/commits/4c0b935457d32e7f364886804092ebd9a8d376e6)

tests/e2e/pkg/common/common.go
bartam1: fixed: [4c0b935](https://github.com/banzaicloud/koperator/pull/1020/commits/4c0b935457d32e7f364886804092ebd9a8d376e6)

tests/e2e/pkg/common/common.go
bartam1: fixed: [4c0b935](https://github.com/banzaicloud/koperator/pull/1020/commits/4c0b935457d32e7f364886804092ebd9a8d376e6)

tests/e2e/pkg/common/common.go
bartam1: fixed: [4c0b935](https://github.com/banzaicloud/koperator/pull/1020/commits/4c0b935457d32e7f364886804092ebd9a8d376e6)

tests/e2e/pkg/tests/mocks.go
bartam1: fixed: [4031097](https://github.com/banzaicloud/koperator/pull/1020/commits/4031097bc70584321d1e3bf013d35a7fa92ad687)

tests/e2e/pkg/common/common.go
Kuvesz: I mean in my head a get function isn't supposed to create things, just get them but what do the others think? @panyuenlau @mihaialexandrescu @pregnor ?

.github/actions/kind-create/action.yaml
Kuvesz: Fair, it's fine then by me.

tests/e2e/pkg/tests/tests_test.go
Kuvesz: What I mean is we only have unit tests here for one struct's functions and nothing for any of the other functions. I know not everything can be unit tested easily but it would be nice to add most we can do.

.github/workflows/e2e-test.yaml
Kuvesz: Looked into it, yeah, we need it, for reference: https://stackoverflow.com/questions/58858429/how-to-run-a-github-actions-step-even-if-the-previous-step-fails-while-still-f#58859404

tests/e2e/const.go
```diff
@@ -53,7 +53,7 @@ const (
 	defaultTopicCreationWaitTime         = 10 * time.Second
 	defaultUserCreationWaitTime          = 10 * time.Second
 
-	kafkaClusterCreateTimeout              = 600 * time.Second // Increased from 600 to 700 for multiple kind
+	kafkaClusterCreateTimeout              = 600 * time.Second
 	kafkaClusterResourceCleanupTimeout     = 120 * time.Second
 	kcatDeleetionTimeout                   = 40 * time.Second
 	zookeeperClusterCreateTimeout          = 7 * time.Minute // Increased from 4 to 7 for multiple kind
```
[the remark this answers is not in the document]
bartam1: Can you please explain why should we remove this?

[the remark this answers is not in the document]
Kuvesz: I'm really not sure what sense it makes to have the explanation for this timeout here. For giving information to reviewers it would be fine as a review comment but I don't think this needs to be a comment on the code.

Kuvesz (commented):
Looks okay to me, though I still have some comments open and there are some TODOs so I'll not approve just yet.

.github/workflows/e2e-test.yaml
bartam1: Because in local environment all of the tests passed when multiple k8s cluster is used in parallel way I suspect that the problem is with timeouts or something else on github

tests/e2e/const.go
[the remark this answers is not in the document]
panyuenlau: +1 that the comment `"// Increased from 4 to 7 for multiple kind"` can be removed since it only provides the context about why the change is made, but it doesn't provide any valuable info itself along with the source code.

Or perhaps we can change the comment to something like this:
```suggestion
        // increase timeout for supporting multiple Kind clusters
	zookeeperClusterCreateTimeout          = 7 * time.Minute
```

tests/e2e/pkg/common/common.go
panyuenlau: yea...I think `CreateRawConfigFromKubeconfigPath`(or `CreateRawConfig`) is a bit more appropriate since the implementation of the function is not a simple getter function

tests/e2e/pkg/tests/tests.go
panyuenlau: @bartam1 would you mind fixing the format of this multi-line comment?

panyuenlau (commented):
Generally LGTM, will approve when the un-resolved comments are addressed

.github/workflows/e2e-test.yaml
bartam1: Please test the PR with multicluster input on your local. 
This PR is not about to run tests on github on multiple kind cluster.

tests/e2e/pkg/tests/tests_test.go
bartam1: I added multiple unit tests for functions where there are some logic.

tests/e2e/const.go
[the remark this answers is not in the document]
bartam1: I can remove but in this case who will face with this issue on multiple kind cluster on github will not know what was the original setting for one kind cluster setup.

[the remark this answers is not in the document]
bartam1: I think default values and the reason of their changes are valuable in this case.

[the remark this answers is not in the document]
bartam1: Because the increased timeout was not helped, I put back to the original value and I removed the comment
fixed: [129e7fe](https://github.com/banzaicloud/koperator/pull/1020/commits/129e7fe35b3f9a4d0c70361a85ce70caec6a99a2)

tests/e2e/pkg/tests/tests.go
panyuenlau: Huhhh, yea, I also just checked it locally and the format is fine with Goland as well. Probably it is indeed something fishy with GitHub's UI

panyuenlau (approved):
Thanks for the excellent work!

bartam1:
PR is now working on 2 kind cluster without problems. The problem was the use of wrong k8s version for test kind cluster.

PR #1023, opened 2023-07-25
WIP - Initial KRaft support

panyuenlau:
## Description

Please provide a meaningful description of what this change will do, or is for. Bonus points for including links to
related issues, other PRs, or technical references.

Note that by _not_ including a description, you are asking reviewers to do extra work to understand the context of this
change, which may lead to your PR taking much longer to review, or result in it not being reviewed at all.

## Type of Change

- [ ] Bug Fix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [ ] I have read the [contributing guidelines](../CONTRIBUTING.md)
- [ ] Existing issues have been referenced (where applicable)
- [ ] I have verified this change is not present in other open pull requests
- [ ] Functionality is documented
- [ ] All code style checks pass
- [ ] New code contribution is covered by automated tests
- [ ] All new and existing tests pass


pkg/resources/kafka/kafka.go
```diff
@@ -1274,6 +1298,8 @@ func (r *Reconciler) getK8sNodeIP(nodeName string, nodeAddressType string) (stri
 }
 
 // determineControllerId returns the ID of the controller broker of the current cluster
+// In KRaft mode, controller accesses are isolated from admin client (see KIP-590 for mode details),
+// therefore, the KRaft metadata caches intentionally choose a random broker node to report as the controller
```
panyuenlau, line 1302: Also see https://github.com/IBM/sarama/issues/2521 for relevant discussions

pkg/resources/kafka/kafka.go
```diff
@@ -306,7 +301,36 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
...
+	}
+
+	// In KRaft mode:
+	// 1. there is no way for admin client to know which node is the active controller, controllerID obtained above is just a broker ID of a random active broker (this is intentional by Kafka)
+	// 2. the follower controllers replicate the data that is written to the active controller and serves as hot standbys if the active controller fails.
+	//    Because the controllers now all track the latest state, controller fail-over will not require a lengthy reloading time to have all the state to transfer to the new controller
+	// Therefore, by setting the controllerID to be -1 to not take the controller identity into consideration when reordering the brokers
+	if r.KafkaCluster.Spec.KRaftMode {
+		controllerID = -1
+	}
```
panyuenlau, line 331: The potential optimization that we can do here is to ensure there at least one controller broker gets up first during the cluster start-up process, this way the normal brokers can register themselves to the controller broker as soon as they are ready to send out the registration requests.

Currently, if all the normal brokers are reordered in the front of the reconciliation queue and we are in a huge Kafka cluster (let's say 997 brokers + 3 controllers), the 997 normal brokers would need to wait for >=1 of the three controllers to be ready so they can register themselves; and if the K8s cluster doesn't have enough resources to the 998th broker pod in the list (which is the first controller node), the entire Kafka cluster will in a completely non-functional state.

Since this doesn't really affect the main logic, I'd like to do this part in a separate PR so this PR doesn't keep growing

config/samples/banzaicloud_v1beta1_kafkacluster.yaml
```diff
@@ -16,10 +18,12 @@ spec:
   # Specify the usable ingress controller, only envoy and istioingress supported can be left blank
   ingressController: "envoy"
   # Specify the zookeeper addresses where the Kafka should store it's metadata
+  # (this should not be set in KRaft mode)
   zkAddresses:
     - "zookeeper-client.zookeeper:2181"
   # Specify the zookeeper path where the Kafka related metadatas should be placed
   # By default it is bound to "/" and can be left blank
+  # (this should not be set in KRaft mode)
```
Kuvesz, line 26: Does this mean it will cause problems or just that it's not needed strictly?

pkg/resources/kafka/configmap.go
```diff
@@ -34,45 +36,74 @@ import (
 	"github.com/banzaicloud/koperator/pkg/resources/templates"
 	"github.com/banzaicloud/koperator/pkg/util"
 	kafkautils "github.com/banzaicloud/koperator/pkg/util/kafka"
-	zookeeperutils "github.com/banzaicloud/koperator/pkg/util/zookeeper"
 	properties "github.com/banzaicloud/koperator/properties/pkg"
 )
 
-func (r *Reconciler) getConfigProperties(bConfig *v1beta1.BrokerConfig, id int32,
+func (r *Reconciler) getConfigProperties(bConfig *v1beta1.BrokerConfig, broker v1beta1.Broker,
```
Kuvesz, line 42: Question: I'm kinda curious about the naming, what should we call brokers that are not actually brokers instead? Like the "brokers" that are just controllers from now on and do not really fit into the naming because of that.

pkg/resources/kafka/util_test.go
```diff
@@ -0,0 +1,340 @@
...
+
+import (
+	"encoding/base64"
+	"reflect"
+	"testing"
+
+	"github.com/banzaicloud/koperator/api/v1beta1"
+)
+
+func TestGenerateClusterID(t *testing.T) {
```
Kuvesz, line 25: This is a nice test idea.

pkg/util/kafka/const.go
```diff
@@ -45,7 +52,13 @@ const (
 
 // used for Cruise Control configurations
 const (
```
Kuvesz, line 54: Question (for the team really, not just Darren): Don't really see why we have to separate const blocks here? Is it just for separation?

Kuvesz (approved):
Looks good to me, left some comments. It's a long one so I may have missed something important, if so, sorry about that.

pkg/resources/kafka/configmap.go
panyuenlau: yea... it is actually confusing because we abuse the term "broker" everywhere in our code base since it was generic enough to represent any Kafka nodes in entire Kafka cluster. We might need a dedicated PR just to refactor the naming convention in the future to make it clear about "broker" vs "controller" vs "combined" across the code base

pkg/util/kafka/const.go
panyuenlau: This file was created by me - the intention was to create a visual separation between the Kafka server configurations and CC configurations keys.

Prior to this file, we used hard-coded strings for all these configuration keys everywhere, see https://github.com/banzaicloud/koperator/pull/917

panyuenlau:
FYI -  the upscale part for KRaft is on the way, but I will open a *separate PR* for that since the automatic scaling features are pretty much self-contained from the rest of the functionalities

panyuenlau:
> FYI - the upscale part for KRaft is on the way, but I will open a **separate PR** for that since the automatic scaling features are pretty much self-contained from the rest of the functionalities

Update: when testing the work for event-based scaling, found a problem that cruise control cannot re-balance for controller-nodes, and CC actually fails at [this sanity check](https://github.com/linkedin/cruise-control/blob/bf532870d31f0ca9792bf0714cd32c622e71c50c/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java#L904) for the controller-only nodes' existence because the check is using metadata returned from the Kafka cluster, and the returned metadata intentionally doesn't expose controller-only nodes. 

I have a working solution, which was to remove the controller-nodes from the `destination_broker_ids` from the rebalance `CruiseControlOperation`, it prevents from seeing the failed CC operation execution. But I think this means we **CANNOT** perform disk rebalancing for all the controller-only nodes - it'd be great if @bartam1 can provide some insights here.

---

Removing controller-only nodes also seems to be problematic - need a bit more time to figure out what's going on

bartam1:
> > FYI - the upscale part for KRaft is on the way, but I will open a **separate PR** for that since the automatic scaling features are pretty much self-contained from the rest of the functionalities
> 
> Update: when testing the work for event-based scaling, found a major problem that cruise control cannot re-balance for controller-nodes, because CC actually fails at [this sanity check](https://github.com/linkedin/cruise-control/blob/bf532870d31f0ca9792bf0714cd32c622e71c50c/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java#L904) for the controller-only nodes' existence since the check is using metadata returned from the Kafka cluster, and the returned metadata intentionally doesn't expose controller-only nodes.
> 
> I have a working solution, which was to remove the controller-nodes from the `destination_broker_ids` from the rebalance `CruiseControlOperation`, it prevents from seeing the failed CC operation execution. But I think this means we **CANNOT** perform disk rebalancing for all the controller-only nodes - it'd be great if @bartam1 can provide some insights here.
> 
> Removing controller-only nodes is also problematic for similar root-cause: the [available brokers](https://github.com/banzaicloud/koperator/blob/master/pkg/resources/kafka/kafka.go#L449) retrieved from CC doesn't have information about the controller-only nodes, and the result of that is no `GracefulDownscale` operation created for the removed broker.
> 
> **Update**: #1027 is my solution

I suggest (now or later) to  move out the pure controllers from the []broker property  and create separate []controller.
As I saw for bigger production kafka clusters the controllers are separate entities and not mixed (borker,controller).
For the user and developer standpoint would be better to handle them separately in koperator.
Advantages:
- This eliminate the chance of missconfiguration and accidental controller removal (When both mixed togeather at same place).
- It would be cleaner to the user that what is what (pure controller or broker).
- It would be also cleaner in the codebase to handle them separate. (Other configurations needed for controllers and other status)
- Lisetner configs (are different. https://github.com/apache/kafka/blob/3ba718e1599dabcf737106df27d260ba518919e1/config/kraft/controller.properties#L35) and some other configs are also not necessary for controllers.
- Controllers cant accept kafka clients and our kafka all-broker service should not forward traffic to them. (When controllers are not brokers then this would be solved easily)
- Our KafkaCluster CR brokerStatus also not so good for controllers. I think we should introduce separate status for controllers and show there which one is follower and which one is leader)
- CruiseControl operations does not work on pure controllers
- We dont need upscale and downscale or rebalance operation for pure controllers
- This separate way would be little bit similar to zookeeper cluster. (Separate nodes which does not handle kafka clients  
and separate configuration and handle) It is more like that we are used to before. 
- KafkaCluster CR size will be smaller because controllers status would be much smaller (we also dont need configuration history because we dont need downscale).
- Controllers does not need that much restarts as brokers.
Disadvantages (mostly extra work):
- Need separate logic for rolling upgrade if we dont use statefulset for controllers.
- When we want more optimization for rolling upgrade then we want to restart the active controller for the last. (with stateful set it is cannot be controllerd)

pkg/resources/kafka/configmap.go
```diff
@@ -34,45 +36,74 @@ import (
...
-	advertisedListenerConf := generateAdvertisedListenerConfig(id, r.KafkaCluster.Spec.ListenersConfig, extListenerStatuses, intListenerStatuses, controllerIntListenerStatuses)
-	if len(advertisedListenerConf) > 0 {
-		if err := config.Set(kafkautils.KafkaConfigAdvertisedListeners, advertisedListenerConf); err != nil {
-			log.Error(err, fmt.Sprintf("setting '%s' in broker configuration resulted an error", kafkautils.KafkaConfigAdvertisedListeners))
-		}
+	// Cruise Control metrics reporter configuration
+	configCCMetricsReporter(r.KafkaCluster, config, clientPass, bootstrapServers, log)
+
+	// Kafka Broker configurations
+	if r.KafkaCluster.Spec.KRaftMode {
```
bartam1, line 56: Have we got protection against that case when the user set KraftMode true but he does not specify any controller for the kafkaCluster?

api/v1beta1/kafkacluster_types.go
```diff
@@ -189,6 +206,11 @@ type Broker struct {
 
 // BrokerConfig defines the broker configuration
 type BrokerConfig struct {
+	// processRoles defines the role(s) for this particular Kafka node: "broker", "controller", or both.
+	// This must be set in KRaft mode. If set in ZooKeeper mode, Koperator ignores this configuration.
+	// +kubebuilder:validation:MaxItems=2
+	// +optional
+	Roles                []string                      `json:"processRoles,omitempty"`
```
bartam1, line 213: This also can be []KraftNodeRole

bartam1 (changes requested):
Nice work! 
I wrote some comments regarding that I found.
I also wrote some future improvement ideas

api/v1beta1/kafkacluster_types.go
```diff
@@ -62,19 +61,34 @@ const (
...
+	brokerNodeProcessRole = "broker"
 )
 
 // KafkaClusterSpec defines the desired state of KafkaCluster
 type KafkaClusterSpec struct {
+	// kRaft is used to decide where the Kafka cluster is under KRaft mode or ZooKeeper mode.
+	// This is default to be true; if set to false, the Kafka cluster is in ZooKeeper mode.
+	// +kubebuilder:default=true
+	// +optional
+	KRaftMode              bool            `json:"kRaft"`
```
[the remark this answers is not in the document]
panyuenlau: I initially had this default to false as well, but then I realized the overall product strategy of the next release of Calisti, where we seem to want to default everything to KRaft as well. Hence I made the change to default the `KafkaCluster` to be in KRaft.

This is a bigger discussion than the flag itself, let's also make the team aware of this

pkg/resources/kafka/wait-for-envoy-sidecar.sh
```diff
@@ -28,5 +28,30 @@ if [[ -n "$ENVOY_SIDECAR_STATUS" ]]; then
   done
 fi
 touch /var/run/wait/do-not-exit-yet
+
+# A few necessary steps if we are in KRaft mode
```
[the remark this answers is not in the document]
panyuenlau: Good idea, I wasn't really aware of the history of this little script

pkg/resources/kafka/kafka.go
```diff
@@ -306,7 +301,47 @@ func (r *Reconciler) Reconcile(log logr.Logger) error {
...
+		if err != nil {
+			return errors.WrapIfWithDetails(err,
+				"failed to generate quorum voters configuration",
+				"component", componentName,
+				"clusterName", r.KafkaCluster.GetName(),
+				"clusterNamespace", r.KafkaCluster.GetNamespace())
+		}
+
+		// In KRaft mode:
+		// 1. there is no way for admin client to know which node is the active controller, controllerID obtained above is just a broker ID of a random active broker (this is intentional by Kafka)
```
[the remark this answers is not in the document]
panyuenlau: > We can know which one is the leader by [requesting metadata](https://kafka.apache.org/protocol.html#The_Messages_Metadata) for the __cluster_metadata. This is a usual kafka topic and the partition leader of this topic is the active controller.

Nice little hack, except it doesnt' really work the way we want it to be...

First of all, Kafka itself intentionally doesn't want clients (either normal or admin) to have direct access toward the controllers, see section in this official design doc: https://cwiki.apache.org/confluence/display/KAFKA/KIP-631%3A+The+Quorum-based+Kafka+Controller#KIP631:TheQuorumbasedKafkaController-Networking

So our admin client in Koperator can only talk to the normal brokers, but those brokers don't have access to the `__cluster_metadata` topic, hence we have this error if we try to describe the topic in Koperator:

```
{"level":"error","ts":"2023-07-30T12:48:56.910Z","msg":"could not find controller broker","controller":"KafkaCluster","controllerGroup":"kafka.banzaicloud.io","controllerKind":"KafkaCluster","KafkaCluster":{"name":"kafka","namespace":"default"},"namespace":"default","name":"kafka","reconcileID":"a897850f-1185-47f2-b46e-ec86b6275541","component":"kafka","clusterName":"kafka","clusterNamespace":"default","error":"could not find describe topic __cluster_metadata: kafka server: Request was for a topic or partition that does not exist on this broker","errorVerbose":"kafka server: Request was for a topic or partition that does not exist on this broker\ncould not find describe topic __cluster_metadata\ngithub.com/banzaicloud/koperator/pkg/resources/kafka.(*Reconciler).determineControllerId\n\t/workspace/pkg/resources/kafka/kafka.go:1341\ngithub.com/banzaicloud/koperator/pkg/resources/kafka.(*Reconciler).Reconcile\n\t....
```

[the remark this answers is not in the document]
panyuenlau: Re: about the optimization

What I tried to explain in the 2 point point in that block of comment: all the other non-active (the followers) controllers in the controller group store the metadata in memory (they are referred to as "hot standbys"), and hence it won't take as much time as in the ZK-world for leadership transfer (arguably it takes very little time).

> reference: https://cwiki.apache.org/confluence/display/KAFKA/KIP-631%3A+The+Quorum-based+Kafka+Controller#KIP631:TheQuorumbasedKafkaController-TheMetadataTopic

With that, I don't actually think putting the active controller at the end of the reconciliation queue would be as useful as in ZK-world (assuming we managed to figure out how to find the active controller, if possible)

pkg/resources/kafka/pod.go
```diff
@@ -43,6 +43,7 @@ var (
```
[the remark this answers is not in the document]
panyuenlau: In this PR https://github.com/banzaicloud/koperator/pull/1027, I added the corresponding labels to the pods so users can distinguish the roles, but changing the pod name also sounds a good idea to me. I will probably make the change in #1027

api/v1beta1/kafkacluster_types.go
panyuenlau: Well, I thought about different potential names, but ended up following what the Kafka community itself uses: https://github.com/apache/kafka/blob/trunk/config/kraft/controller.properties#L24

This way, the users (I assume they are someone familiar with Kafka) would immediately know what this field is about. Hope that makes sense

pkg/resources/kafka/configmap.go
```diff
@@ -92,58 +123,109 @@ func (r *Reconciler) getConfigProperties(bConfig *v1beta1.BrokerConfig, id int32
...
+	if controllerListenerName != "" {
+		if err := config.Set(kafkautils.KafkaConfigControllerListenerName, controllerListenerName); err != nil {
+			log.Error(err, fmt.Sprintf(kafkautils.BrokerConfigErrorMsgTemplate, kafkautils.KafkaConfigControllerListenerName))
+		}
 	}
 
-	if len(mountPathsMerged) != 0 {
-		if err := config.Set(kafkautils.KafkaConfigBrokerLogDirectory, strings.Join(mountPathsMerged, ",")); err != nil {
-			log.Error(err, fmt.Sprintf("setting '%s' in broker configuration resulted an error", kafkautils.KafkaConfigBrokerLogDirectory))
+	// Add listener configuration
```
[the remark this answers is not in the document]
panyuenlau: The [generateListenerSpecificConfig](https://github.com/banzaicloud/koperator/blob/kraft/pkg/resources/kafka/configmap.go#L349) that is called in the next line implicitly sets the `listeners` section: https://github.com/banzaicloud/koperator/blob/kraft/pkg/resources/kafka/configmap.go#L372-L374, and this existing implementation works fine for the combined roles (and it of course still works in zk-world). The implementation later overwrites this configuration only when the node is controller-only, or broker-only.

I found the current way requires the least amount of changes from what we already have, I will probably add a comment about this implication.

pkg/resources/kafka/configmap.go
panyuenlau: Not in the main implementation, I was thinking to consider this as a invalid configuration of the `KafkaCluster` and validation webhook will reject such configuration - the validation webhook changes will be in a separate PR because I don't want this PR to keep growing. There will be other validation checks needed as well

api/util/util.go
```diff
@@ -39,3 +39,13 @@ func MergeLabels(l ...map[string]string) map[string]string {
 func LabelsForKafka(name string) map[string]string {
 	return map[string]string{"app": "kafka", "kafka_cr": name}
 }
+
+// StringSliceContains returns true if list contains s
+func StringSliceContains(list []string, s string) bool {
```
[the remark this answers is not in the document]
panyuenlau: Thx! I just copied this func and the test from the util package without taking a second thought about it. Will look into the suggestion

panyuenlau:
@bartam1 Thanks for the thoughtful suggestions! As for separating the `Controllers` from current [`Broker`](https://github.com/banzaicloud/koperator/blob/master/api/v1beta1/kafkacluster_types.go#L180), some of the advantages that you listed can be handled the same way that https://github.com/banzaicloud/koperator/pull/1027 is doing, but I do see your point that separating them would make things easier to understand and implementation would be cleaner.

However, I have some other concerns as well:
1. Despite the "controller" nodes have some distinct configurations compared to the "broker" nodes (hence we have the necessary changes in configmap.go based on the node's roles, e.g. the `listeners` config that you mentioned), there are also quite some configurations in common. Separating them from the `KafkaCluster` would be fine, but I think eventually we will end up having quite some duplicated code. For example, we might have:
```golang
type KafkaClusterSpec struct {
	BrokerConfigGroups                map[string]BrokerConfig `json:"brokerConfigGroups,omitempty"`
	Brokers                                      []Broker                `json:"brokers"`

	ControllerConfigGroups          map[string]ControllerConfig `json:"controllerConfigGroups,omitempty"`
	Controllers                                []Controller                `json:"controllers"`
```
  And the `ControllerConfig` will probably end up very similar to the [`BrokerConfig`](https://github.com/banzaicloud/koperator/blob/master/api/v1beta1/kafkacluster_types.go#L191) that we currently have, so as many of its methods.
2. Although it might be true that the combined nodes are not as common as the distinct nodes (I am actually not sure about this), it'd be nice if we can still support it in Koperator because that's what the upstream Kafka supports as well. But if we want to support this, we might want to add another field(s) under the `KafkaClusterSpec`, just for the "combined" nodes.
3. My biggest concern: if we do commit to the redesign of the `KafkaCluster` (both spec and status), I can only see it will require way more changes than we currently need to, we are NOT going to finish all these by our commit date - lets sync on this with the team.

panyuenlau:
Per internal discussion - we will move forward with separating `controller` from `broker` in both the `KafkaCluster` API level, as well as the implementation.

Therefore, closing this PR

PR #1024, opened 2023-07-25
Add koperator/api changes for KRaft support

panyuenlau:
## Description

Necessary API changes for the implementation in https://github.com/banzaicloud/koperator/pull/1023. Summary of the `KafkaCluster` API changes:

### KafkaClusterSpec

- Added a boolean `spec.kRaft` to signal if the Kafka cluster will be on KRaft mode or not, this defaults to be true because KRaft is the future! And we will likely deprecate this boolean along with all the other ZK configurations when we reach to the point to completely ditch support on ZK mode

### BrokerConfig
- Added `processRoles` to let the users to define which process role(s) the Kafka node is

### KafkaClusterStatus

- Added `ClusterID` so all the Kafka nodes in the same cluster can share this. This value is generated by Koperator at the beginning of the KafkaCluster reconciliation, and used for starting up the Kafka servers. See how it is used here: https://github.com/banzaicloud/koperator/pull/1023/files#diff-62bcca48fc04f44a3460472ea7884241469cc07272b6173d986049642eed7505R33-R37

## Type of Change

- [x] New Feature
- [x] Breaking Change

## Checklist


api/v1beta1/kafkacluster_types.go
```diff
@@ -161,18 +176,24 @@ type DisruptionBudgetWithStrategy struct {
...
 	// +kubebuilder:validation:ExclusiveMaximum=true
 	Id                int32         `json:"id"`
 	BrokerConfigGroup string        `json:"brokerConfigGroup,omitempty"`
 	ReadOnlyConfig    string        `json:"readOnlyConfig,omitempty"`
 	BrokerConfig      *BrokerConfig `json:"brokerConfig,omitempty"`
+	// processRoles defines the role(s) for this particular broker node: broker, controller, or both.
+	// This must be set in KRaft mode.
+	// +kubebuilder:validation:MaxItems=2
+	// +optional
+	Roles []string `json:"processRoles,omitempty"`
```
panyuenlau, line 196: An alternative option is to make this field as a part of the `BrokerConfig` struct. This way users can use `brokerConfigGroup` to set the process roles for the corresponding Kafka nodes.

But honestly, I can't really tell when we should put things under `BrokerConfig`, and when to put under `Broker`

Kuvesz: I'd agree with going with BrokerConfig instead as I can imagine that it might be useful in some (probably very few) scenarios to set a group of brokers as either of these roles, for example when trying to do some sort of rack awareness scenario or going for some sort of geological distribution maybe? I'm not sure if it's really worth the effort though or if it would even help in those scenarios.

Kuvesz (dismissed):
LGTM, just one reply really.

api/v1beta1/kafkacluster_types.go
panyuenlau: Another potential use case of moving this `Roles` under `BrokerConfig` - let's give all the Kafka nodes with this specific role certain amount of resources  - e.g. "controller" might need less resources than "brokers", and the "combined" nodes might need the most resources. This way, users can just use the `brokerConfigGroup` to make their life a tidy bit easier.

panyuenlau: I think i will make this change in the `KafkaCluster` API, and will update the main PR accordingly

panyuenlau: I made the change in https://github.com/banzaicloud/koperator/pull/1024/commits/54af7053ae40fae45bc2a6806b350b5c90436a53, reviewers - please let me know which version you think is better

panyuenlau:
Main PR has been closed, see this https://github.com/banzaicloud/koperator/pull/1023#issuecomment-1658637489 for details

PR #1028, opened 2023-07-31
kafkauser: copy ca to secret

LuciferInLove:
## Description

* Copy istio CA from the cert-manager ClusterIssuer Secret to KafkaUser Secret
* Use `cert-manager-namespace` flag in both `certmanagerpki` and `k8scsrpki` packages

## Type of Change
- [x] New Feature

## Checklist


pkg/pki/k8scsrpki/k8scsr_user.go
```diff
@@ -344,3 +345,45 @@ func (c *k8sCSR) Approve(ctx context.Context, signingReq *certsigningreqv1.Certi
 
 	return nil
 }
+
+func (c *k8sCSR) getCAChain(ctx context.Context, signingReq *certsigningreqv1.CertificateSigningRequest, certs []*certutil.CertificateContainer) ([]byte, error) {
+	var caChain []byte
+	if strings.Split(signingReq.Spec.SignerName, "/")[0] == v1alpha1.CertManagerSignerNamePrefix {
+		clusterIssuer := &certv1.ClusterIssuer{}
+		clusterIssuerName := strings.Split(signingReq.Spec.SignerName, "/")[1]
```
Kuvesz, line 353: I think it may be worth having a variable instead of doing the string splitting twice. For me it would make the code look better, but it's very optional.

Kuvesz (approved):
LGTM, left a comment.

pkg/pki/k8scsrpki/k8scsr_user.go
LuciferInLove: Thanks for noticing

Kuvesz (approved):

pkg/pki/k8scsrpki/k8scsr_user.go
```diff
@@ -344,3 +345,47 @@ func (c *k8sCSR) Approve(ctx context.Context, signingReq *certsigningreqv1.Certi
...
+			Name:      clusterIssuer.GetSpec().CA.SecretName,
+			Namespace: pkicommon.NamespaceCertManager,
+		}, certManagerSecret)
+		if err != nil {
+			return caChain, errors.WrapIfWithDetails(err,
+				"failed to get secret from K8s", "secretName", clusterIssuer.GetSpec().CA.SecretName,
+				"namespace", certManagerSecret.GetNamespace())
+		}
+
+		caChain = certManagerSecret.Data[defaultCertManagerIssuerSecretCertificateFile]
```
panyuenlau, line 380: Question: would it be possible that the `tls.crt` is somehow not present in the secret and we should return an error here if this happens?

panyuenlau (approved):
Left one question, but overall lgtm

pkg/pki/k8scsrpki/k8scsr_user.go
LuciferInLove: Since I get a standard cert-manager secret, this field should always exist there. But I'll add a check just in case, thanks.

panyuenlau (dismissed):

pregnor (dismissed):
I'm not a cert expert, @bartam1 is better in this stuff, found 1 thing, otherwise LGTM to me

pkg/pki/k8scsrpki/k8scsr_user.go
```diff
@@ -344,3 +346,53 @@ func (c *k8sCSR) Approve(ctx context.Context, signingReq *certsigningreqv1.Certi
...
 	return nil
 }
+
+func (c *k8sCSR) getCAChain(ctx context.Context, signingReq *certsigningreqv1.CertificateSigningRequest, certs []*certutil.CertificateContainer) ([]byte, error) {
+	var caChain []byte
+	signerName := strings.Split(signingReq.Spec.SignerName, "/")
+
+	if signerName[0] == v1alpha1.CertManagerSignerNamePrefix {
+		clusterIssuer := &certv1.ClusterIssuer{}
+		clusterIssuerName := signerName[1]
```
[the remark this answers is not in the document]
LuciferInLove: Even if the signer name is `{{signingReq.Spec.SignerName}}/` by accident, `getCAChain` won't be triggered because `cert-manager` can't issue a certificate with a wrong signerName. So, I can't imagine the case.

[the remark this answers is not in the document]
pregnor: Yeah but we are relying on the fact that some outside factor deems it unlikely code ever gets here.

But if it does get here because
A, cert-manager lets this through in the future or
B, we refactor something and this accidentally gets called
this will result in a panic that could easily have been avoided.

I would rather see this fixed before merging, but I can let this go if you think it should go in as is.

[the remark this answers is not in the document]
pregnor: Whatever, I guess I can fix it once it's in if it only bothers me.

pregnor (dismissed):

pkg/pki/k8scsrpki/k8scsr_user.go
```diff
@@ -344,3 +346,56 @@ func (c *k8sCSR) Approve(ctx context.Context, signingReq *certsigningreqv1.Certi
...
+		chain, ok := certManagerSecret.Data[defaultCertManagerIssuerSecretCertificateFile]
+		if !ok {
+			return caChain, errorfactory.New(errorfactory.FatalReconcileError{}, errors.New(notFoundCertManagerSecretField),
+				"failed to get field", "secretName", clusterIssuer.GetSpec().CA.SecretName,
+				"namespace", certManagerSecret.GetNamespace(), "field", defaultCertManagerIssuerSecretCertificateFile)
+		}
+		caChain = chain
+	} else {
+		for _, cr := range certs {
+			if cr.Certificate.IsCA {
```
bartam1, line 393: LGTM! 
Just one question.
Is the  certManagerSecret.Data[defaultCertManagerIssuerSecretCertificateFile] contains the leaf certificate? If yes then we should also include the leaf certificate when not cert-manager is used for k8s CSR signing.

bartam1 (approved):
LGTM!

pregnor:
IMO it's mergable but I didn't dare to merge because didn't want to cause any headaches to #1020

LuciferInLove:
> IMO it's mergable but I didn't dare to merge because didn't want to cause any headaches to #1020

Such a long-suffering request... 😃
If we don't plan to move to cert-manager anymore, I can close this PR. Otherwise, @bartam1 can approve if it creates more mess for #1020 or not. As another option, I can merge it later.

Issue #999, opened 2023-06-20
Inconsistent input expectations for .Values.imagePullSecrets


cotej:
## Description

The kafka-operator helm chart allows for an input value `imagePullSecrets` as defined here:
https://github.com/banzaicloud/koperator/blob/6871af2c13c68d810ea32493736c0060b6f40124/charts/kafka-operator/values.yaml#L8

In the template for the operator `Deployment`, there is an expectation that this should hold an array of objects which can be directly rendered to yaml here:
https://github.com/banzaicloud/koperator/blob/6871af2c13c68d810ea32493736c0060b6f40124/charts/kafka-operator/templates/operator-deployment-with-webhook.yaml#L140

This would suggest to use an input value that looks something like (**Example A**):
```yaml
imagePullSecrets:
  - name: my-img-pull-secret
  - name: another-img-pull-secret
```

However, in the template for authproxy `ServiceAccount`, it attempts to use this value as if it is an array of simple strings:
https://github.com/banzaicloud/koperator/blob/6871af2c13c68d810ea32493736c0060b6f40124/charts/kafka-operator/templates/authproxy-rbac.yaml#L5

Which suggests that the input value should rather look like (**Example B**):
```yaml
imagePullSecrets:
  - my-img-pull-secret
  - another-img-pull-secret
```

## Expected Behavior

The usage of this value should be consistent across all templates of the chart.

## Actual Behavior

If you attempt to install the chart with a value like **Example B** above, an error is encountered.

The chart will successfully install when using a value like **Example A**, _however_ it ends up rendering some incorrect yaml on the authproxy `ServiceAccount`, which ends up looking like:
```yaml
apiVersion: v1
imagePullSecrets:
- name: map[name:my-img-pull-secret]
kind: ServiceAccount
...
```

## Affected Version

kafka-operator chart version **0.24.1**

## Steps to Reproduce

See examples given above.

## Checklist

- [x] I have read the [contributing guidelines](../CONTRIBUTING.md)
- [x] I have verified this does not duplicate an existing issue

panyuenlau:
Thanks for reporting this, @cotej! It'd be greatly appreciated if you can come up with a PR for fixing the chart.

PR #1045, opened 2023-08-21
Fix inconsistent imagePullSecrets inputs

rappizs:
## Description
Fixes https://github.com/banzaicloud/koperator/issues/999 

## Type of Change

- [x] Bug Fix

## Checklist


pregnor (approved):
LGTM, @panyuenlau , @bartam1 please take a look at this.

bartam1 (dismissed):

bartam1:
LGTM but is there any reason why we need both?
I mean add imagePullSecret in deployment and serviceaccount same time.

charts/kafka-operator/values.yaml
```diff
@@ -6,7 +6,7 @@ replicaCount: 1
 
 # Lists the secrets you need to use to pull kafka-operator image from a private registry.
 imagePullSecrets: []
```
bartam1, line 8: I think we also can use this here:
imagePullSecrets:
or 
# imagePullSecrets:

This way we dont need to use [] which can be mislead

rappizs: Good idea, thanks!

pregnor:
> > LGTM but is there any reason why we need both?
> 
> The Deployment is for the kafka-operator itself and the ServiceAccount is for the Prometheus auth proxy. As far as I understand the Deployment uses a different ServiceAccount. Correct me @pregnor if I'm wrong.

https://github.com/banzaicloud/koperator/blob/7edcd84cef2c267bb8c788c7278e1ab57f55dfc5/charts/kafka-operator/values.yaml#L74
This is the `kafka-operator-authproxy` serviceAccount we are giving access to the images.

The other place for the imagePullSecret is at the Koperator deployment: https://github.com/banzaicloud/koperator/blob/4b4ea64892334f1cd6c8d07f5bcb9f372ab4cbce/charts/kafka-operator/templates/operator-deployment-with-webhook.yaml#L139
But the serviceAccount used there is different: https://github.com/banzaicloud/koperator/blob/4b4ea64892334f1cd6c8d07f5bcb9f372ab4cbce/charts/kafka-operator/templates/operator-deployment-with-webhook.yaml#L143
which is `kafka-operator`: https://github.com/banzaicloud/koperator/blob/4b4ea64892334f1cd6c8d07f5bcb9f372ab4cbce/charts/kafka-operator/values.yaml#L38.

So we need the `imagePullSecrets` definition at both places.

pregnor (approved):

bartam1 (approved):

Issue #1048, opened 2023-08-23
Add file integrity checks to CI pipeline


rappizs:
## Problem Statement

At the moment machine generated files are not validated in this repo. These would be essential jobs to ensure generated files are in sync with the rest of the codebase. 

## Proposed Solution

Example implementation from the pipeline repo: https://github.com/banzaicloud/pipeline/blob/master/.github/workflows/ci.yml#L28

PR #1049, opened 2023-08-23
Add file integrity checks to CI workflow

rappizs:
## Description

Fixes https://github.com/banzaicloud/koperator/issues/1048

## Type of Change

- [x] New Feature

## Checklist


.github/workflows/ci.yml
```diff
@@ -7,14 +7,53 @@ on:
...
   build:
+    name: Build
     runs-on: ubuntu-latest
 
     steps:
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.19
+          go-version: 1.20.7
```
pregnor, line 56: **Question**: Why are we upgrading the running Go version here?
The module still uses 1.19, hence my question.

Obviously I mean the other place as well to be in sync with this.

.github/workflows/ci.yml
```diff
@@ -7,14 +7,53 @@ on:
...
+            cd - > /dev/null
+          done
+
+      - name: Check generated file integrity
+        run: |
+          make generate manifests
+          git status --porcelain
+          git version
+          if [ "$(git status --porcelain)" != "" ]; then
+            printf >&2 '\n`make generate and manifests` results in a dirty state, generated files are not in sync with the source code files, differences:\n\n%s\n\n' "$(git diff)"
```
pregnor, line 43: **Optional**:  We **MAY** either use `\`make generate manifests\`` or `generating code files and manifests`, but the backtick should only be used with exact code quotes.

```suggestion
            printf >&2 '\n`make generate manifests` results in a dirty state, generated files are not in sync with the source code files, differences:\n\n%s\n\n' "$(git diff)"
```

.github/workflows/ci.yml
rappizs: We have four Go modules in this repo, the script validates them all. Three of them use 1.19 but one is 1.20 (tests/e2e):
https://github.com/banzaicloud/koperator/blob/9c2ac523143facf437711698b4aa271ac03268a5/tests/e2e/go.mod#L3

Because of this, for `go mod tidy` to work we need `>=1.20`. Also I wanted to specify an exact version, not just `1.20`.

.github/workflows/ci.yml
rappizs: Fixed

.github/workflows/ci.yml
pregnor: The fourth one is being downgraded in a subsequent PR to 1.19 because it causes other problems as well.
For now I would suggest using 1.20  because of this.

Kuvesz (dismissed):
lgtm once dependency is resolved

pregnor (approved):

.github/workflows/ci.yml
pregnor: Thanks

pregnor (dismissed):

mihaialexandrescu:
> Depends on #1032, merging is blocked for now.

https://github.com/banzaicloud/koperator/pull/1032 is now merged.

pregnor (dismissed):

.github/workflows/ci.yml
```diff
@@ -7,20 +7,57 @@ on:
...
+          go-version: 1.19
+
+      - name: Checkout code
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 #v3.5.3
+
+      - name: Check Go modules dependency file integrity
+        run: |
+          for module_file in $(find . -type f -name go.mod); do
+            module=$(dirname $module_file)
+            cd "$module"
```
szykes, line 27: `pushd` & `popd` MAY fit better to change directory.

.github/workflows/ci.yml
```diff
@@ -7,20 +7,57 @@ on:
...
+        with:
+          go-version: 1.19
+
+      - name: Checkout code
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 #v3.5.3
+
+      - name: Check Go modules dependency file integrity
+        run: |
+          for module_file in $(find . -type f -name go.mod); do
+            module=$(dirname $module_file)
```
szykes, line 26: Right now there is no folder name with whitespace but I still RECOMMEND to prepare the script for the worst case.

This snippet handles correctly the whitespace:
```
for module_file in "$(find . -type f -name go.mod)"; do
    module=$(dirname "$module_file")
```

.github/workflows/ci.yml
```diff
@@ -7,20 +7,57 @@ on:
   pull_request:
 
 jobs:
+  check-integrity:
+    name: Check integrity
+    runs-on: ubuntu-latest
```
szykes, line 12: Using the latest version of Ubuntu can lead two unexpected problems:
1. It may happen when the latest version changes to the newer Ubuntu, the Github actions will fail because the version of tool set changes.
2. You can fix the issue 1. on master but it may happen you have an old branch with the "old" Ubuntu that will not work properly because it has not been updated with the latest master.

Therefore, I RECOMMEND to speify the ubuntu version to ubuntu-22.04. [More info](https://github.com/actions/runner-images)

szykes (commented):
Added some comment.

Very welcome change!

.github/workflows/ci.yml
rappizs: Good idea, but this was actually a copy from the `build` job in this workflow. I guess I should fix that one as well.

.github/workflows/ci.yml
rappizs: I have some problems applying this suggestion, de script doesn't seem to work. I will try to figure it out why but could you please take a look @szykes and validate that this suggestion works properly?

szykes: Mea culpa. This is a truly working solution:

```
find . -type f -name go.mod | while read module_file ; do
    module=$(dirname "$module_file")
```

szykes (approved):

pregnor (approved):

Issue #661, opened 2021-09-02
request to add liveness /readiness probes in the koperator


priyavj08:
**Is your feature request related to a problem? Please describe.**
currently no health checks like liveness/readiness probes available for the operator pod

**Describe the solution you'd like to see**
Add these health checks to improve the reliability and provide a way to configure the timers through helm values

**Describe alternatives you've considered**

**Additional context**

evanfoster:
I'd like to voice my support for this issue. Liveness and readiness probes are *absolutely essential* for any workload running in Kubernetes. Doing cluster upgrades using standard draining without liveness probes leads to major stability issues.

Issue #419, opened 2020-07-16
Need information about rolling upgrade of kafka cluster.


kaushiksrinivas:
As per the doc 
https://banzaicloud.com/blog/kafka-rolling-upgrade/

We see rolling upgrade is achieved in banzai via super tubes.
And description highlights about static and dynamic property update for kafka.

From the code reading on banzai kafka, observed that dynamic reconfig is triggered via 
func (r *Reconciler) reconcilePerBrokerDynamicConfig which is using "sarama" client to update the dynamic configs without needing a broker restart.

Can you please throw some light on updating the static properties like zk connect url. ?
Is the rolling upgrade supported in kafka operator when updating a static config property (Does one have to install super tubes seperately as a pre req for kafka operator installation ?)?

Thanks in advance,
Kaushik.

kaushiksrinivas:
Hi @stoader @baluchicken 
Need your inputs.
Thanks,
Kaushik.

kaushiksrinivas:
I see 
err = r.Client.Delete(context.TODO(), currentPod) in the kafka.go > reconcileKafkaPod function.
Does the reconciler delete each pod to apply all the config updates at every reconcile (delete pod, so the pod will restart with new static configuration) ?

Would this mean, even if there is no change in the configs, kafka pods are restarted always ?

Thanks in advance.
Kaushik

baluchicken:
Hi @kaushiksrinivas if you have code related questions (how is this works, where is that), I recommend to join our [community slack](https://slack.banzaicloud.io)  and look for the #kafka-operator room. Some of your questions are already answered there. Moreover there is a huge community so maybe you will leverage from it as well.

Regarding Supertubes if you are interested on the features please contact us on our [web form.](https://banzaicloud.com/contact/)

kaushiksrinivas:
Thanks @baluchicken

kaushiksrinivas:
Hi @baluchicken 
I tried to post the queries in the banzai cloud community with #kafka-operator. But did not get responses.
is there any other channel which can be used ?

Can you please let us know if the rolling upgrade (for example config update to the kafka cluster) is only availble in the enterprise edition ?

Thanks,
Kaushik

Issue #297, opened 2020-02-11
Add e2e tests for operator


amuraru:
**Is your feature request related to a problem? Please describe.**

Add e2e tests for operator

**Describe the solution you'd like to see**

The solution should allow running the e2e tests from a dev box and in a CI environment
We can burrow the idea of integrating kind and go tests from our friends at Improbable:
https://github.com/improbable-eng/etcd-cluster-operator/blob/master/internal/test/e2e/e2e_test.go

panyuenlau:
@hong6914 FYI about this e2e testing request from the community

hong6914:
Thank you --- an internal JIRA ticket was created to track the progress.

PR #1050, opened 2023-08-24
Add liveness and readiness probes to koperator's manager pod

matewolf:
## Description

Implementation of readiness and liveness probe of koperators' manager pod. Referring to #661.

## Type of Change

- [x] New Feature

## Checklist


charts/kafka-operator/templates/operator-deployment-with-webhook.yaml
```diff
@@ -199,9 +199,24 @@ spec:
...
           {{- end }}
+          {{- if .Values.healthProbe.port }}
+            - --health-probe-addr=:{{ .Values.healthProbe.port }}
+          {{- end }}
           image: "{{ .Values.operator.image.repository }}:{{ .Values.operator.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.operator.image.pullPolicy }}
           name: manager
+          livenessProbe:
+            httpGet:
+              port: health-probes
```
Kuvesz, line 210: It might be nice to indicate this refers to health AND readiness in some way but it's not a requirement from me.

Kuvesz (dismissed):
LGTM, left a comment

charts/kafka-operator/templates/operator-deployment-with-webhook.yaml
matewolf: I have chosen this way based on some resource (other codes) in which readiness and liveness probes are referred as health checks. But I can understand your point.

config/base/manager/manager.yaml
```diff
@@ -22,6 +22,18 @@ spec:
...
+          periodSeconds: 15
+          httpGet:
+            port: 8081
+            path: /readyz
+        livenessProbe:
+          initialDelaySeconds: 15
+          periodSeconds: 10
+          httpGet:
+            port: 8081
+            path: /healthz
```
pregnor, line 36: **Optional**: In charts/kafka-operator/templates/operator-deployment-with-webhook.yaml the order was reversed, we **MAY** want to be consistent in both places, logically readiness, liveness is favored, alphabetically liveness readiness is favored. I can live with both, has my preference but leave it to you.

main.go
```diff
@@ -89,10 +90,12 @@ func main() {
 		certSigningDisabled               bool
 		certManagerEnabled                bool
 		maxKafkaTopicConcurrentReconciles int
+		healthProbeAddr                   string
```
pregnor, line 93: **Question**: Is there a particular reason we put this field here? :)

pregnor (changes requested):
Overall LGTM, there was one misplaced import line and a couple questions.

config/base/manager/manager.yaml
matewolf: I agree.

charts/kafka-operator/templates/operator-deployment-with-webhook.yaml
```diff
@@ -223,14 +238,17 @@ spec:
             - containerPort: {{ .Values.alertManager.port }}
               name: alerts
               protocol: TCP
+            - containerPort: {{ .Values.livenessReadinessProbe.port | default 8081 }}
```
panyuenlau: The `| default 8081` in this line is duplicating with https://github.com/banzaicloud/koperator/blob/7083c023df51951829184be5cf7691ad741c0859/charts/kafka-operator/values.yaml#L77 

IMO we should just use either one instead of both

charts/kafka-operator/templates/operator-deployment-with-webhook.yaml
```diff
@@ -223,14 +238,17 @@ spec:
             - containerPort: {{ .Values.alertManager.port }}
               name: alerts
               protocol: TCP
+            - containerPort: {{ .Values.healthProbes.port | default 8081 }}
```
panyuenlau, line 241: The `| default 8081` part of this line is duplicating the specified default value here https://github.com/banzaicloud/koperator/blob/1288fcf55626eaa4c77cd556bcc03a5c2de42d97/charts/kafka-operator/values.yaml#L77

This is unnecessary and creates potential confusion, IMO we should only use either way to set the default value

See`webhook.serverPort` as an example:
https://github.com/banzaicloud/koperator/blob/ca1483ec553d99dc6b3df8d6f76f06d24fca0f7e/charts/kafka-operator/templates/operator-deployment-with-webhook.yaml#L231

https://github.com/banzaicloud/koperator/blob/ca1483ec553d99dc6b3df8d6f76f06d24fca0f7e/charts/kafka-operator/values.yaml#L42

main.go
matewolf: I placed this variable on the current position because I recognized there is no alphabetical order and that's why I chose historical order.

charts/kafka-operator/templates/operator-deployment-with-webhook.yaml
matewolf: Finally I kept the name because the name must be at most 15 characters long.

szykes:
I did some investigation around the readiness and liveness probes, specifically on the implementation side, so within the controller.

I did not know, when the controller will report readyz for the first time. Luckily, the kubebuilder did a good work here I think. The HTTP endpoints registration will happen in `mgr.Start(...)` of `main.go` of `koperator` and if any error happens in that function, the `koperator` process will exit due to error check in the `main` function. This means you can do whatever you want in the `main` function because the `koperator` will not report the ready state wrongly. The `koperator` process can even hang and if it cannot continue its operation, the kubernetes will restart the pod due to missing readyz.

On the other hand the liveness probe may not do what you think. The healthyz probe is registered this way: `go cm.httpServe("health probe", cm.logger, server, cm.healthProbeListener)` from 
sigs.k8s.io/controller-runtime@v0.14.6/pkg/manager/internal.go:362. So, if the `kopeartor` hangs due to whatever reason, it will report healthz because the `healthz.Ping` is just a `{ return nil }`. Or at least this is my understanding.

I have found the original request: [Support health and liveness probes #297](https://github.com/kubernetes-sigs/controller-runtime/issues/297) and the correspoding PR: [added health probes 
#419](https://github.com/kubernetes-sigs/controller-runtime/pull/419/files). This is nice but as I understand this implementation does not check the controllers. I miss that function that is described here [Need a way to notify contoller-manager from each controller #387](https://github.com/kubernetes-sigs/controller-runtime/issues/387).

Do you know what happens when the `reconcile` function hangs?

What do you think?

matewolf:
Thank you for the detailed investigation @szykes! :))
I also checked the detailed implementation of bult-in readiness and liveness function and I raised this question to myself. I was also considering that the liveness probe should be failed if one of the controller stop running. After that I consulted @bartam1 about the desired operation of liveness probe and I think it's a correct solution if the liveness probe answers if the manager is running because @bartam1 told me there were no case when manager was running but one of the reconcile functions was hanging.

pregnor (approved):

Kuvesz (approved):

PR #1051, opened 2023-08-25
lint tests/e2e

mihaialexandrescu:
## Description

* Add linting for module `tests/e2e` in `make lint`
* Fix linting issues in module `tests/e2e`
* Annotate Makefile

## Type of Change

- [x] Refactor

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,14 +1,16 @@
 run:
     deadline: 2m
-
     skip-dirs:
-        - client
+      - client
+    go: '1.19'
+    build-tags:
+      - e2e
 
 linters:
   disable-all: true
   enable:
     - bodyclose
-    - depguard
+    # - depguard        # NOTE: we want to allow all packages
     - dogsled
     # - dupl            # NOTE: intentionally disabled
     - errcheck
@@ -45,8 +47,6 @@ linters:
     - whitespace
 
 linters-settings:
-  gocyclo:
-    min-complexity: 38
   dogsled:
     max-blank-identifiers: 3 # default: 2
   funlen:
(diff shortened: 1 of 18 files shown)


panyuenlau (approved):
LGTM, thanks for cleaning up the formatting!

tests/e2e/k8s.go
```diff
@@ -481,11 +481,9 @@ func applyK8sResourceFromTemplate(kubectlOptions k8s.KubectlOptions, templateFil
 // with the apiGroupSelector parameter the result can be narrowed by the resource group.
 // extraArgs can be any kubectl api-resources parameter.
 func listK8sResourceKinds(kubectlOptions k8s.KubectlOptions, apiGroupSelector string, extraArgs ...string) ([]string, error) {
-	logMsg := "Listing K8s resource kind"
 	args := []string{"api-resources", "--verbs", "list", "--output", "name", "--sort-by", "name"}
 
 	if apiGroupSelector != "" {
-		logMsg = fmt.Sprintf("%s group selector: %s", logMsg, apiGroupSelector)
```
pregnor, line 488: **Question**: Was this intentionally removed?

pregnor (approved):

tests/e2e/k8s.go
mihaialexandrescu: Yes - it was flagged by the linter as smth like "assigned value never used" for line 488 and once I removed that the linter was flagging the first instance of `logMsg`  as never used because it suddenly was never used (after the first correction was applied)

.golangci.yml
```diff
@@ -1,14 +1,16 @@
...
+      - e2e
 
 linters:
   disable-all: true
   enable:
     - bodyclose
-    - depguard
+    # - depguard        # NOTE: we want to allow all packages
     - dogsled
     # - dupl            # NOTE: intentionally disabled
```
Kuvesz, line 15: Love these comments, maybe later we should expand on these.

Kuvesz (approved):

.golangci.yml
mihaialexandrescu: Those are Patrik's idea - he gets the credit 😄

tests/e2e/k8s.go
pregnor: Understood, thanks.

PR #1052, opened 2023-08-25
Using up to date koperator image for e2e test

bartam1:
### Problem: 
E2E test used not freshly built Koperator image 
### Fix:
- Added workflow to build image from the current commit
- Load the image into the kind cluster
- Run E2E test with the freshly built Koperator image

## Type of Change

- [x] Bug Fix

## Checklist


.github/workflows/e2e-test.yaml
```diff
@@ -18,15 +21,24 @@ jobs:
 
       - name: Checkout code
         uses: actions/checkout@v3
+      
+      - name: Build docker image
+        run: |
+          IMG=$REPOSITORY:$GITHUB_SHA make docker-build
```
pregnor, line 27: **Recommended**: IMO we **SHOULD** use `$GITHUB_REF-$GITHUB_SHA` for the tag because that would make it easier to compare for human readers when we don't just match the commit SHA - which is difficult for humans to read/keep in mind when multiple used - but rather also have a human-readable branch/tag/PR number.

Not entirely sure if GITHUB_REF is perfect for this, but I think it's a good candidate.

tests/e2e/global.go
```diff
@@ -16,8 +16,30 @@ package e2e
...
+			koperatorImageTag = koperatorImagePathSplit[1]
+		}
+
+		koperatorLocalHelmDescriptor.SetValues = map[string]string{
+			"operator.image.repository": koperatorImageRepository,
+			"operator.image.tag":        koperatorImageTag,
+		}
+	}
+}
+
```
pregnor, line 42: **Optional**: I **MAY** have opted to use an inline local lambda function on the Helm descriptor variable setting to do this on the `koperatorLocalHelmDescriptor`. IMO it makes the logic local to there and still fulfills the purpose because the global var init runs at package loading.

.github/workflows/e2e-test.yaml
bartam1: Im not sure it is necessary because it is used only in github workflow temporary and no need readings by human.
It is not a big change I can add it if you like.

tests/e2e/global.go
bartam1: Good idea thank you!

.github/workflows/e2e-test.yaml
bartam1: It seems it need to be little bit format: ERROR: invalid tag "koperator_e2e_test:refs/pull/1052/merge-a8aa32bae40a8023c55027a5792610e8edff9d0e": invalid reference format

pregnor: We could replace slashes with dashes.

pregnor (dismissed):
LGTM

.github/workflows/e2e-test.yaml
pregnor: Also it is only true no human reads it until everything goes well, when it breaks we have to debug.

bartam1: We also should take consider that case when the branch name contains some specific character which is not accepted by docker for tag. I dont think it is worth the effort to investigate and write bash script to change specific characters to something else and use it multiple time. Im not sure about the importance of this.

pregnor: Ok, let's leave this at that for now, when we need it we will improve upon it.

Kuvesz (approved):

Kuvesz (approved):

